PR #12, opened 2021-01-13
Create vocabulary overview exercise screen

ayashanaa:


android/settings.gradle
```diff
@@ -1,4 +1,8 @@
 rootProject.name = 'LunesAppUI'
+include ':react-native-tts'
+project(':react-native-tts').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-tts/android')
+include ':react-native-tts'
+project(':react-native-tts').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-tts/android')
```
DaveOuds, line 5: Is this supposed to be here twice?

android/app/src/main/java/com/lunesappui/MainApplication.java
```diff
@@ -4,6 +4,8 @@
 import android.content.Context;
 import com.facebook.react.PackageList;
 import com.facebook.react.ReactApplication;
+import net.no_mad.tts.TextToSpeechPackage;
+import net.no_mad.tts.TextToSpeechPackage;
```
DaveOuds, line 8: twice?

android/settings.gradle
ayashanaa: I don't really know! I didn't add anything here, they were added when I installed the library! @DaveOuds

PR #17, opened 2021-01-17
Feedback fixes

ayashanaa:


components/VocabularyOverviewListItem/index.tsx
```diff
@@ -18,15 +19,21 @@ const VocabularyOverviewListItem = ({
...
-      androidParams: {
-        KEY_PARAM_PAN: -1,
-        KEY_PARAM_VOLUME: 0.5,
-        KEY_PARAM_STREAM: 'STREAM_MUSIC',
-      },
-    });
+  const handlaSpeakerClick = (audio: string) => {
+    if (audio) {
+      //audio from API
+      SoundPlayer.playUrl(audio);
```
DaveOuds, line 27: Something seems to be wrong with which audio default. The default isnt using TTS and doesn't work on IOS

PR #27, opened 2021-01-27
Create exercise 4 success fail state

ayashanaa:


components/FeedbackSection/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+          <Text style={styles.text}>
+            Great, keep it up!{'\n'}
+            The Word you filled in is correct.
+          </Text>
+        </View>
+      </View>
+    )}
+
+    {isIncorrect && (
+      <View style={[styles.messageContainer, styles.failedMessage]}>
```
DaveOuds, line 34: This can all be abstracted out into a simple modular return statement

components/AnswerSection/index.tsx
```diff
@@ -188,39 +333,84 @@ const AnswerSection = ({
...
-      <TouchableOpacity
-        onPress={checkEntry}
-        disabled={!word}
-        style={[styles.checkEntryButton, !word && styles.disabledButton]}>
-        <Text
-          style={[styles.checkEntryLabel, !word && styles.disabledButtonLabel]}>
-          Check entry
-        </Text>
-      </TouchableOpacity>
+      {(isCorrect || isIncorrect || isAlmostCorrect) && (
```
DaveOuds, line 367: This check can be done one the Feedback component

components/AnswerSection/index.tsx
```diff
@@ -21,6 +21,9 @@ import {
   Tts,
```
DaveOuds, line 21: In General I think this file needs to be split up into multiple separate components

PR #38, opened 2021-02-08
Results overview screen

ayashanaa:


DaveOuds:
Init:
- Repeat: count isnt reset

Result screens:
- Repeat button isn't showing up

PR #61, opened 2021-04-06
Lun 17 setup circle ci

SteffiStuffel:


.circleci/config.yml
```diff
@@ -0,0 +1,17 @@
+version: 2.1
+jobs:
+  check:
+    docker:
+      - image: circleci/node:12.16.3
+    steps:
+      - checkout
+      - run:
+          name: Install dependencies
+          command: yarn install
```
klinzo, line 10: I think we should use `-- frozen-lockfile` to ensure the behavior is the same as on local machines, see
https://github.com/Integreat/integreat-app/blob/main/.circleci/src/commands/restore_yarn_cache.yml
Not sure how many dependencies there already are whether a cache restoring makes sense and decreases the runtime.

hepperle00 (approved):

klinzo:
I think we should probably already go with the same circleci config file setup as in integreat, i.e.constructing the config.yml from different files. Its really not that much to implement I think and will prevent huge changes later. And I am quite sure it will be necessary at some point of time.

.circleci/config.yml
SteffiStuffel: Good point. Added it.

.circleci/config.yml
```diff
@@ -0,0 +1,52 @@
...
+                    setcommit || true
+                    if [ -z "$FILES_MODIFIED" ]
+                    then
+                      echo "The CircleCI config is up to date."
+                      exit 0;
+                    else
+                      echo "The CircleCI config is not up to date. You can update it by running `yarn run circleci:update-config`."
+                      exit 1;
+                    fi
+                name: CircleCI config up to date
```
SteffiStuffel, line 29: @klinzo Why is it needed, that the pipeline fails if dirty. Could also be ignored, because the pipeline builds an new config anyway. So why force to do it manually?

klinzo: The pipeline does not build a new config but just uses the config.yml. In order to avoid discrepancies between the actual config code in src and the stuff that is actually executed, this checks if both are the same.

klinzo (approved):

package.json
```diff
@@ -8,7 +8,8 @@
     "start": "react-native start",
     "test": "jest --silent --verbose",
     "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
-    "lint:fix": "eslint . --fix --ext .js,.jsx,.ts,.tsx"
+    "lint:fix": "eslint . --fix --ext .js,.jsx,.ts,.tsx",
```
maxammann, line 11: You can add here a ci specific linting command directly:

```
"lint:ci": "yarn run eslint --quiet --format junit -o reports/lint/junit-lint.xml",
```
Mabye it is easier to do this now.

.circleci/src/jobs/check.yml
```diff
@@ -0,0 +1,11 @@
...
+  - image: circleci/node:12.16.3
+steps:
+  - checkout
+  - check_config
+  - run:
+      name: Install dependencies
+      command: yarn install --frozen-lockfile
+  - run:
+      name: Linting check
+      command: yarn lint
```
maxammann, line 11: New line missing

maxammann (approved):

.circleci/config.yml
SteffiStuffel: So what does the "Build circle config" step do? Can it be removed? Validation is done below anyway?

PR #63, opened 2021-04-07
LUN-2: Merge index, imports and styles into one file

hepperle00:


components/Actions-test.tsx
```diff
@@ -0,0 +1,92 @@
+import 'react-native';
```
klinzo, line 1: I think this is a duplicate. Or the other file in `__tests__/components`

tsconfig.json
```diff
@@ -1,27 +1,28 @@
-
 {
```
klinzo, line 1: Maybe we should exclude this file from prettier and revert those changes

klinzo (approved):
Lets merge this asap!

SteffiStuffel (approved):

PR #64, opened 2021-04-13
LUN-22: Setup fastlane

SteffiStuffel:


android/fastlane/Fastfile
```diff
@@ -0,0 +1,38 @@
...
+#     https://docs.fastlane.tools/plugins/available-plugins
+#
+
+# Uncomment the line if you want fastlane to automatically update itself
+# update_fastlane
+
+default_platform(:android)
+
+platform :android do
+  desc "Runs all the tests"
```
klinzo, line 19: Do we need this here?

android/fastlane/Fastfile
```diff
@@ -0,0 +1,38 @@
...
+  lane :test do
+    gradle(task: "test")
+  end
+
+  desc "Submit a new Beta Build to Crashlytics Beta"
+  lane :beta do
+    gradle(task: "clean assembleRelease")
+    crashlytics
```
klinzo, line 27: I think we don't want crashlytics

docs/cicd.md
```diff
@@ -0,0 +1,25 @@
...
+### Fastlane Setup
+- Make sure ruby is installed and `ruby --version` reports the correct version
+- run `bundle config set --local path 'vendor/bundle` to set the local path
+- run `bundle install` in the project root **AND** in `./android/` **AND** in `./ios/`.
+
+
+### Lanes
+To run a lane run `bundle exec fastlane [lane]`
+
+-
```
klinzo, line 25:
```suggestion
```

android/fastlane/Appfile
```diff
@@ -0,0 +1,2 @@
+json_key_file("") # Path to the json secret file - Follow https://docs.fastlane.tools/actions/supply/#setup to get one
```
klinzo, line 1: Strange, seems like we don't even have this in integreat.

klinzo, line 1: Strange, seems like we don't even have this in integreat.

klinzo:
@maxammann maybe you could have a look at the basic setup and things we should tackle before others

android/fastlane/Appfile
maxammann: This is fine if we only deploy one app here. This avoid setting it in the Fastfile

maxammann: This should be removed and directly passed via environment variables as base64 in the fastfile

maxammann (approved):
Looks fine for just getting fastlane into the projects. Steps which should follow:

* Build a release (including signing keys) for android locally via fastlane
* Same for iOS
* Run the same build in CI for Android and iOS
* Upload the APK and IPA to testing lanes
* (Setup Sentry if wanted)

android/fastlane/Appfile
SteffiStuffel: done

SteffiStuffel: oke so i keep it? What about you comment above?

android/fastlane/Fastfile
SteffiStuffel: removed it.

klinzo (approved):

PR #67, opened 2021-04-19
LUN-7: Linting and prettier

klinzo:
It starts with the keyword *LUN*.


package.json
```diff
@@ -94,13 +100,5 @@
...
       "enzymeAdapter": "react16"
     }
-  },
-  "husky": {
-    "hooks": {
-      "pre-commit": "lint-staged"
-    }
-  },
-  "lint-staged": {
-    "*.{js,jsx,ts,tsx}": "npm run lint:fix"
```
klinzo, line 104: I can't work with this... is this fine for you? I think if you want you can add these on commit checks for yourself

.eslintrc.js
```diff
@@ -1,9 +1,64 @@
...
+    'react/prefer-es6-class': ['error', 'always'],
+
+    'react-hooks/rules-of-hooks': 'error',
+    'react-hooks/exhaustive-deps': 'error',
+
+    // TODO Use errors again for these typescript rules
+    '@typescript-eslint/explicit-function-return-type': 'warn',
+    '@typescript-eslint/no-floating-promises': 'warn',
+    '@typescript-eslint/restrict-template-expressions': 'warn',
+    '@typescript-eslint/strict-boolean-expressions': 'warn',
```
klinzo, line 37: I had to disable these because there where quite some errors :/

hepperle00 (approved):

package.json
SteffiStuffel: I'm fine with removing.

.eslintrc.js
SteffiStuffel: but we should fix tem in the future, shouldn't we? Can you open a ticket for it?

.prettierignore
```diff
@@ -0,0 +1,11 @@
...
+
+android/**/*
+ios/**/*
+
+# Ignore node_modules
+**/node_modules/**
+# Ignore all reports of code quality tools
+**/reports/**
+# Ignore autogenerated fastlane READMEs.
+**/fastlane/README.md
```
SteffiStuffel, line 11:
```suggestion
**/fastlane/README.md
```
**/vendor/**

.prettierignore
```diff
@@ -0,0 +1,11 @@
+.circleci/
+
+android/**/*
+ios/**/*
```
SteffiStuffel, line 4:
```suggestion
ios/**/*
**/vendor
```

.prettierignore
```diff
@@ -0,0 +1,11 @@
+.circleci/
```
SteffiStuffel, line 1: I think it would be nice, if we only ignore circleci/config.yml
Why not format the other files with prettier? (Tested it and works fine, output config,yml is the same for formatted src and not formatted src.

SteffiStuffel (approved):
You think it makes sense to add it to yarn scripts?
Maybe yarn lint does linting and prettier check
and a yarn prettier command does the write task?

.eslintrc.js
klinzo: Done, https://issues.tuerantuer.org/browse/LUN-55

PR #74, opened 2021-04-26
LUN-56: Type async storage

klinzo:


hepperle00 (approved):

src/navigation/NavigationTypes.ts
```diff
@@ -63,7 +63,7 @@ export type RoutesParamsType = {
       disciplineIcon: string
       trainingSetId: number
       trainingSet: string
-      exercise: string
+      exercise: ExerciseKeyType
```
SteffiStuffel, line 66: Können wir hier aus den extraParams vielleicht einen eigenen Typen machen, ist sehr unschön, wenn sich das alles x-Fach definiert ist.

src/constants/data.ts
```diff
@@ -1,8 +1,10 @@
 import { easy, hard, CorrectEntriesIcon, IncorrectEntriesIcon, AlmostCorrectEntriesIcon } from '../../assets/images'
 import { RoutesParamsType } from '../navigation/NavigationTypes'
 
+export type ExerciseKeyType = 'vocabularyOverview' | 'vocabularyTrainer'
```
SteffiStuffel, line 4: Kannst du hier bitte den union type so ändern, wie in dem issue beschrieben:

https://issues.tuerantuer.org/secure/RapidBoard.jspa?rapidView=27&view=detail&selectedIssue=LUN-57

SteffiStuffel (approved):

src/constants/data.ts
klinzo: Hmm why if there is an own issue for this? :see_no_evil:

SteffiStuffel: Thoght because you created a new type, so why not do it the correct way in the first place. Or this is an new type from you, isn't it?

PR #79, opened 2021-05-18
Lun-38: Reusable Components for Multiple Choice Exercises

hepperle00:


src/components/SingleChoiceListItem.tsx
```diff
@@ -0,0 +1,138 @@
...
+  article: string
+  correct: boolean
+  selected: boolean
+  addOpacity: boolean
+}
+
+const SingleChoiceListItem = ({ word, article, correct, selected, addOpacity }: ISingleChoiceListItemProps) => {
+  const [Container, setContainer] = useState<styled.Component>(StyledContainer)
+  const [Article, setArticle] = useState<styled.Component>(StyledArticle)
+  const [Word, setWord] = useState<styled.Component>(StyledWord)
```
klinzo, line 95: Why do you keep components in the state? Why don't you just render different components based on props? If you need to keep a state for the timeout, just save e.g. whether the timeout is expired.

src/screens/SingleChoiceExerciseScreen.tsx
```diff
@@ -1,8 +1,40 @@
...
 
 const SingleChoiceExerciseScreen = () => {
-  return <Text>Single Choice</Text>
+  const answerOptions: ISingleChoiceListItemProps[] = [
+    {
+      article: 'das',
+      word: 'Schraube',
+      correct: false,
+      selected: false,
+      addOpacity: true
```
klinzo, line 12: Which items get opacity? I don't understand this :see_no_evil:

klinzo: Also, why are items selected automatically already? I think items should only be selected once I click on them.

src/components/SingleChoiceListItem.tsx
```diff
@@ -0,0 +1,138 @@
...
+          setArticle(StyledArticleIncorrect)
+        }
+      }, 400)
+    }
+  }, [])
+
+  return (
+    <Container>
+      <Article testID='article' article={article}>
+        {article?.toLowerCase() === ARTICLES.diePlural ? 'Die' : capitalizeFirstLetter(article)}
```
klinzo, line 121: Why is this line need? I think the first letter should always be uppercase, no matter whether it is die plural or something else.

Also article should be defined all the time according to prop types.
```suggestion
        {article.toLowerCase() === ARTICLES.diePlural ? 'Die' : capitalizeFirstLetter(article)}
```

src/components/SingleChoiceListItem.tsx
```diff
@@ -0,0 +1,138 @@
...
+  selected: boolean
+  addOpacity: boolean
+}
+
+const SingleChoiceListItem = ({ word, article, correct, selected, addOpacity }: ISingleChoiceListItemProps) => {
+  const [Container, setContainer] = useState<styled.Component>(StyledContainer)
+  const [Article, setArticle] = useState<styled.Component>(StyledArticle)
+  const [Word, setWord] = useState<styled.Component>(StyledWord)
+
+  useEffect(() => {
```
klinzo, line 97: I think this timeout is not needed at all though. Normally, the other color (blue) should only be shown while tapping on the item and afterwards either green or red should be displayed according to the fact whether it is correct or not.

klinzo (changes requested):
I am not sure if I understand this correctly but I think what you should normally do here is the following:
* Add selected state to `SingleChoice.tsx` in which the selected item is saved
* Pass a callback to `SingleChoiceListItem.tsx` to change this state
* Call this callback if an item is clicked
* Set the color shown on click to blue
Since according to the issue this is not part of the issue, I'd just remove the whole timeout thing for now.

klinzo:
Looking good already though, nice work with the css :tada: :art:

src/components/SingleChoiceListItem.tsx
hepperle00: This line comes from Kiitos. ARTICLES.diePlural is defined as "die (plural)", therefore it has to be  replaced with "Die". All the other articles only have to be made uppercase (else Statement). Do you suggest a different approach?

src/components/SingleChoiceListItem.tsx
hepperle00: But wouldn`t the color blue be instantly overwritten if there was no timeout? Or do you mean to display the blue color as long as the finger still touches the button?

src/screens/SingleChoiceExerciseScreen.tsx
hepperle00: That is just mockdata to show that the styling of selected items and items that have addOpacity set to true is correct. The functionality of setting the props to the right values should be implemented in a later issue.

src/components/SingleChoiceListItem.tsx
klinzo: No, I guess that sounds reasonable then :+1:

src/components/SingleChoiceListItem.tsx
klinzo: The latter, yes. I think that is the default behavior.

src/components/SingleChoice.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react'
+import { StyleSheet, View } from 'react-native'
+import SingleChoiceListItem, { ISingleChoiceListItemProps } from './SingleChoiceListItem'
+
+export const styles = StyleSheet.create({
+  container: {
+    marginHorizontal: 35
+  }
+})
```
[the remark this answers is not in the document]
SteffiStuffel: Yes, I would also stick to StyledComponts.

src/components/SingleChoiceListItem.tsx
```diff
@@ -0,0 +1,138 @@
...
+  display: flex;
+  justify-content: flex-start;
+  flex-direction: row;
+  align-items: center;
+`
+
+const StyledContainerClicked = styled(StyledContainer)`
+  background-color: ${COLORS.lunesBlack};
+  border-color: ${COLORS.lunesBlackUltralight};
+`
```
SteffiStuffel, line 24: Can we just merge these several StyledContainers into one. As it the article is send you:

color: ${props => props.correct ? "green" : "red"};

https://styled-components.com/docs/basics#adapting-based-on-props

src/components/SingleChoiceListItem.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+const StyledContainerIncorrect = styled(StyledContainer)`
+  background-color: ${COLORS.lunesFunctionalIncorrectDark};
+  border-color: ${COLORS.lunesBlackUltralight};
+`
+
+const StyledArticle = styled.Text`
+  width: 37px;
+  height: 21px;
```
SteffiStuffel, line 38: Use px values like this in css is unfortunatly bad practice. Can you please use a flexbox and set the values there correctly? So e.g. articel is 1/5 of the container and the word is 4/5 of the container and the items are centered on [baseline](https://css-tricks.com/snippets/css/a-guide-to-flexbox/#align-items), and are justifyed with [space-around](https://css-tricks.com/snippets/css/a-guide-to-flexbox/#justify-content)

Then you would only have to set one px-value: either the padding of the text or the height of the box. (and the border-radius of course, which is fine)

src/components/SingleChoiceListItem.tsx
SteffiStuffel: I have the same opinion as @klinzo.

SteffiStuffel (commented):
Should we also add a test for this?

src/components/SingleChoiceListItem.tsx
hepperle00: Sure!

hepperle00:
> Should we also add a test for this?

I would prefer adding the tests in a seperate Issue.

src/components/SingleChoiceListItem.tsx
```diff
@@ -0,0 +1,127 @@
...
+  height: 55px;
+  margin-bottom: 1.5%;
+  border-radius: 2px;
+  border-width: 1px;
+  border-style: solid;
+  display: flex;
+  justify-content: flex-start;
+  flex-direction: row;
+  align-items: baseline;
+  border-color: ${(props: any) => {
```
klinzo, line 18: Please type this correctly :) Same for other occurrences.

klinzo (approved):
@hepperle00 should I merge this or do you want to wait for another review from Steffi?

hepperle00:
> @hepperle00 should I merge this or do you want to wait for another review from Steffi?

I would wait for Steffi.

src/components/SingleChoiceListItem.tsx
```diff
@@ -0,0 +1,127 @@
+import React, { useEffect, useState } from 'react'
+import { View } from 'react-native'
+import { capitalizeFirstLetter, getArticleColor } from '../utils/helpers'
+import { ARTICLES } from '../constants/data'
+import { COLORS } from '../constants/colors'
+import styled from 'styled-components/native'
+
+const StyledContainer = styled.View`
+  height: 55px;
```
SteffiStuffel, line 9: Do not like this height solution. Achieving the same result with padding would be nicer.

src/components/SingleChoiceListItem.tsx
```diff
@@ -0,0 +1,127 @@
...
+      return 'transparent'
+    } else {
+      return COLORS.lunesBlackUltralight
+    }
+  }};
+  background-color: ${(props: any) => {
+    if (props.pressed) {
+      return COLORS.lunesBlack
+    } else if (props.selected && props.correct) {
+      return COLORS.lunesFunctionalCorrectDark
```
SteffiStuffel, line 29: Naming the colors like this is bad practice. If we have design changes, e.g. the now lunesFunctionalCorrectDark becomes light, we have to rename this. When we have named it just after its funtion, e.g. lunesFunctionalCorrectSelected, we only have to change the color value. 
Same for other color occurences.

src/components/SingleChoiceListItem.tsx
```diff
@@ -0,0 +1,127 @@
...
+  selected,
+  addOpacity,
+  pressed
+}: SingleChoiceListItemPropsType) => {
+  return (
+    <StyledContainer pressed={pressed} correct={correct} selected={selected}>
+      <StyledArticle testID='article' article={article} selected={selected} correct={correct} pressed={pressed}>
+        {article.toLowerCase() === ARTICLES.diePlural ? 'Die' : capitalizeFirstLetter(article)}
+      </StyledArticle>
+      <StyledWord testID='word' pressed={pressed} selected={selected} correct={correct}>
```
SteffiStuffel, line 119: Why the testID?

SteffiStuffel:
> > Should we also add a test for this?
> 
> I would prefer adding the tests in a seperate Issue.

Just create an Issue for this then, please.

src/components/SingleChoiceListItem.tsx
hepperle00: they are removed now

src/components/SingleChoiceListItem.tsx
hepperle00: added a new issue for this as the colors weren't defined in this issue.

src/components/SingleChoiceListItem.tsx
hepperle00: As this creates problems with the StyledOverlay I created a new issue in which the styling should be generally improved.

SteffiStuffel (approved):

PR #86, opened 2021-06-08
LUN-66: Fix Summary Screen

SteffiStuffel:


src/components/AnswerSection.tsx
```diff
@@ -165,7 +165,7 @@ const AnswerSection = ({
 
   const storeResult = async (score: SimpleResultType): Promise<void> => {
     try {
-      const exercise = (await AsyncStorage.getExercise('vocabularyTrainer')) ?? {}
+      const exercise = (await AsyncStorage.getExercise(3)) ?? {}
```
klinzo, line 168: I think we should use a constant here

klinzo (approved):

src/components/AnswerSection.tsx
SteffiStuffel: you are right. Changed it.

PR #87, opened 2021-06-08
LUN-67: Remove empty elements

SteffiStuffel:


src/screens/ExercisesScreens.tsx
```diff
@@ -162,7 +162,10 @@ const ExercisesScreen = ({ route, navigation }: ExercisesScreenPropsType): JSX.E
     </Title>
   )
 
-  const Item = ({ item }: { item: ExerciseType }): JSX.Element => {
+  const Item = ({ item }: { item: ExerciseType }): JSX.Element | null => {
+    if (item.key === 1 || item.key === 2) {
```
klinzo, line 166: I think we should use constants here.

klinzo (approved):

PR #93, opened 2021-06-17
LUN-84: Add audio player for choice exercises

SteffiStuffel:


src/components/AnswerSection.tsx
```diff
@@ -224,67 +178,57 @@ const AnswerSection = ({
...
-          style={styles.textInput}
-          placeholder={secondAttempt ? 'Neuer Versuch' : 'Wort mit Artikel eingeben'}
-          placeholderTextColor={COLORS.lunesBlackLight}
-          value={input}
-          onChangeText={text => setInput(text)}
-          editable={result === ''}
-          onFocus={() => setIsFocused(true)}
-          onBlur={() => setIsFocused(false)}
+    <>
+      <AudioPlayer document={document} disabled={result === '' && !secondAttempt} />
```
klinzo, line 183: Is it on purpose that the AudioPlayer is not wrapped in the container view?

klinzo (approved):

src/components/AnswerSection.tsx
SteffiStuffel: Yes, for the styling.

PR #95, opened 2021-06-21
LUN-80: Fix almost correct answers

SteffiStuffel:

Did LUN-80, LUN-70, LUN-78 in this ticket.


src/components/FeedbackSection.tsx
```diff
@@ -71,7 +71,7 @@ const Feedback = ({ result, document, input, secondAttempt }: FeedbackPropsType)
       ? `Schade, deine Eingabe ist falsch. Die richtige Antwort ist: ${
           document?.article?.toLowerCase() === ARTICLES.diePlural ? 'die' : document?.article
         } ${document?.word}`
-      : `Deine Eingabe ${input} ist fast richtig. Überprüfe Groß- und Kleinschreibung.`
+      : `Deine Eingabe ${input} ist fast richtig.`
```
klinzo, line 74:
```suggestion
      : `Deine Eingabe "${input}" ist fast richtig.`
```

src/components/AnswerSection.tsx
```diff
@@ -217,7 +230,7 @@ const AnswerSection = ({
         <Actions
           tryLater={tryLater}
           giveUp={async () => {
-            await storeResult('incorrect')
+            await storeResult(secondAttempt ? 'similar' : 'incorrect')
```
klinzo, line 233: Why is an answer similar just because it is the second attempt?

klinzo (approved):

src/components/AnswerSection.tsx
SteffiStuffel: You only have a second attempt, if your first input was 'similar'.

klinzo: Sounds weird :see_no_evil: What is the reason behind this?

SteffiStuffel: If the correct answer is "Der Hammer" and you enter "Das Hammer" or "Der Hamer" you get an second attempt, because you are really close. If you enter "Der Meisel" you just get the correct answer and you input is counted as 'incorrect'. If you try and it is similar, and then give up, you solution is still counted as 'similar'. It is an design question, if you disagree or have other better ideas about it, we should discuss it on tuesday evening.

klinzo: No, just did not make sense for me at first, but with your explanation it does, thanks :+1:

PR #96, opened 2021-06-21
LUN-82: show correct if give up

SteffiStuffel:

### DIFF
--- a/src/components/FeedbackSection.tsx
+++ b/src/components/FeedbackSection.tsx
@@ -31,7 +31,7 @@ const StyledText = styled.Text`
 export interface FeedbackPropsType {
   secondAttempt: boolean
   result: string
-  document?: DocumentType
+  document: DocumentType
   input: string
 }
 
@@ -45,8 +45,8 @@ const Feedback = ({ result, document, input, secondAttempt }: FeedbackPropsType)
     Icon = IncorrectFeedbackIcon
     background = incorrect_background
     message = `Die richtige Antwort ist: ${
-      document?.article?.toLowerCase() === ARTICLES.diePlural ? 'die' : document?.article
-    } ${document?.word}`
+      document.article.toLowerCase() === ARTICLES.diePlural ? 'die' : document.article
+    } ${document.word}`
   } else {
     Icon = AlmostCorrectFeedbackIcon
     background = hint_background
--- a/src/components/FeedbackSection.tsx
+++ b/src/components/FeedbackSection.tsx
@@ -1,46 +1,32 @@
 import React from 'react'
-import { View, Text, ImageBackground, StyleSheet } from 'react-native'
 import {
-  CorrectFeedbackIcon,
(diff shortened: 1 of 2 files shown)


src/components/FeedbackSection.tsx
```diff
@@ -60,15 +60,15 @@ const Feedback = ({ result, document, input, secondAttempt }: FeedbackPropsType)
...
 
   const message =
     result === 'correct'
       ? 'Toll, weiter so! \nDeine Eingabe ist richtig.'
       : result === 'incorrect' || !secondAttempt
-      ? `Schade, deine Eingabe ist falsch. Die richtige Antwort ist: ${
+      ? `${result === 'incorrect' ? `Schade, deine Eingabe ist falsch.` : ``} Die richtige Antwort ist: ${
           document?.article?.toLowerCase() === ARTICLES.diePlural ? 'die' : document?.article
         } ${document?.word}`
       : `Deine Eingabe ${input} ist fast richtig. Überprüfe Groß- und Kleinschreibung.`
```
klinzo, line 74: Maybe this should be split in serveral vars/conditions. Maybe same for above

klinzo (approved):

src/components/FeedbackSection.tsx
SteffiStuffel: You mean the complete ? : - nesting?

klinzo: Yes, it's kind of confusing imo. But I don't really mind, so you can keep this as well.

SteffiStuffel: No, i agree, it is confusing.

src/components/FeedbackSection.tsx
```diff
@@ -50,40 +36,30 @@ export interface FeedbackPropsType {
...
+  let Icon, background, message
+  if (result === 'correct') {
+    Icon = CorrectFeedbackIcon
+    background = correct_background
+    message = 'Toll, weiter so! \nDeine Eingabe ist richtig.'
+  } else if (result === 'incorrect' || result === 'giveUp' || !secondAttempt) {
+    Icon = IncorrectFeedbackIcon
+    background = incorrect_background
+    message = `Die richtige Antwort ist: ${
+      document?.article?.toLowerCase() === ARTICLES.diePlural ? 'die' : document?.article
```
klinzo, line 48: I think you don't need those non null checks:
```suggestion
      document.article.toLowerCase() === ARTICLES.diePlural ? 'die' : document.article
```

src/components/FeedbackSection.tsx
SteffiStuffel: refactored it.

src/components/FeedbackSection.tsx
SteffiStuffel: Ja document war in den Props optional, aber ich denke das ist nicht nötig / sinnvoll. Hab es mal unoptional gemacht.

PR #98, opened 2021-06-21
LUN-82: fix

SteffiStuffel:

Undid the changes, as they break the app.


klinzo (approved):
Ups sorry :see_no_evil:

PR #99, opened 2021-06-21
LUN-96: Use Json for german labels

SteffiStuffel:


src/constants/labels.json
```diff
@@ -0,0 +1,73 @@
...
+      "description": "Wähle das richtige Wort"
+    },
+    "articleSingleChoice": {
+      "title": "Artikel üben",
+      "description": "Wähle den richtigen Artikel"
+    },
+    "write": {
+      "title": "Wort schreiben",
+      "description": "Wort mit Artikel eingeben",
+      "placeholder": {
```
klinzo, line 31: I'd not use these technical separations like `placeholder` and `button`. If you use another UI element for the same purpose it does not make sense anymore, so I think these constants should be completely independent of the concrete implementation. I guess I'd just omit the additional separation and put it one level up (directly in `write` in this example).

src/constants/data.ts
```diff
@@ -28,29 +29,29 @@ export interface ExerciseType {
 export const EXERCISES: ExerciseType[] = [
   {
     key: ExerciseKeys.vocabularyOverview,
-    title: 'Vokabelübersicht',
-    description: 'Alle Wörte anschauen',
+    title: labels.exercises.overview.title,
+    description: labels.exercises.overview.description,
```
klinzo, line 33: I think we should be consistent with naming of the exercises in all constants, labels and files... I am really confused which exercise is which :see_no_evil:

src/constants/labels.json
```diff
@@ -0,0 +1,73 @@
+{
+  "overview": {
+    "welcome": "Willkommen bei Lunes!\nLerne Vokabeln für deinen Beruf.",
+    "unit": "Lektion",
+    "units": "Lektionen",
+    "word": "Wort",
+    "words": "Wörter",
+    "exercises": "Übungen"
+  },
+  "header": {
```
klinzo, line 10: I'd just move those to overview/home/... from above as well, see comment below.

src/constants/labels.json
```diff
@@ -0,0 +1,73 @@
+{
+  "overview": {
```
klinzo, line 2: I think `main`, `home` or `general` would be more fitting. It also kind of conflicts with the exercise `overview`.

src/constants/labels.json
```diff
@@ -0,0 +1,73 @@
...
+  },
+  "header": {
+    "overview": "Übersicht",
+    "overviewExercises": "Übersicht der Übungen",
+    "cancelExercise": "Übung beenden"
+  },
+  "exercises": {
+    "overview": {
+      "title": "Vokabelübersicht",
+      "description": "Alle Wörte anschauen"
```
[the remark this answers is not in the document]
SteffiStuffel: done

src/constants/data.ts
SteffiStuffel: yes, there is a ticket for this.

src/constants/labels.json
SteffiStuffel: done.

src/constants/labels.json
SteffiStuffel: done

src/constants/labels.json
SteffiStuffel: done

klinzo (approved):

PR #100, opened 2021-06-22
LUN-81: Check for alternative article

SteffiStuffel:

### DIFF
--- a/src/screens/choice-exercises/components/SingleChoiceExercise.tsx
+++ b/src/screens/choice-exercises/components/SingleChoiceExercise.tsx
@@ -63,10 +63,18 @@ const ChoiceExerciseScreen = ({ documents, documentToAnswers, onExerciseFinished
     return found
   }
 
+  const correctAlternatives: Answer[] = currentDocument.alternatives.map(it => ({
+    article: it.article,
+    word: it.alt_word
+  }))
+
   const onClickAnswer = (selectedAnswer: Answer) => {
     setSelectedAnswer(selectedAnswer)
+    const correctSelected = [correctAnswer, ...correctAlternatives].find(
+      it => isAnswerEqual(it, selectedAnswer) ?? null
+    )
 
-    if (isAnswerEqual(selectedAnswer, correctAnswer) || isCorrectAlternative(selectedAnswer)) {
+    if (correctSelected !== null) {
       const result: DocumentResultType = { ...documents[currentWord], result: SIMPLE_RESULTS.correct }
       setResults([...results, result])
     } else {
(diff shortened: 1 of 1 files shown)


klinzo (commented):
Do you have an example where it is the case that only the article differs?

SteffiStuffel:
> Do you have an example where it is the case that only the article differs?

Metall, Elektro, Maschinen -> Grundlagen Elektronik -> Der Spachtel / Die Spachtel

klinzo:
Hmm okay... wouldn't it make more sense to just highlight both options as valid then? I.e. if an option is selected, both are highlighted green? At least I would be confused otherwise why one time `die` and the other time `der` is shown as correct. So correctAnswer would become a correctAnswers array.

SteffiStuffel:
> Hmm okay... wouldn't it make more sense to just highlight both options as valid then? I.e. if an option is selected, both are highlighted green? At least I would be confused otherwise why one time `die` and the other time `der` is shown as correct. So correctAnswer would become a correctAnswers array.

Lets wait for a design / input here, how this should be handled, an keep it like this for now.

src/screens/choice-exercises/components/SingleChoiceExercise.tsx
```diff
@@ -29,31 +29,44 @@ interface SingleChoiceExercisePropsType {
...
+  const isCorrectAlternative = (selectedAnswer: Answer): boolean => {
+    let found = false
+    currentDocument.alternatives.forEach(alternative => {
+      if (alternative.article === selectedAnswer.article && alternative.alt_word === selectedAnswer.word) {
+        setCorrectAnswer(selectedAnswer)
+        found = true
+      }
+    })
+    return found
+  }
```
klinzo, line 64:
```suggestion
  const correctAlternatives: Answer[] = currentDocument.alternatives.map(it => ({
    article: it.article,
    word: it.alt_word
  }))
```

src/screens/choice-exercises/components/SingleChoiceExercise.tsx
```diff
@@ -29,31 +29,44 @@ interface SingleChoiceExercisePropsType {
...
+
   const onClickAnswer = (selectedAnswer: Answer) => {
     setSelectedAnswer(selectedAnswer)
 
-    if (isAnswerEqual(selectedAnswer, correctAnswer)) {
+    if (isAnswerEqual(selectedAnswer, correctAnswer) || isCorrectAlternative(selectedAnswer)) {
```
klinzo, line 69:
```suggestion
    const correctSelected = [correctAnswer, ...correctAlternatives].find(it => isAnswerEqual(it, selectedAnswer)) ?? null
    if (correctSelected !== null) {
      setCorrectAnswer(correctSelected)
```

SteffiStuffel: done

src/screens/choice-exercises/components/SingleChoiceExercise.tsx
SteffiStuffel: done

PR #101, opened 2021-06-22
LUN-99: Fix result screen bugs

SteffiStuffel:

### DIFF
--- a/src/screens/ResultScreen.tsx
+++ b/src/screens/ResultScreen.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import { FlatList, StyleSheet, Text, TouchableOpacity, View } from 'react-native'
 import { RouteProp, useFocusEffect } from '@react-navigation/native'
-import { BUTTONS_THEME, RESULTS } from '../constants/data'
+import { BUTTONS_THEME, ExerciseKeys, RESULTS } from '../constants/data'
 import { COLORS } from '../constants/colors'
 import { CircularFinishIcon, NextArrow, RepeatIcon } from '../../assets/images'
 import Title from '../components/Title'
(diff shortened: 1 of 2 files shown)


src/screens/ResultScreen.tsx
```diff
@@ -124,7 +132,7 @@ const ResultScreen = ({ route, navigation }: ResultScreenPropsType): JSX.Element
 
   const Footer = (
     <>
-      {retryButton}
+      {exercise === ExerciseKeys.vocabularyTrainer && retryButton}
```
klinzo, line 135: Why do we only have a retry button for the vocabulary trainer?

src/screens/ResultScreen.tsx
```diff
@@ -72,10 +72,18 @@ interface ResultScreenPropsType {
...
   const { Icon, title, order } = resultType
-  const nextResultType = RESULTS.find(result => result.order === (order + 1) % RESULTS.length) ?? RESULTS[0]
+
+  let nextResultType = RESULTS.find(result => result.order === (order + 1) % RESULTS.length) ?? RESULTS[0]
+  if (
+    nextResultType.key === 'similar' &&
+    (exercise === ExerciseKeys.learnArticles || exercise === ExerciseKeys.singleChoice)
+  ) {
+    nextResultType = RESULTS[2]
+  }
```
klinzo, line 86: What is this doing? :thinking:

src/screens/ResultScreen.tsx
SteffiStuffel: I guess, the writeExercise is considered to be more complex and therefore has the opportunity to repeat the wrong entries. Maybe in the future, we will add this for singleChoice exercises, too, if it seems necessary.

src/screens/ResultScreen.tsx
SteffiStuffel: For single choice exercises there is no option for "almost correct" results, and therefore they are skipped.

klinzo (approved):

PR #103, opened 2021-06-22
LUN-83: Rename Exercises

SteffiStuffel:

### DIFF
--- a/src/constants/data.ts
+++ b/src/constants/data.ts
@@ -11,10 +11,10 @@ import { RoutesParamsType } from '../navigation/NavigationTypes'
 import labels from './labels.json'
 
 export const ExerciseKeys = {
-  vocabularyOverview: 0,
-  singleChoice: 1,
-  learnArticles: 2,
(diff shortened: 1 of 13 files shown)


src/constants/data.ts
```diff
@@ -28,32 +28,32 @@ export interface ExerciseType {
 
 export const EXERCISES: ExerciseType[] = [
   {
-    key: ExerciseKeys.vocabularyOverview,
+    key: ExerciseKeys.vocabularyList,
     title: labels.exercises.vocabularyList.title,
     description: labels.exercises.vocabularyList.description,
     Level: easy,
-    nextScreen: 'VocabularyOverview'
+    nextScreen: 'VocabularyList'
```
klinzo, line 35: Do we still need this? It is basically the same as the exercise anyway...

SteffiStuffel: not sure, maybe we can refactor this too, but will merge this now, and we can open a ticket if we want to.

PR #104, opened 2021-06-23
LUN-100: Implement article cms changes

SteffiStuffel:

### DIFF
--- a/src/components/VocabularyListItem.tsx
+++ b/src/components/VocabularyListItem.tsx
@@ -89,7 +89,7 @@ const VocabularyListItem = ({ document }: VocabularyListItemPropType) => {
           />
           <View>
             <Text testID='article' style={[styles.title, { backgroundColor: getArticleColor(article) }]}>
-              {article?.toLowerCase() === ARTICLES.diePlural ? 'Die' : capitalizeFirstLetter(article)}
+              {capitalizeFirstLetter(article.value)}
             </Text>
             <Text testID='word' style={styles.description}>
               {word}
--- a/src/components/VocabularyListItem.tsx
+++ b/src/components/VocabularyListItem.tsx
@@ -1,8 +1,7 @@
 import React from 'react'
 import { Image, StyleSheet, Text, View } from 'react-native'
 import { COLORS } from '../constants/colors'
-import { ARTICLES } from '../constants/data'
-import { capitalizeFirstLetter, getArticleColor } from '../utils/helpers'
+import { getArticleColor } from '../utils/helpers'
 import { widthPercentageToDP as wp } from 'react-native-responsive-screen'
 import AudioPlayer from './AudioPlayer'
 import { DocumentType } from '../constants/endpoints'
@@ -81,7 +80,7 @@ const VocabularyListItem = ({ document }: VocabularyListItemPropType) => {
         <View style={styles.item}>
           <Image
             source={{
-              uri: document.document_image[0].image
+              uri: document?.document_image[0].image
             }}
             width={24}
             height={24}
@@ -89,7 +88,7 @@ const VocabularyListItem = ({ document }: VocabularyListItemPropType) => {
           />
           <View>
             <Text testID='article' style={[styles.title, { backgroundColor: getArticleColor(article) }]}>
-              {capitalizeFirstLetter(article.value)}
+              {article.value}
             </Text>
             <Text testID='word' style={styles.description}>
               {word}
--- a/src/components/VocabularyListItem.tsx
+++ b/src/components/VocabularyListItem.tsx
@@ -78,14 +78,16 @@ const VocabularyListItem = ({ document }: VocabularyListItemPropType) => {
     <View style={styles.wrapper}>
       <View style={styles.container}>
         <View style={styles.item}>
-          <Image
-            source={{
-              uri: document?.document_image[0].image
-            }}
-            width={24}
-            height={24}
-            style={styles.image}
-          />
+          {document.document_image.length > 0 && (
+            <Image
+              source={{
+                uri: document?.document_image[0].image
+              }}
+              width={24}
+              height={24}
+              style={styles.image}
+            />
+          )}
           <View>
             <Text testID='article' style={[styles.title, { backgroundColor: getArticleColor(article) }]}>
               {article.value}
--- a/src/components/VocabularyListItem.tsx
+++ b/src/components/VocabularyListItem.tsx
@@ -81,7 +81,7 @@ const VocabularyListItem = ({ document }: VocabularyListItemPropType) => {
           {document.document_image.length > 0 && (
             <Image
               source={{
-                uri: document?.document_image[0].image
+                uri: document.document_image[0].image
               }}
               width={24}
               height={24}
--- a/src/hooks/useLoadDocuments.ts
+++ b/src/hooks/useLoadDocuments.ts
@@ -1,7 +1,6 @@
 import useLoadFromEndpoint, { ReturnType } from './useLoadFromEndpoint'
 import { AlternativeWordType, DocumentType, ENDPOINTS } from '../constants/endpoints'
 import { ARTICLES } from '../constants/data'
-import { useEffect, useState } from 'react'
 
 export interface DocumentTypeFromServer {
   id: number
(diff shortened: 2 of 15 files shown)


src/components/VocabularyListItem.tsx
```diff
@@ -79,17 +78,19 @@ const VocabularyListItem = ({ document }: VocabularyListItemPropType) => {
...
+              }}
+              width={24}
+              height={24}
+              style={styles.image}
+            />
+          )}
           <View>
             <Text testID='article' style={[styles.title, { backgroundColor: getArticleColor(article) }]}>
-              {article?.toLowerCase() === ARTICLES.diePlural ? 'Die' : capitalizeFirstLetter(article)}
+              {article.value}
```
klinzo, line 93: Previously this was capitalized:
```suggestion
              {capitalizeFirstLetter(article.value)}
```

src/components/VocabularyListItem.tsx
```diff
@@ -79,17 +78,19 @@ const VocabularyListItem = ({ document }: VocabularyListItemPropType) => {
...
-              uri: document.document_image[0].image
-            }}
-            width={24}
-            height={24}
-            style={styles.image}
-          />
+          {document.document_image.length > 0 && (
+            <Image
+              source={{
+                uri: document?.document_image[0].image
```
klinzo, line 84:
```suggestion
                uri: document.document_image[0].image
```
as the check is not used in line 81 as well.

src/hooks/useLoadDocuments.ts
```diff
@@ -1,9 +1,38 @@
...
+      return article.id === doc.article
+    })
+    if (!formattedArticle) {
+      throw new Error('Invalid Server Response')
+    }
+    const formattedDocument: DocumentType = { ...rest, article: formattedArticle }
+    documentsWithTransformedArticle.push(formattedDocument)
+  })
+  const { data, ...rest } = documents
+  return { ...rest, data: documentsWithTransformedArticle }
```
klinzo, line 29: Using `.map()` this becomes simpler :)

src/screens/choice-exercises/components/SingleChoiceExercise.tsx
```diff
@@ -49,32 +49,20 @@ const ChoiceExerciseScreen = ({ documents, documentToAnswers, onExerciseFinished
...
-        found = true
-      }
-    })
-    return found
+    return answer1.article.id === answer2.article.id && answer1.word === answer2.word
   }
 
   const correctAlternatives: Answer[] = currentDocument.alternatives.map(it => ({
-    article: it.article,
+    article: ARTICLES[it.article],
```
klinzo, line 56: Sometimes you actually compare the id, sometimes you just access the array with the id as index, I guess we should just use one of both.

src/screens/choice-exercises/WordChoiceExerciseScreen.tsx
```diff
@@ -55,11 +55,14 @@ const WordChoiceExerciseScreen = ({ navigation, route }: WordChoiceExerciseScree
   }
 
   return (
-    <SingleChoiceExercise
-      documents={documents}
-      documentToAnswers={documentToAnswers}
-      onExerciseFinished={onExerciseFinished}
-    />
+    documents !== null &&
+    documents.length !== 0 && (
```
klinzo, line 59: Why is this needed?

src/screens/WriteExerciseScreen.tsx
```diff
@@ -79,29 +78,14 @@ const WriteExerciseScreen = ({ navigation, route }: WriteExerciseScreenPropsType
...
-      getDocuments()
-    }, [retryData, trainingSetId])
-  )
+  const [currentDocumentNumber, setCurrentDocumentNumber] = useState(0)
+
+  let documents = useLoadDocuments(trainingSetId).data
+
+  if (retryData) {
+    documents = retryData.data
+  }
```
klinzo, line 88:
```suggestion
  const documents = retryData?.data ?? useLoadDocuments(trainingSetId).data
```

src/screens/WriteExerciseScreen.tsx
```diff
@@ -113,7 +97,7 @@ const WriteExerciseScreen = ({ navigation, route }: WriteExerciseScreenPropsType
           </TouchableOpacity>
         ),
         headerRight: () => (
-          <Text style={styles.headerText}>{`${currentDocumentNumber + 1} of ${documents.length}`}</Text>
+          <Text style={styles.headerText}>{`${currentDocumentNumber + 1} of ${documents?.length}`}</Text>
```
klinzo, line 100: Is this displayed if there are no documents as well? If so, it probably says something like `1 of undefined`

src/screens/WriteExerciseScreen.tsx
```diff
@@ -132,18 +116,21 @@ const WriteExerciseScreen = ({ navigation, route }: WriteExerciseScreenPropsType
...
   }
 
   const finishExercise = (): void => {
     AsyncStorage.clearSession().catch(e => console.error(e))
+    setCurrentDocumentNumber(0)
     navigation.navigate('InitialSummary', { extraParams: { ...extraParams, results: [] } })
   }
 
-  const docsLength = documents.length
+  const docsLength = documents ? documents.length : 0
```
klinzo, line 133:
```suggestion
  const docsLength = documents?.length ?? 0
```

klinzo (approved):

src/components/VocabularyListItem.tsx
SteffiStuffel: Ich habe jetzt in data.ts einfach mal die ersten Buchstaben groß gemacht, dann können wir uns das hier sparen und ich denke wir brauchen es nie in klein.

src/components/VocabularyListItem.tsx
SteffiStuffel: done

src/screens/choice-exercises/components/SingleChoiceExercise.tsx
SteffiStuffel: yeah, this is my quickfix for the alternative-words, where the mapping from id to Article is not jet implemented. I created an issue for it.

src/screens/choice-exercises/WordChoiceExerciseScreen.tsx
SteffiStuffel: it is not i guess. removed it.

SteffiStuffel: ok, it is needed, added it again. If loading is not finished, it needs to block rendering.

src/screens/WriteExerciseScreen.tsx
SteffiStuffel: if shows `1 of 0`, happens while loading is not finished, think this behaviour is oke for now.

src/screens/WriteExerciseScreen.tsx
SteffiStuffel: done

src/screens/WriteExerciseScreen.tsx
SteffiStuffel: Yeah, wanted to do this as well, but it does not work as `ESLint: React Hook "useLoadDocuments" is called conditionally. React Hooks must be called in the exact same order in every component render.(react-hooks/rules-of-hooks`

src/hooks/useLoadDocuments.ts
SteffiStuffel: yeah, strugggling a bit with this, because it changes the types of the object. I will have a closer look at this later (after deployment), as i do not have time and patience now for it.

src/screens/choice-exercises/WordChoiceExerciseScreen.tsx
klinzo: But shouldn't there be a `loading` prop which does exactly that?

SteffiStuffel: Yeah it should and it is, but if removing i get ` Type 'DocumentType[] | null' is not assignable to type 'DocumentType[]'.   Type 'null' is not assignable to type 'DocumentType[]'.`

Because ReturnType<DocumentType[]> also allows `null`. I don't know if there is a good way to tell typescript that it will only be null if loading is true.

SteffiStuffel: oke sorry, i was talking about the first line, you are right, the second line can be replaced with `!loading`

PR #107, opened 2021-06-24
LUN-xx: Fix ios font error

SteffiStuffel:


ios/Lunes.xcodeproj/project.pbxproj
```diff
@@ -785,12 +785,12 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
-				CURRENT_PROJECT_VERSION = 1;
+				CURRENT_PROJECT_VERSION = 2;
```
klinzo, line 788: Not sure if we should commit these changes, I think we don't do for Integreat as those will be outdated all the time.

klinzo (approved):

ios/Lunes.xcodeproj/project.pbxproj
SteffiStuffel: oke, i will have a closer look at this, when doing the ios-ci-stuff.

PR #112, opened 2021-06-29
LUN-90: Highlight correct word/own input in feedback section

klinzo:


src/constants/labels.json
```diff
@@ -30,9 +30,9 @@
       "newTry": "Neuer Versuch",
       "feedback": {
         "articleMissing": "Ups, deine Eingabe ist nicht vollständig. Füge einen Artikel ein.",
-        "wrong": "Die richtige Antwort ist: ",
-        "almostCorrect1": "Deine Eingabe ",
-        "almostCorrect2": " ist fast richtig.",
+        "wrong": "Die richtige Antwort ist:",
+        "almostCorrect1": "Deine Eingabe",
+        "almostCorrect2": "ist fast richtig.",
```
SteffiStuffel, line 35: Sieht das nicht doof aus, wenn die Leerzeichen fehlen?

SteffiStuffel (approved):

src/constants/labels.json
klinzo: There were double spaces, one from the labels here and one inserted in the jsx code. I think it is better to just always have labels with no padding spaces, that way you can use it whatever way you want and can easily insert spaces if needed.

PR #115, opened 2021-07-14
LUN-25: Add testflight delivery

SteffiStuffel:


ios/Lunes/Info.plist
```diff
@@ -22,6 +22,8 @@
 	<string>????</string>
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
+	<key>ITSAppUsesNonExemptEncryption</key>
+    <false/>
```
klinzo, line 26: Not sure about this one

ios/Gemfile.lock
```diff
@@ -240,7 +240,7 @@ GEM
     terminal-table (1.8.0)
       unicode-display_width (~> 1.1, >= 1.1.1)
     thread_safe (0.3.6)
-    trailblazer-option (0.1.1)
+    trailblazer-option (0.1.0)
```
klinzo, line 243: Why did you downgrade?

klinzo (approved):

ios/Gemfile.lock
SteffiStuffel: Error message due to wrong version. This way it works fine.

ios/Lunes/Info.plist
SteffiStuffel: Added this, because otherwise the export complience form must be sent manually every time.

SteffiStuffel:
@maxammann 
Would be nice if you can review. :)

ios/Lunes/Info.plist
maxammann: Yes, this is correct.

maxammann (approved):

PR #118, opened 2021-07-15
LUN-114: added variable 'documents' to react lifecycle

hepperle00:


SteffiStuffel (changes requested):
Not sure what your code should do, but for me, when checking out your branch, the write exercise is not loading anything and is always "empty" with zero words.

src/routes/write-exercise/WriteExerciseScreen.tsx
```diff
@@ -44,23 +45,31 @@ const WriteExerciseScreen = ({ navigation, route }: WriteExerciseScreenPropsType
...
     AsyncStorage.setSession(route.params).catch(e => console.error(e))
   }, [route.params])
 
+
+  useEffect(()=>{
+    if (documents && documents.length !== 0 && docs.length===0){
+      setDocs(documents)
+    }
+  }, [documents])
+
```
SteffiStuffel, line 61: I think you should change this to:
`  useEffect(()=>{
    if (documents.data){
      setDocs(documents.data)
    }
  }, [documents.loading])`

Makes more sense and also avoids the rerender-loop.

src/routes/write-exercise/WriteExerciseScreen.tsx
```diff
@@ -73,44 +82,44 @@ const WriteExerciseScreen = ({ navigation, route }: WriteExerciseScreenPropsType
   const docsLength = documents?.length ?? 0
```
SteffiStuffel, line 82: this should be changed to docs?.length

src/routes/write-exercise/WriteExerciseScreen.tsx
```diff
@@ -44,23 +45,31 @@ const WriteExerciseScreen = ({ navigation, route }: WriteExerciseScreenPropsType
...
+  useEffect(()=>{
+    if (documents && documents.length !== 0 && docs.length===0){
+      setDocs(documents)
+    }
+  }, [documents])
+
+
   if (retryData) {
-    documents = retryData.data
+    setDocs(retryData.data)
```
SteffiStuffel, line 64: Retry Data is used if you do a exercise and in the result screen click on the option, to redo/retry the wrong entries. With you changes this does not work anymore, this will work:

`  useEffect(()=>{
    if (documents.data){
      setDocs(documents.data)
    }
  }, [documents.loading])`

src/routes/write-exercise/WriteExerciseScreen.tsx
```diff
@@ -44,23 +45,31 @@ const WriteExerciseScreen = ({ navigation, route }: WriteExerciseScreenPropsType
   const { trainingSet, trainingSetId, disciplineTitle } = extraParams
   const [isLoading, setIsLoading] = useState(true)
   const [currentDocumentNumber, setCurrentDocumentNumber] = useState(0)
-
-  let documents = useLoadDocuments(trainingSetId).data
+  const [docs, setDocs] = useState<DocumentType[]>([])
+  const documents = useLoadDocuments(trainingSetId).data
```
SteffiStuffel, line 49: Renaming would make sense here, i suggest: documents and setDocuments for the State and response = useLoadDocuments(id) (without the .data, because we also need the .loading attribute)

src/routes/write-exercise/WriteExerciseScreen.tsx
hepperle00: Thanks that improves readability a lot

src/routes/write-exercise/WriteExerciseScreen.tsx
hepperle00: I see!

SteffiStuffel (commented):
Tested and works perfectlly fine.

Can you add a test for this (react-native testing library)?

src/routes/write-exercise/WriteExerciseScreen.tsx
```diff
@@ -44,23 +45,31 @@ const WriteExerciseScreen = ({ navigation, route }: WriteExerciseScreenPropsType
...
   }, [route.params])
 
-  if (retryData) {
-    documents = retryData.data
-  }
+  useEffect(() => {
+    if (response.data) {
+      setDocuments(response.data)
+    }
+  }, [response.loading])
```
Taggotty, line 59: the dependencies don't seem to be correct. 
```suggestion
  }, [response])
```

Taggotty (approved):

src/routes/write-exercise/WriteExerciseScreen.tsx
```diff
@@ -44,33 +45,33 @@ const WriteExerciseScreen = ({ navigation, route }: WriteExerciseScreenPropsType
...
-    documents = retryData.data
-  }
-
   const tryLater = (): void => {
     if (documents !== null) {
       const currDocument = documents[currentDocumentNumber]
       const newDocuments = documents.filter(d => d !== currDocument)
       newDocuments.push(currDocument)
-      documents = newDocuments
+      setNewDocuments(newDocuments)
```
Taggotty, line 62: should be a `useCallback` shouldn't it?

src/routes/write-exercise/WriteExerciseScreen.tsx
```diff
@@ -44,33 +45,33 @@ const WriteExerciseScreen = ({ navigation, route }: WriteExerciseScreenPropsType
...
   const finishExercise = (): void => {
     AsyncStorage.clearSession().catch(e => console.error(e))
     setCurrentDocumentNumber(0)
+    setNewDocuments(null)
     navigation.navigate('InitialSummary', { extraParams: { ...extraParams, results: [] } })
   }
 
-  const docsLength = documents?.length ?? 0
+
+  const docsLength = documents?.length
```
Taggotty, line 74: typescript error in line 82. `docsLength` should not be undefined

src/routes/write-exercise/WriteExerciseScreen.tsx
```diff
@@ -44,33 +45,33 @@ const WriteExerciseScreen = ({ navigation, route }: WriteExerciseScreenPropsType
...
     if (documents !== null) {
       const currDocument = documents[currentDocumentNumber]
       const newDocuments = documents.filter(d => d !== currDocument)
       newDocuments.push(currDocument)
-      documents = newDocuments
+      setNewDocuments(newDocuments)
     }
   }
 
   const finishExercise = (): void => {
```
Taggotty, line 66: should also be a react callback

Taggotty (approved):

src/routes/write-exercise/WriteExerciseScreen.tsx
Taggotty: @SteffiStuffel what do you think?

src/routes/write-exercise/WriteExerciseScreen.tsx
Taggotty: @SteffiStuffel what do you think?

src/routes/write-exercise/WriteExerciseScreen.tsx
SteffiStuffel: yes, makes sense here.

src/routes/write-exercise/WriteExerciseScreen.tsx
SteffiStuffel: You mean the hole tryLater-function? Yes, that would make sense.

SteffiStuffel (approved):

src/routes/choice-exercises/WordChoiceExerciseScreen.tsx
```diff
@@ -15,12 +15,15 @@ interface WordChoiceExerciseScreenPropsType {
...
+  const response = useLoadDocuments(trainingSetId)
+  const [documents, setDocuments] = useState<DocumentType[] | null>(null)
 
-  if (documents === null || loading) {
+  if (documents === null || response.loading) {
     return null
   }
 
+  useEffect(() => {}, [])
+
```
klinzo, line 26:
```suggestion
```

klinzo: It should not be allowed to run hooks conditionally, i.e. after an if with a return statement. Do we miss a linting rule here?

src/routes/choice-exercises/WordChoiceExerciseScreen.tsx
```diff
@@ -15,12 +15,15 @@ interface WordChoiceExerciseScreenPropsType {
 const WordChoiceExerciseScreen = ({ navigation, route }: WordChoiceExerciseScreenPropsType): ReactElement | null => {
   const { extraParams } = route.params
   const { trainingSetId } = extraParams
-  const { data: documents, loading } = useLoadDocuments(trainingSetId)
+  const response = useLoadDocuments(trainingSetId)
+  const [documents, setDocuments] = useState<DocumentType[] | null>(null)
```
klinzo, line 19: Is this used anywhere? I think this whole exercise does not work like this. `documents` will always be `null` (initial value) as it is never reset to anything else. Instead you should probably just revert this file.

src/testing/mockUseLoadFromEndpoint.ts
```diff
@@ -0,0 +1,41 @@
+import useLoadFromEndpoint from '../hooks/useLoadFromEndpoint'
```
klinzo, line 1: Just copied from Integreat, right? Normally I'd say you should use arrow functions but if it is just copied nvm :D

src/routes/write-exercise/__tests__/WriteExerciseScreen.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+        exerciseDescription: 'Description',
+        trainingSet: 'Set',
+        trainingSetId: 0,
+        level: 0
+      }
+    }
+  }
+  it('should allow to skip an exercise and try it out later', async () => {
+    mockUseLoadFromEndpointWitData(testDocuments)
+    // @ts-expect-error
```
klinzo, line 57: Please add a comment why we expect an error here

src/routes/write-exercise/WriteExerciseScreen.tsx
```diff
@@ -82,7 +81,7 @@ const WriteExerciseScreen = ({ navigation, route }: WriteExerciseScreenPropsType
             resetScrollToCoords={{ x: 0, y: 0 }}
             enableOnAndroid
             keyboardShouldPersistTaps='always'>
-            {loading && <ActivityIndicator style={styles.spinner} />}
+            {response.loading && <ActivityIndicator style={styles.spinner} />}
```
klinzo, line 84: I guess this should rather be:
```suggestion
            {!newDocuments && <ActivityIndicator style={styles.spinner} />}
```
as we don't need to wait if there are documents in the retry data.

src/routes/write-exercise/__tests__/WriteExerciseScreen.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+import WriteExerciseScreen from '../WriteExerciseScreen'
+import createNavigationMock from '../../../testing/createNavigationPropMock'
+import { RouteProp } from '@react-navigation/native'
+import { RoutesParamsType } from '../../../navigation/NavigationTypes'
+import { DocumentTypeFromServer } from '../../../hooks/useLoadDocuments'
+import labels from '../../../constants/labels.json'
+import { ReactTestInstance } from 'react-test-renderer'
+
+jest.mock('../../../components/AudioPlayer', () => {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
```
klinzo, line 13: Maybe we could just disable this rule for tests, does not have to be done in this PR oc though.

klinzo (changes requested):

src/routes/choice-exercises/WordChoiceExerciseScreen.tsx
SteffiStuffel: We have this linting rule, but the ci does not fail though. I will fix this in a separate ticket.

src/routes/write-exercise/__tests__/WriteExerciseScreen.spec.tsx
hepperle00: @SteffiStuffel what should I put in the comment?

src/testing/mockUseLoadFromEndpoint.ts
hepperle00: You're right I'll change it to arrow functions for consistency

src/routes/write-exercise/__tests__/WriteExerciseScreen.spec.tsx
SteffiStuffel: just add ` //@ts-expect-error because typescript does not know FiberNode`

klinzo (approved):
Not tested again

src/routes/write-exercise/__tests__/WriteExerciseScreen.spec.tsx
```diff
@@ -0,0 +1,79 @@
...
+        level: jest.fn()
+      }
+    }
+  }
+  it('should allow to skip an exercise and try it out later', async () => {
+    mockUseLoadFromEndpointWitData(testDocuments)
+    // @ts-expect-error because typescript does not know FiberNode
+    const getUri = (image: ReactTestInstance) => image._fiber.stateNode.props.source.uri
+
+    const { getByRole, getByText } = render(<WriteExerciseScreen route={route} navigation={navigation} />)
```
Taggotty, line 61: To fix the issue with styled component you'll have to wrap the component with a ThemeProvider
```suggestion
    const { getByRole, getByText } = render(<WriteExerciseScreen route={route} navigation={navigation} />, { wrapper: wrapWithTheme})
```

src/routes/write-exercise/__tests__/WriteExerciseScreen.spec.tsx
```diff
@@ -58,7 +59,9 @@ describe('WriteExerciseScreen', () => {
     // @ts-expect-error because typescript does not know FiberNode
     const getUri = (image: ReactTestInstance) => image._fiber.stateNode.props.source.uri
 
-    const { getByRole, getByText } = render(<WriteExerciseScreen route={route} navigation={navigation} />)
+    const { getByRole, getByText } = render(<WriteExerciseScreen route={route} navigation={navigation} />, {
+      wrapper: wrapWithTheme
+    })
     const image = await getByRole('image')
```
hepperle00, line 65: running the test fails with error message `Error: No instances found with accessibilityRole "image"`.
Any ideas on how to fix this? @SteffiStuffel @Taggotty

src/routes/write-exercise/__tests__/WriteExerciseScreen.spec.tsx
hepperle00: Works great, thanks!

PR #120, opened 2021-07-27
LUN-59: Support multiple images

SteffiStuffel:


src/components/ImageCarousel.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const { width: viewportWidth } = Dimensions.get('window')
+
+const StyledImage = styled.Image`
+  width: 100%;
+  height: 100%;
+  position: relative;
+`
+
+const styles = StyleSheet.create({
```
SteffiStuffel, line 15: How can I avoid the use of StyleSheet, if I don't want to use inline styles?

src/components/__tests__/ImageCarousel.spec.tsx
```diff
@@ -0,0 +1,22 @@
...
+      image: '1'
+    },
+    {
+      id: 2,
+      image: '2'
+    }
+  ]
+  describe('ImageCarousel', () => {
+    it('should render correctly', function () {
+      const { getAllByTestId } = render(<ImageCarousel images={images} />)
```
SteffiStuffel, line 18: Any other option than getByTestId for images?
Is it possible to trigger a swipe event?

src/components/ImageCarousel.tsx
klinzo: Just pass `style={{ position: 'absolute', marginTop: -10, alignSelf: 'center' }}` as prop I guess :thinking:

src/components/ImageCarousel.tsx
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  return (
+    <Container>
+      <Carousel
+        layout={'default'}
+        layoutCardOffset={20}
+        data={images}
+        renderItem={renderItem}
+        onSnapToItem={(index: number) => setActiveImage(index)}
```
klinzo, line 49:
```suggestion
        onSnapToItem={setActiveImage}
```

src/components/__tests__/ImageCarousel.spec.tsx
klinzo: https://callstack.github.io/react-native-testing-library/docs/api-queries#bya11yrole-byaccessibilityrole-byrole should work I guess.

klinzo:
Where can I test this?

src/components/ImageCarousel.tsx
SteffiStuffel: yeah, that's inline style :D

src/components/__tests__/ImageCarousel.spec.tsx
SteffiStuffel: unfortunately not.
Error: No instances found with accessibilityRole "image"

src/components/ImageCarousel.tsx
klinzo: ups yeah I see :D no idea than :man_shrugging:

src/components/__tests__/ImageCarousel.spec.tsx
hepperle00: Is using querySelector a possibility? https://medium.com/@drake_beth/how-to-test-images-in-react-a70053b1634a

hepperle00: I don't know how to trigger a swipe event, would it be enough to just call the snapToNext method from snap-carousel?

klinzo: @hepperle00 I think that only works in the DOM and not on native. You can set `accessibilityRole='image'` on the images though and use `getByRole`.

Regarding the swipe events: AFAICT image carousel uses scroll views under the hood so you may be able to use simple `fireEvent.scroll`: https://callstack.github.io/react-native-testing-library/docs/api/#fireeventscroll-element-reacttestinstance-data-arrayany--void
Otherwise it seems to be possible to define custom events, but not sure if that works: https://callstack.github.io/react-native-testing-library/docs/api/#fireevent

klinzo: Just tested it, the following seems to work:
```
    const eventData = {
      nativeEvent: {
        contentOffset: {
          x: 200
        },
        contentSize: {
          // Dimensions of the scrollable content
          height: 500,
          width: 100
        },
        layoutMeasurement: {
          // Dimensions of the device
          height: 100,
          width: 100
        }
      }
    }
    fireEvent.scroll(getAllByRole('image')[0], eventData)

```

SteffiStuffel:
> Where can I test this?

You can use the test-cms (change the baseUrl in axios.ts), the trainingset there has a word with two images.

src/components/__tests__/ImageCarousel.spec.tsx
SteffiStuffel: yeah thanks. This works.

klinzo:
Nice! You could add the image carousel in the write word as well.

src/routes/write-exercise/components/AnswerSection.tsx
```diff
@@ -200,7 +200,7 @@ const AnswerSection = ({
   }
 
   return (
-    <>
+    <Pressable onPress={() => Keyboard.dismiss()}>
```
klinzo, line 203: Do we need it after all?

klinzo (approved):

reports/lint/junit-lint.xml
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
```
klinzo, line 1: I think we should exclude those files

src/routes/write-exercise/components/AnswerSection.tsx
SteffiStuffel: yes when clicking not on the image but on the background.

src/services/axios.ts
```diff
@@ -1,7 +1,7 @@
 import axios from 'axios'
 
-const baseURL = 'https://lunes.tuerantuer.org/api'
-// const baseURL = 'https://lunes-test.tuerantuer.org/api'
+// const baseURL = 'https://lunes.tuerantuer.org/api'
+const baseURL = 'https://lunes-test.tuerantuer.org/api'
```
SteffiStuffel, line 4: TODO undo

PR #121, opened 2021-08-03
LUN-29: Promote Android

SteffiStuffel:


.circleci/src/jobs/promote_android.yml
```diff
@@ -0,0 +1,16 @@
+# Promote the app from the beta to the production track in the Google Play Store.
+docker:
+  - image: circleci/android:api-29-node
+resource_class: small
+shell: /bin/bash -eo pipefail
+environment:
+  FASTLANE_SKIP_UPDATE_CHECK: true
+steps:
+  - checkout
+  - restore_yarn_cache # Needed for babel-node run in read_build_config.rb. May be refactored and optimized later.
```
klinzo, line 10: Is this still true? Didn't we replace it with TS compiler?

klinzo (approved):

.circleci/src/jobs/promote_android.yml
SteffiStuffel: hmh don't know, just copied from Integreat to be honest.

SteffiStuffel: removed it, and seems to work.

PR #122, opened 2021-08-04
LUN-101: Use test cms for development

SteffiStuffel:

### DIFF
--- a/src/services/axios.ts
+++ b/src/services/axios.ts
@@ -1,7 +1,13 @@
 import axios from 'axios'
 
-const baseURL = 'https://lunes.tuerantuer.org/api'
-// const baseURL = 'https://lunes-test.tuerantuer.org/api'
+
+let baseURL: string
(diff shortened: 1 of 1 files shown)


src/services/axios.ts
```diff
@@ -1,7 +1,12 @@
 import axios from 'axios'
 
-const baseURL = 'https://lunes.tuerantuer.org/api'
-// const baseURL = 'https://lunes-test.tuerantuer.org/api'
+let baseURL: string
```
klinzo, line 3: If you wanted to get rid of the let inline if would work as well :D

klinzo (approved):

src/services/axios.ts
SteffiStuffel: yeah you are right. Will change this and be a little ashamed :D

Taggotty (approved):

PR #125, opened 2021-08-09
LUN-117: Adjust filename of browserstack uploads

SteffiStuffel:


.circleci/src/jobs/deliver_browser_stack.yml
```diff
@@ -8,13 +8,25 @@ steps:
   - prepare_workspace
   - restore_ruby_cache:
       directory: android
+  - run:
+      name: Rename apk
+      command: mv ~/attached_workspace/app-release.apk ~/attached_workspace/${CIRCLE_BRANCH}.apk
+  - run:
+      name: Rename ipa
+      command: mv ~/attached_workspace/app-release.ipa ~/attached_workspace/${CIRCLE_BRANCH}.ipa
```
klinzo, line 16: I guess you could also just copy the artifacts instead of moving them, then no undo would be necessary.

klinzo (approved):

.circleci/src/jobs/deliver_browser_stack.yml
SteffiStuffel: Yeah also thought about it, but i guess this will take longer and having a faster pipeline is more important.

PR #129, opened 2021-08-18
LUN-6: Add Theming

Taggotty:


src/components/__tests__/Button.spec.tsx
```diff
@@ -1,88 +1,93 @@
...
 import { BUTTONS_THEME } from '../../constants/data'
+import theme from '../../constants/theme'
+import { COLORS } from '../../constants/theme/colors'
+import { mocked } from 'ts-jest/utils'
+import wrapWithTheme from '../../testing/wrapWithTheme'
+
+type ButtonPropsType = ComponentProps<typeof Button>
+
+const expectStylesToMatch = (component: ShallowWrapper<ButtonPropsType>, expectedStyles: Record<string, any>): void => {
+  expect(component.dive().dive().dive().prop<Record<string, any>>('style')[0]).toEqual(
```
Taggotty, line 14: Not really pretty. Apparently there is a [package for testing styled components ](https://github.com/styled-components/jest-styled-components). But it is currently not working for native. I'll keep my eyes open on this.

tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "lib": ["es6"],                           /* Specify library files to be included in the compilation. */
     "allowJs": true,                          /* Allow javascript files to be compiled. */
     // "checkJs": true,                       /* Report errors in .js files. */
-    "jsx": "react-native",                    /* Specify JSX code generation: 'preserve', 'react-native', or 'react'. */
+    "jsx": "react",                           /* Specify JSX code generation: 'preserve', 'react-native', or 'react'. */
     // "declaration": true,                   /* Generates corresponding '.d.ts' file. */
```
Taggotty, line 11: Don't know why you would use react-native. In all the setup tutorials the use react and ts-jest basically also only works with react here...

SteffiStuffel (approved):
:tada:

src/components/AudioPlayer.tsx
```diff
@@ -91,8 +100,12 @@ const AudioPlayer = (props: AudioPlayerProps): ReactElement => {
 
   return (
     <StyledView>
-      <VolumeIcon testID='volume-button' disabled={disabled} onPress={() => handleSpeakerClick(document?.audio)}>
-        <VolumeUp fill={disabled ? COLORS.lunesBlackUltralight : isActive ? COLORS.lunesRed : COLORS.lunesRedDark} />
+      <VolumeIcon
+        theme={theme}
```
klinzo, line 104: Passing `theme` as prop should not be necessary, is it?

src/navigation/NavigationTypes.ts
```diff
@@ -1,5 +1,7 @@
 import { DocumentsType, DocumentType } from '../constants/endpoints'
 import { ExerciseKeyType, ResultType, SimpleResultType } from '../constants/data'
+import { SvgProps } from 'react-native-svg'
+import { ComponentType } from 'enzyme'
```
klinzo, line 4: Is importing the type from `enzyme` correct here?

.circleci/config.yml
```diff
@@ -197,7 +197,7 @@ jobs:
                 root: ios/output/gym
     bump_version:
         docker:
-            - image: circleci/node:12.16.3
+            - image: circleci/node:14.16.0
```
SteffiStuffel, line 200: Is it possible these updates make the build a lot slower? Before the e.g. check-step takes aroung 30sec and now more than a few minutes.

Taggotty: fixed after adding `--maxWorkers=1` to yarn test

PR #134, opened 2021-08-30
LUN-125: Articles in overview misplaced on bigger screens

ansamshtaiwi1:

### DIFF
--- a/src/routes/vocabulary-list/components/VocabularyListItem.tsx
+++ b/src/routes/vocabulary-list/components/VocabularyListItem.tsx
@@ -49,7 +49,7 @@ const StyledTitle = styled.Text`
   overflow: hidden;
   height: ${wp('5%')};
   text-align: center;
-  line-height: 18;
+
 `
 const Description = styled.Text` 
   font-size: ${wp('4%')};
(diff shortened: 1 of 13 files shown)


src/routes/vocabulary-list/components/VocabularyListItem.tsx
```diff
@@ -49,7 +49,7 @@ const StyledTitle = styled.Text`
   overflow: hidden;
   height: ${wp('5%')};
   text-align: center;
-  line-height: 18;
+
```
klinzo, line 52:
```suggestion
```

klinzo (approved):
Tested on Android, fixes the issue! Well done :)

Taggotty (approved):
Nice :)
Is there a reason why you chose LUN-13 as base branch? It's hard to find the relevant change under all those files.

YasminNajar:
@Taggotty I thought it was better to make the changes directly to the new styleComponents but now I believe it's better to do it on the main branch instead as it's a small change that can be easily applied. So I made a new branch from main and I will push it and close this one if that's okay.

Taggotty:
> @Taggotty I thought it was better to make the changes directly to the new styleComponents but now I believe it's better to do it on the main branch instead as it's a small change that can be easily applied. So I made a new branch from main and I will push it and close this one if that's okay.

Sounds like a great idea actually 🙈

klinzo:
@YasminNajar can we delete this branch or do you still need it?

PR #135, opened 2021-08-30
LUN-126: Recursive diciplines

SteffiStuffel:


src/components/ExerciseHeader.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from 'react'
-import Modal from './Modal'
+import ConfirmationModal from './Modal'
```
klinzo, line 2: Should we rename the file as well then?

src/constants/endpoints.ts
```diff
@@ -1,19 +1,12 @@
 import { Article } from './data'
 
-export interface ProfessionType {
+export interface DisciplineType {
```
klinzo, line 3: Do we have a glossary somewhere for these type of terms? I am really confused all the time what is a document, word, discipline, profession and so on. This is not an app specific topic though, but a project wide. Should I mention this in ~lunes-general?

src/hooks/useLoadDisciplines.ts
```diff
@@ -0,0 +1,40 @@
...
+  return { ...serverResponse, data: formattedServerResponse }
+}
+
+export const useLoadDisciplines = (parent: DisciplineType | null): ReturnType<DisciplineType[]> => {
+  let url: string
+  if (parent === null) {
+    url = ENDPOINTS.professions.all
+  } else {
+    url = `${!parent.isLeaf ? ENDPOINTS.professions.all : ENDPOINTS.subCategories.all}/${parent?.id}`
+  }
```
klinzo, line 37: To make it clearer you could extract the different urls to separate constants.
```suggestion
  const url = parent === null ? ENDPOINTS.professions.all : `${!parent.isLeaf ? ENDPOINTS.professions.all : ENDPOINTS.subCategories.all}/${parent?.id}`
```

src/hooks/useLoadDisciplines.ts
```diff
@@ -0,0 +1,40 @@
...
+  const disciplines: ReturnType<ServerResponse[]> = useLoadFromEndpoint(url)
```
klinzo, line 38:
```suggestion
  const disciplines = useLoadFromEndpoint<ServerResponse[]>(url)
```

src/hooks/useLoadDisciplines.ts
```diff
@@ -0,0 +1,40 @@
...
+    serverResponse.data?.map(item => ({
+      ...item,
+      numberOfChildren:
+        item.total_discipline_children > 0
+          ? item.total_discipline_children
+          : item.total_training_sets > 0
+          ? item.total_training_sets
+          : item.total_documents > 0
+          ? item.total_documents
+          : 0,
```
klinzo, line 25: Isn't this the same as:
```suggestion
        item.total_discipline_children ||  item.total_training_sets ||  item.total_documents
```

src/constants/endpoints.ts
```diff
@@ -41,10 +34,10 @@ export type DocumentsType = DocumentType[]
 
 export const ENDPOINTS = {
   professions: {
-    all: '/disciplines'
+    all: '/disciplines_by_level'
   },
   subCategories: {
-    all: '/training_set/:id'
+    all: '/training_set'
```
klinzo, line 40: Why is there no `id` anymore here?

I am really confused with the naming here as well, probably nothing for this PR though. I think we should be consistent. +1 for a glossary and consistent naming :D

src/hooks/useLoadDisciplines.ts
```diff
@@ -0,0 +1,40 @@
...
+      isLeaf: item.total_discipline_children <= 0
```
klinzo, line 26: Why `<=`?
```suggestion
      isLeaf: item.total_discipline_children === 0
```

src/navigation/Navigator.tsx
```diff
@@ -49,8 +49,9 @@ const Stack = createStackNavigator<RoutesParamsType>()
 
 const Navigator = (): JSX.Element => {
   const [isPressed, setIsPressed] = React.useState(false)
+  const [isHomeButtonPressed, setIsHomeButtonPressed] = React.useState(false)
```
klinzo, line 52:
```suggestion
  const [isPressed, setIsPressed] = React.useState<boolean>(false)
  const [isHomeButtonPressed, setIsHomeButtonPressed] = React.useState<boolean>(false)
```

src/navigation/Navigator.tsx
```diff
@@ -74,36 +86,51 @@ const Navigator = (): JSX.Element => {
...
+              !!route.params.extraParams.parentTitle
+            )
           }
           name='ProfessionSubcategory'
           component={ProfessionSubcategoryScreen}
         />
         <Stack.Screen
           options={({ route, navigation }: any) =>
-            defaultOptions(route.params.extraParams.disciplineTitle, BackButton, navigation, 'ProfessionSubcategory')
+            defaultOptions(route.params.extraParams.disciplineTitle, BackButton, navigation, true)
```
klinzo, line 104: Why did you remove the `ProfessionSubcategory` string? Not needed anymore?

src/routes/vocabulary-list/VocabularyListScreen.tsx
```diff
@@ -46,21 +46,6 @@ interface VocabularyListScreenPropsType {
...
-        <TouchableOpacity
-          onPress={() => navigation.popToTop()}
-          onPressIn={() => setIsHomeButtonPressed(true)}
-          onPressOut={() => setIsHomeButtonPressed(false)}
-          activeOpacity={1}>
-          {isHomeButtonPressed ? <HomeButtonPressed /> : <Home />}
-        </TouchableOpacity>
-      )
-    })
-  }, [navigation, isHomeButtonPressed])
```
klinzo, line 63: Nice refactoring :heart_eyes:

src/routes/ProfessionScreen.tsx
```diff
@@ -104,19 +91,19 @@ const ProfessionScreen = ({ navigation }: ProfessionScreenPropsType): ReactEleme
...
+  const handleNavigation = (item: DisciplineType): void => {
     setSelectedId(item.id)
     navigation.navigate('ProfessionSubcategory', {
       extraParams: {
-        disciplineID: item.id,
-        disciplineTitle: item.title,
-        disciplineIcon: item.icon
+        module: {
+          ...item
+        }
```
klinzo, line 106:
```suggestion
        module: item
```

src/routes/ProfessionSubcategoryScreen.tsx
```diff
@@ -77,51 +77,36 @@ interface ProfessionSubcategoryScreenPropsType {
...
     const selected = item.id === selectedId
     const descriptionStyle = selected ? styles.clickedItemDescription : styles.description
     const badgeStyle = selected ? styles.clickedItemBadgeLabel : styles.badgeLabel
+    const description = module.isLeaf
+      ? item.numberOfChildren === 1
+        ? labels.home.word
+        : labels.home.words
+      : item.numberOfChildren === 1
+      ? labels.home.unit
+      : labels.home.units
```
klinzo, line 109: Indentation

klinzo (changes requested):

src/constants/endpoints.ts
SteffiStuffel: Yes you are right, it is really confusing and we were talking about this already in lunes calls, but nothing already.
Currently the api has the endpoints:
- disciplines -> trainings_sets -> documents (wich have words)
In the app we have:
- profession -> subprofessions -> documents
Non-technical used naming:
- Bereichen -> Module

As with this PR in the frontend we merge disiplines and trainingssets into one type. I created a ticket and suggested the name "module" for it, but not really happy with it tough, so feed free to suggest an other name for disciplines/trainingssets (actually these are just like "folders" where the documents are inside)
https://issues.tuerantuer.org/browse/LUN-142

But feel also free to open a discussion in ~lunes-general

src/hooks/useLoadDisciplines.ts
SteffiStuffel: yes it is, and so much nicer :+1:

src/constants/endpoints.ts
SteffiStuffel: In useLoadDisicplines we append the id, but just appending it is easier, than searching for `:id` and replace it, isn't it?

src/navigation/Navigator.tsx
SteffiStuffel: Yes, not needed.

src/routes/ProfessionSubcategoryScreen.tsx
SteffiStuffel: its prettier :unamused: 
made it a one-liner

SteffiStuffel: oke is also not working...

src/constants/endpoints.ts
klinzo: Okay, thanks for explaining :+1: And `subCategories` in `endpoints.ts` is just a legacy of the old structure and basically the same as a training set, i.e. a module with documents as children? Will open a discussion :)

klinzo: Maybe just only using the term profession/disclipine would also work instead of renaming to module. That would probably make the purpose a little bit clearer.

SteffiStuffel: oke, i will adjust the ticket so it will be renamed to discipline.

klinzo:
Where do we have recursive disciplines already, i.e. where can this be tested?

src/constants/endpoints.ts
klinzo: I like that, that way we have nearly the same wording as the api as well :)

klinzo (approved):
Seems to work fine :tada: One thing I noticed: If I navigate to a 1st level discipline, e.g. `Metall, Elektro, Maschinen` the home button in the header is not displayed, is this because the behavior is the same as for the back button? Maybe I'd show it anyway as it may be confusing for users. But not really sure about that.

SteffiStuffel:
> Seems to work fine  One thing I noticed: If I navigate to a 1st level discipline, e.g. `Metall, Elektro, Maschinen` the home button in the header is not displayed, is this because the behavior is the same as for the back button? Maybe I'd show it anyway as it may be confusing for users. But not really sure about that.

This was a design decision, so would just keep it.

SteffiStuffel:
> Where do we have recursive disciplines already, i.e. where can this be tested?
> Edit: Seems to be testable at `Sonstige Bereiche` > `Test Sub 1`.
> Weird thing I noticed: It shows `28 Modules`, but there actually only two (or three if you count recursively). @SteffiStuffel probably an cms problem right?

Yeah. Lukes fixed this already, but wasn't deployed on test-cms, when you tested. I will check out if it now works fine.

PR #137, opened 2021-09-01
LUN-103: Article choose exercise add plural to last option

ansamshtaiwi1:


src/routes/choice-exercises/components/SingleChoiceListItem.tsx
```diff
@@ -175,7 +175,7 @@ const SingleChoiceListItem = ({
         </StyledArticleText>
       </StyledArticleBox>
       <StyledWord selected={selected} pressed={pressed} correct={showCorrect} delayPassed={delayPassed}>
-        {word}
+        {word} {article.id === 4 ? "(plural)" : " "}
```
klinzo, line 178: I think this needs some more work :)
The current way, there is always a space added and for articles different than `die (Plural)` even two spaces are added. There should only be a space if needed though, i.e. before the `(Plural)` text. Can you find a way to do so?

All string constants should be added to the `labels.json` in an appropriate place and used from there.

Also: I think your code style is a little bit off, we always use single quotes `'`. Are you using IntelliJ or something different like VSCode?

klinzo (changes requested):

src/routes/choice-exercises/components/SingleChoiceListItem.tsx
YasminNajar: Hello, thank you for the feedback. For the first point, would "{word}{article.id === 4 && '(Plural)'}" solve the issue ?

klinzo: Yes, I think so. However, I would not put the space into the labels.json. Not sure whether to only put `Plural` or `(Plural)`. Probably even the first one without brackets.

SteffiStuffel: In the ticket it was explicitly defines as `(Plural)` with brackets.

klinzo: Yes, but my point was whether to add it to labels.json with brackets or without. The user should always see brackets.

SteffiStuffel: ahh, ok I get it. Would put it to labels.json without brackets, so we can reuse it if needed.

klinzo: I agree :)

YasminNajar:
Changes have been done, let me know if you want me to change anything else :)

src/constants/labels.json
```diff
@@ -66,5 +66,8 @@
       "cancelExercise": "Übung beenden",
       "of": "von"
     }
+  },
+  "pluralArticle": {
+    "plurals": "Plural"
   }
```
SteffiStuffel, line 72: I think you can put this insede general and without nesting.

I suggest:

```
  "general": {
    "header": {
      "overview": "Übersicht",
      "overviewExercises": "Übersicht der Übungen",
      "cancelExercise": "Übung beenden",
      "of": "von"
    }
 "plural": "Plural"
  },
```

SteffiStuffel (approved):
Tested, looks good :tada:

PR #138, opened 2021-09-01
LUN-143: Fix ci-lint checking

SteffiStuffel:


klinzo (approved):

package.json
```diff
@@ -9,7 +9,7 @@
     "start": "react-native start",
     "test": "jest --silent --verbose",
     "test:update": "yarn run test -u",
-    "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
+    "lint": "eslint .",
     "lint:fix": "yarn lint --fix",
     "lint:ci": "yarn run eslint --quiet --format junit -o reports/lint/junit-lint.xml",
```
Taggotty, line 14: then do you still need this one?

Taggotty (approved):

package.json
SteffiStuffel: No I guess not, but in Integreat it is also still available.

PR #140, opened 2021-09-07
LUN-144: iOS Production Upload

SteffiStuffel:


.circleci/src/workflows/prod_delivery.yml
```diff
@@ -7,18 +7,18 @@ jobs:
...
-  #      requires:
-  #        - check
-  #        - bump_version
+  - build_ios:
+      requires:
+        - check
+        - bump_version
   - deliver_android:
-      production_delivery: true # TODO test in mid of august
+      production_delivery: true # TODO test in mid of september
```
klinzo, line 15: What is meant by this comment? Aren't you on vacation then?

android/fastlane/Fastfile
```diff
@@ -126,9 +126,12 @@ platform :android do
             version_code: version_code,
             version_name: version_name,
             package_name: "app.lunes",
-            metadata_path: "./android/fastlane/lunes/metadata",
+            metadata_path: "./fastlane/metadata",
```
klinzo, line 129: Shouldn't it be `./android/fastlane/metadata`?

klinzo (approved):

android/fastlane/Fastfile
SteffiStuffel: no, because the working directory is /android/
Tested it and is working :)

.circleci/src/workflows/prod_delivery.yml
SteffiStuffel: No, I will try to deploy the app with the prod_delivery workflow after my vacations. This comment is just a reminder for me, that these workflow/jobs haven't not be tested and I need to do this.

android/fastlane/Fastfile
klinzo: Ah okay. Any specific reason why you choose android as working directory but ios not?

SteffiStuffel:
> you already tested this right?

No not yet. Can't make a production deploy right now, because of api-changes.

android/fastlane/Fastfile
SteffiStuffel: For clarification: The release-notes for android are working, the prod_delivery is not tested as not possible right now, therefore the comment to do this in mid of september.

SteffiStuffel: yeah, because this is an android job? For ios the working_directory is `ios`

klinzo: But the metadata path for ios includes `ios` while the metadata path for android does not include `android`, that's all I wanted to say :D

SteffiStuffel: oke, will fix this.

PR #145, opened 2021-09-22
LUN-120 Almost correct word

ansamshtaiwi1:


Taggotty (changes requested):
Please add release notes for this.

Taggotty:
> Please add release notes for this.

Nevermind. I just realized that our release notes are in German. So you of course don't have to do it.

Taggotty (approved):

ansamshtaiwi1:
If the entered word is almost correct, the previous input is deleted. Now it is more user-friendly that input does not deleted and could be improved instead.

PR #148, opened 2021-09-27
LUN-152: Fix audio files

SteffiStuffel:

### DIFF
--- a/src/components/AudioPlayer.tsx
+++ b/src/components/AudioPlayer.tsx
@@ -4,6 +4,8 @@ import Tts, { TtsError } from 'react-native-tts'
 import { VolumeUp } from '../../assets/images'
 import { DocumentType } from '../constants/endpoints'
 import styled from 'styled-components/native'
+import { EventListenerCallback } from '@react-navigation/native'
+import { EmitterSubscription } from 'react-native'
 
 export interface AudioPlayerProps {
   document: DocumentType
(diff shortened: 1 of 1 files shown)


src/components/AudioPlayer.tsx
```diff
@@ -15,14 +16,19 @@ const StyledView = styled.View`
...
   width: 40px;
   height: 40px;
   border-radius: 50px;
-  background-color: ${props => props.theme.colors.lunesFunctionalIncorrectDark};
+  background-color: ${props =>
+    props.disabled
+      ? props.theme.colors.lunesBlackUltralight
+      : props.isActive
+      ? props.theme.colors.lunesRed
+      : props.theme.colors.lunesRedDark};
```
klinzo, line 31: You are doing it as well :astonished: Nested ternary expressions, tzzz :smile:

klinzo (commented):
Playing audio files does not work at all if there is no tts engine available (which is the case for me). I think it also makes no sense to initialize both all the time. I created a PR hopefully fixing the issue with no tts engine which also tidies stuff up and only initializes the necessary audio: #149

src/components/AudioPlayer.tsx
```diff
@@ -83,7 +89,10 @@ const AudioPlayer = (props: AudioPlayerProps): ReactElement => {
     if (isInitialized) {
       setIsActive(true)
       if (audio) {
-        SoundPlayer.playUrl(document?.audio)
+        SoundPlayer.loadUrl(audio)
+        _onFinishedLoadingSubscription = SoundPlayer.addEventListener('FinishedLoadingURL', ({ success, url }) => {
+          SoundPlayer.play()
+        })
```
klinzo, line 95: Not necessary to do this each time you play an audio.

src/components/AudioPlayer.tsx
SteffiStuffel: Its not mine, i just copied :smile:

PR #155, opened 2021-10-04
LUN-166: Unit test for MenuItem

SteffiStuffel:

### DIFF
--- a/src/components/__tests__/MenuItem.spec.tsx
+++ b/src/components/__tests__/MenuItem.spec.tsx
@@ -1,5 +1,4 @@
 import { fireEvent, render } from '@testing-library/react-native'
-import { shallow } from 'enzyme'
 import React from 'react'
 import { Text } from 'react-native'
 
@@ -12,7 +11,7 @@ describe('Components', () => {
       selected: false,
       icon: '',
       title: 'MenuItemTitle',
-      children: <Text>description</Text>,
+      children: <Text>Text of children</Text>,
       onPress: () => {}
     }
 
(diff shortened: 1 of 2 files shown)


src/components/__tests__/MenuItem.spec.tsx
```diff
@@ -35,26 +33,32 @@ describe('Components', () => {
...
       }
 
-      const component = shallow(<MenuItem {...menuItemProps} />)
-      expect(component.find('[testID="arrow"]').props().fill).toBe(COLORS.lunesBlack)
+      const { getByTestId, getByText } = render(<MenuItem {...menuItemProps} />)
+      const arrowIcon = getByTestId('arrow')
+      expect(arrowIcon.props.fill).toBe(COLORS.lunesBlack)
+      const title = getByText('MenuItemTitle')
+      // @ts-expect-error
+      expect(title._fiber.pendingProps.style[0].color).toBe(COLORS.lunesGreyDark)
```
SteffiStuffel, line 61: Any better idea how to test for styles?

Taggotty: See here: https://github.com/digitalfabrik/lunes-app/blob/7fe3012bcd9ee05ad5f428b0bd3b325a58e8e157/src/components/__tests__/MenuItem.spec.tsx
I accidentally refactored MenuItem already, as the tests were failing due to StyledComponents and I didn't know you were already working on it. As it was not part of my issue though I reverted and let you have the honor. Still, maybe the changes can help you here :)

Taggotty: how about title.instance.props.style[0].color? :)

src/components/__tests__/MenuItem.spec.tsx
```diff
@@ -35,26 +33,32 @@ describe('Components', () => {
         title: 'Menu item title'
       }
```
Taggotty, line 34: personally I prefer having a render function in tests instead of overriding the default props inside the tests. This also avoids having to change all tests if you e.g. have to wrap the component into a themeProvider and stuff. But as said only personal preference

Taggotty (approved):

PR #158, opened 2021-10-05
LUN-157: Fix styling of writing exercise

klinzo:

### DIFF
--- a/src/routes/write-exercise/WriteExerciseScreen.tsx
+++ b/src/routes/write-exercise/WriteExerciseScreen.tsx
@@ -6,6 +6,7 @@ import { KeyboardAwareScrollView } from 'react-native-keyboard-aware-scroll-view
 import { heightPercentageToDP as hp } from 'react-native-responsive-screen'
 import styled from 'styled-components/native'
 
+import AudioPlayer from '../../components/AudioPlayer'
 import ExerciseHeader from '../../components/ExerciseHeader'
 import ImageCarousel from '../../components/ImageCarousel'
 import { DocumentType } from '../../constants/endpoints'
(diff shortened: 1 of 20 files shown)


src/routes/write-exercise/components/__tests__/__snapshots__/ArticleMissingPopoverContent.spec.tsx.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
```
SteffiStuffel, line 2: You can just remove it, we are planning to remove all snapshot tests.

SteffiStuffel (approved):
Really nice refactoring! :rocket: :tada:

src/routes/write-exercise/components/__tests__/__snapshots__/ArticleMissingPopoverContent.spec.tsx.snap
klinzo: Will just keep it, not in scope of PR.

src/routes/write-exercise/WriteExerciseScreen.tsx
```diff
@@ -56,38 +58,41 @@ const WriteExerciseScreen = ({ navigation, route }: WriteExerciseScreenPropsType
...
-              documents={documents}
-              finishExercise={finishExercise}
-              tryLater={tryLater}
-              trainingSet={trainingSet}
-              disciplineTitle={disciplineTitle}
-            />
-          </KeyboardAwareScrollView>
-        </>
+      {documents && document ? (
+        <ScrollView
```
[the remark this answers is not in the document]
klinzo: I actually like that the content is moved to the top and the image is only displayed partly if opening the keyboard as more of the buttons are visible which makes it easier for users to solve the exercise imo.

I fixed the text cut off issue though, wanted to do it in a separate issue because I struggled with it but all good now I hope. I also reduced the margins a little bit so that more content can be shown.

Won't do the keyboard here if there is a separate issue/was there before.

PR #165, opened 2021-10-14
LUN-140: Refactor navigation types

SteffiStuffel:


src/services/AsyncStorage.ts
```diff
@@ -1,41 +1,21 @@
...
-    [trainingSet: string]: {
-      [word: string]: DocumentResultType
-    }
+  [disciplineId: number]: {
+    [word: string]: DocumentResultType
   }
 }
 
-export const setSession = async (session: SessionType): Promise<void> => {
+export const setSession = async (session: any): Promise<void> => {
```
klinzo, line 14: Why did you remove the type? Same below

src/services/AsyncStorage.ts
```diff
@@ -1,41 +1,21 @@
...
-  }
-  retryData?: { data: DocumentsType }
-}
-
 export interface ExerciseType {
-  [disciplineTitle: string]: {
-    [trainingSet: string]: {
-      [word: string]: DocumentResultType
-    }
+  [disciplineId: number]: {
```
klinzo, line 9: Why is the training set id not needed anymore?

src/routes/write-exercise/WriteExerciseScreen.tsx
```diff
@@ -54,7 +57,13 @@ const WriteExerciseScreen = ({ navigation, route }: WriteExerciseScreenPropsType
     AsyncStorage.clearSession().catch(e => console.error(e))
     setCurrentDocumentNumber(0)
     setNewDocuments(null)
-    navigation.navigate('InitialSummary', { extraParams: { ...extraParams, results: [] } })
+    navigation.navigate('InitialSummary', {
+      result: {
+        discipline: { ...discipline },
+        results: [],
```
klinzo, line 63: Was that way before already but why don't we need to pass results?

PR #166, opened 2021-10-14
LUN-121: Random order of vocabulary

hepperle00:

### DIFF
--- a/src/hooks/useLoadDocuments.ts
+++ b/src/hooks/useLoadDocuments.ts
@@ -1,6 +1,7 @@
 import { ARTICLES } from '../constants/data'
 import { DocumentType, ENDPOINTS } from '../constants/endpoints'
 import useLoadFromEndpoint, { ReturnType } from './useLoadFromEndpoint'
+import { shuffleArray } from "../services/helpers";
 
 export interface AlternativeWordTypeFromServer {
   article: number
@@ -33,6 +34,15 @@ const useLoadDocuments = (trainingSetId: number): ReturnType<DocumentType[]> => 
   const url = ENDPOINTS.documents.replace(':id', `${trainingSetId}`)
   const documents: ReturnType<DocumentTypeFromServer[]> = useLoadFromEndpoint(url)
   return transformArticle(documents)
+}
+
+export const useLoadDocumentsRandomOrder = (trainingSetId: number): ReturnType<DocumentType[]> => {
+  const documents: ReturnType<DocumentType[]> = useLoadDocuments(trainingSetId)
+  if (!documents.data) {
+    return documents
+  }
+  const shuffledDocuments = shuffleArray(documents.data)
+  return { ...documents, data: shuffledDocuments }
 }
 
 export default useLoadDocuments
--- a/src/hooks/useLoadDocuments.ts
+++ b/src/hooks/useLoadDocuments.ts
@@ -1,7 +1,7 @@
 import { ARTICLES } from '../constants/data'
 import { DocumentType, ENDPOINTS } from '../constants/endpoints'
+import { shuffleArray } from '../services/helpers'
 import useLoadFromEndpoint, { ReturnType } from './useLoadFromEndpoint'
-import { shuffleArray } from "../services/helpers";
 
 export interface AlternativeWordTypeFromServer {
   article: number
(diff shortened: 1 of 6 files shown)


src/hooks/useLoadDocuments.ts
```diff
@@ -35,4 +36,13 @@ const useLoadDocuments = (trainingSetId: number): ReturnType<DocumentType[]> =>
   return transformArticle(documents)
 }
 
+export const useLoadDocumentsRandomOrder = (trainingSetId: number): ReturnType<DocumentType[]> => {
+  const documents: ReturnType<DocumentType[]> = useLoadDocuments(trainingSetId)
+  if (!documents.data) {
```
SteffiStuffel, line 41: you can simplify this to `return documents.data ? { ...documents, data: shuffleArray(documents.data) } : documents`

src/services/helpers.ts
```diff
@@ -19,3 +19,15 @@ export const getArticleColor = (article: Article): string => {
       return COLORS.lunesArtikelDer
   }
 }
+
+/* Randomize array using Durstenfeld shuffle algorithm with O(n) complexity */
+export const shuffleArray = <T>(array: T[]): T[] => {
+  const shuffledArray = array.slice()
+  for (let i = shuffledArray.length - 1; i > 0; i--) {
```
klinzo, line 26: Better to use build in array functions like `forEach`

src/services/helpers.ts
```diff
@@ -19,3 +19,15 @@ export const getArticleColor = (article: Article): string => {
       return COLORS.lunesArtikelDer
   }
 }
+
+/* Randomize array using Durstenfeld shuffle algorithm with O(n) complexity */
```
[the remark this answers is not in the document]
maxammann: It could be transformed to JSDoc.

src/services/helpers.ts
maxammann: It iterates backwards. Even through it should not make a difference, this is the original algorithm.

There is a shorter version of this:

```
function shuffleArray(array) {
    for (let i = array.length - 1; i > 0; i--) {
        const j = Math.floor(Math.random() * (i + 1));
        [array[i], array[j]] = [array[j], array[i]];
    }
}
```
https://stackoverflow.com/a/12646864/1763110

src/services/helpers.ts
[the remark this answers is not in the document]
maxammann: Maybe a link from where this is copied is nice

src/routes/write-exercise/__tests__/WriteExerciseScreen.spec.tsx
```diff
@@ -28,9 +28,15 @@ jest.mock('react-native/Libraries/LogBox/Data/LogBoxData')
 
 describe('WriteExerciseScreen', () => {
   beforeEach(() => {
+    // mocking math makes normally random order of documents predictable
+    jest.spyOn(Math, 'random').mockReturnValue(0.9)
     jest.clearAllMocks()
   })
 
+  afterEach(() => {
+    jest.spyOn(Math, 'random').mockRestore()
```
maxammann, line 37:
```suggestion
    jest.restoreAllMocks()
```
not tested but should do the same

src/routes/choice-exercises/WordChoiceExerciseScreen.tsx
```diff
@@ -16,7 +16,7 @@ interface WordChoiceExerciseScreenPropsType {
 const WordChoiceExerciseScreen = ({ navigation, route }: WordChoiceExerciseScreenPropsType): ReactElement | null => {
   const { extraParams } = route.params
   const { trainingSetId } = extraParams
-  const { data: documents, loading } = useLoadDocuments(trainingSetId)
+  const { data: documents, loading } = useLoadDocumentsRandomOrder(trainingSetId)
```
maxammann, line 19: Wouldn't it be simpler to just apply the shuffle function here instead of introducing a new hook?
Maybe the useLoadDocuments hook could also just take a paramter `shuffle`

maxammann (approved):

src/services/helpers.ts
maxammann: @hepperle00 looks good, there is one additional ;

PR #169, opened 2021-10-19
LUN-188: Adjust home screen header

ztefanie:


src/routes/HomeScreen.tsx
```diff
@@ -118,23 +117,19 @@ const HomeScreen = ({ navigation, customDisciplines }: HomeScreenPropsType): JSX
   }
 
   return (
```
maxammann, line 119: I think using a SafeArea here is important here to adjust for different devices. For me the previous solution looks better even though it obvisouly has a bug somehwere

PR #179, opened 2021-11-02
LUN 69: Play TTS for alternative solutions

klinzo:

### DIFF
--- a/src/routes/write-exercise/components/WriteExercise.tsx
+++ b/src/routes/write-exercise/components/WriteExercise.tsx
@@ -5,6 +5,8 @@ import stringSimilarity from 'string-similarity'
 import styled from 'styled-components/native'
 
 import { CloseIcon } from '../../../../assets/images'
+import AudioPlayer from '../../../components/AudioPlayer'
+import ImageCarousel from '../../../components/ImageCarousel'
 import { ExerciseKeys, SimpleResultType } from '../../../constants/data'
 import { DocumentType } from '../../../constants/endpoints'
 import labels from '../../../constants/labels.json'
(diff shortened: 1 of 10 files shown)


src/routes/write-exercise/components/WriteExercise.tsx
```diff
@@ -195,47 +192,60 @@ const AnswerSection = ({
...
-            onFocus={() => setIsFocused(true)}
-            onBlur={() => setIsFocused(false)}
-            onSubmitEditing={checkEntry}
+    <>
+      <ImageCarousel images={document.document_image} />
+      <AudioPlayer
+        document={document}
+        disabled={!result || result === 'similar'}
+        submittedAlternative={submittedAlternative}
+      />
```
klinzo, line 208: This is added, the rest stayed nearly the same.

klinzo:
Please test this. @SteffiStuffel where do we have alternatives?

SteffiStuffel (changes requested):
Just tested this and get errors /unexpected behaviour:

When entering the correct (not alternative) solution, I get an no tts-engine-installed error, although it is installed, but also it is not needed.

When entering a correct alternative solution, the normal audio is played.

E.g. Zollstock has alternative solution Meterstab, it is in the Test-Module (Test -> Level2 -> Test)

klinzo:
> Just tested this and get errors /unexpected behaviour:
> 
> When entering the correct (not alternative) solution, I get an no tts-engine-installed error, although it is installed, but also it is not needed.
> 
> When entering a correct alternative solution, the normal audio is played.
> 
> E.g. Zollstock has alternative solution Meterstab, it is in the Test-Module (Test -> Level2 -> Test)

Should be fixed. Thanks!

SteffiStuffel (approved):
Tested again, works as expected

SteffiStuffel:
Can you add a release-note as well, please?

f1sh1918 (approved):
Tested on ios native, nice work :)

PR #181, opened 2021-11-11
LUN-154: Make screen scrollable

f1sh1918:

Tested on Android (native & emulator), tested on ios native
Check comments below

**Updated:**
Now i calculate the screen height by an hook.
Added dismiss keyboard by clicking on "Try Later" so the switch is better to recognize for the user


src/routes/write-exercise/WriteExerciseScreen.tsx
```diff
@@ -66,7 +68,7 @@ const WriteExerciseScreen = ({ navigation, route }: WriteExerciseScreenPropsType
 
       <ServerResponseHandler error={error} loading={loading} refresh={refresh}>
         {documents && document && (
-          <ScrollView contentContainerStyle={{ flex: 1 }} keyboardShouldPersistTaps='always'>
+          <KeyboardAwareScrollView contentContainerStyle={{ height }}>
```
f1sh1918, line 71: By setting `flexGrow: 1` i'm getting an issue on android by wrapping the ImageCarousel. Maybe someone finds a nicer solution. On IOS everything works as expected :/

According to @maxammann 's comment on the closed pr. 
I tested:
https://reactnative.dev/docs/keyboardavoidingview in combination with a `ScrollView`
It causes strange scrolling issues each time i open the keyboard, the scroll view increased. So i moved back to this library

src/routes/write-exercise/WriteExerciseScreen.tsx
```diff
@@ -23,6 +24,7 @@ const WriteExerciseScreen = ({ navigation, route }: WriteExerciseScreenPropsType
   const { id } = discipline
   const [currentDocumentNumber, setCurrentDocumentNumber] = useState(0)
   const [newDocuments, setNewDocuments] = useState<DocumentType[] | null>(null)
+  const { height } = useWindowDimensions()
```
SteffiStuffel, line 27: Normally we use the `react-native-responsive-screen` library to get window-dimension. See it makes sense to use `useWindowDimenstions()` here, but not sure though because it is somehow "inconsistent". 
What do you think?
(@klinzo)

SteffiStuffel (commented):
This ticket was not really ready (to-be-discussed label and comment in the description) as we wanted some input on a designers point of view, if we should somehow resize the image or something if opening the keyboard.
As it is still better than before, I think it would be oke to merge it.

But the scrolling on ios seems a bit of, there is to much white-space on the bottom.
LUN-154

f1sh1918:
> This ticket was not really ready (to-be-discussed label and comment in the description) as we wanted some input on a designers point of view, if we should somehow resize the image or something if opening the keyboard. As it is still better than before, I think it would be oke to merge it.
> 
> But the scrolling on ios seems a bit of, there is to much white-space on the bottom. LUN-154

Yes indeed the problem in this case is the mentioned `flex` issue on Android. If we want the correct height, i would have to reduce the height by navigation header and notification bar

src/routes/write-exercise/WriteExerciseScreen.tsx
klinzo: No idea about `react-native-responsive-screen` tbh, never worked with it yet.

klinzo (commented):
+1 for reducing the height of the content

One other thing: When closing the keyboard and one scrolled down, the screen jumps a little bit. This is especially annoying if I try to click on a button (e.g. `Später versuchen`). There should be options to pass the click though such that you don't have to click twice.

src/routes/write-exercise/WriteExerciseScreen.tsx
Taggotty: Just using the ScrollView without a wrapper and setting `flexGrow` seems to work for me on android.

f1sh1918: its just working for emulator. thats what i experienced

klinzo (approved):

src/routes/write-exercise/WriteExerciseScreen.tsx
f1sh1918: i added a custom hook for getting screen size without navbar and statusbar. pls check

src/hooks/useScreenHeight.ts
```diff
@@ -0,0 +1,17 @@
...
+ * This hook returns the screen height without navigation header and notification bar
+ * For IOS there are no proper libraries or ways to receive status bar height, so the default height of ios is used
+ */
+
+const DEFAULT_IOS_STATUS_BAR_HEIGHT = 20
+const useScreenHeight = (): number => {
+  const { height: deviceHeight } = useWindowDimensions()
+  const statusBarHeight = StatusBar.currentHeight ?? DEFAULT_IOS_STATUS_BAR_HEIGHT
+  const height = deviceHeight - useHeaderHeight() - statusBarHeight
+  return height
```
klinzo, line 14:
```suggestion
  return deviceHeight - useHeaderHeight() - statusBarHeight
```

src/routes/write-exercise/WriteExerciseScreen.tsx
```diff
@@ -68,7 +72,10 @@ const WriteExerciseScreen = ({ navigation, route }: WriteExerciseScreenPropsType
 
       <ServerResponseHandler error={error} loading={loading} refresh={refresh}>
         {documents && document && (
-          <KeyboardAwareScrollView contentContainerStyle={{ height }}>
+          <KeyboardAwareScrollView
+            contentContainerStyle={{ height }}
+            keyboardShouldPersistTaps='always'
+            innerRef={(ref: Element) => (scrollRef.current = ref as ScrollView)}>
```
klinzo, line 78: Do you need to pass a function? Isn't it normally enough to pass the ref?

src/routes/write-exercise/WriteExerciseScreen.tsx
```diff
@@ -39,6 +41,8 @@ const WriteExerciseScreen = ({ navigation, route }: WriteExerciseScreenPropsType
     if (documents !== null) {
       setNewDocuments(moveToEnd(documents, currentDocumentNumber))
     }
+    scrollRef.current?.scrollTo({ y: 0, x: 0, animated: true })
+    Keyboard.dismiss()
```
klinzo, line 45: Why only for `tryLater`?

SteffiStuffel (approved):
Still the issue with too much whitespace, when scrolling down on both ios and android emulator.

PR #182, opened 2021-11-13
LUN-201: Refactor fetching logic and AddCustomDisciplineScreen

klinzo:


klinzo:
Also fixes short showing of error when adding a custom discipline.

klinzo:
@SteffiStuffel it would now be really easy to replace the hoc `withCustomDisciplines` with a equivalent hook or function :)

src/hooks/useLoadDocuments.ts
```diff
@@ -17,23 +17,24 @@ export interface DocumentTypeFromServer {
...
   const url = ENDPOINTS.documents.replace(':id', `${discipline.id}`)
-  const documents: ReturnType<DocumentTypeFromServer[]> = useLoadFromEndpoint(url, discipline.apiKey)
-  return transformArticle(documents)
+
+  const response = await getFromEndpoint<DocumentTypeFromServer[]>(url, discipline.apiKey)
+  return formatServerResponse(response)
 }
 
+const useLoadDocuments = (discipline: Discipline): ReturnType<DocumentType[]> =>
+  useLoadFromEndpoint(loadDocuments, discipline)
```
SteffiStuffel, line 38: Is this "PingPong" really needed?
-> In useLoadDocuments call method from useLoadFromEndpoint
-> in useLoadFromEndpoint call method (loadDocuments) from useLoadDocuments
-> In useLoadDocuments (loadDocuments-method) call method (getFromEndpoint) from useLoadFromEndpoint

Would it maybe make more sense to just pass the formatting-method to useLoadFromEndpoint?

tsconfig.json
```diff
@@ -1,60 +1,21 @@
...
-    // "outFile": "./",                       /* Concatenate and emit output to single file. */
-    // "outDir": "./",                        /* Redirect output structure to the directory. */
-    // "rootDir": "./",                       /* Specify the root directory of input files. Use to control the output directory structure with --outDir. */
-    // "removeComments": true,                /* Do not emit comments to output. */
-    "noEmit": true,                           /* Do not emit outputs. */
-    // "incremental": true,                   /* Enable incremental compilation */
-    // "importHelpers": true,                 /* Import emit helpers from 'tslib'. */
-    // "downlevelIteration": true,            /* Provide full support for iterables in 'for-of', spread, and destructuring when targeting 'ES5' or 'ES3'. */
-    "isolatedModules": true,                  /* Transpile each file as a separate module (similar to 'ts.transpileModule'). */
+    "target": "esnext",
```
SteffiStuffel, line 4: Just deleted the comments, right?

SteffiStuffel (approved):
Really nice PR! :tada: 

So now it is possible to either use the hook or the promise-returning-method if doing a server request? 

Thanks for adding the ts:check, had this in my mind for weeks and never did it :thinking: 

And nice to have tests now :rocket:

SteffiStuffel:
> @SteffiStuffel it would now be really easy to replace the hoc `withCustomDisciplines` with a equivalent hook or function :) If you want I could create a draft PR for that.

I'm already working (and almost finished) with this.

klinzo:
> So now it is possible to either use the hook or the promise-returning-method if doing a server request?
> 
Yes, it should be possible for all requests.

src/hooks/useLoadDocuments.ts
klinzo: I think that is not as it is, there should be no function call of `loadDocuments` in `useLoadFromEndpoint` :thinking: So we just have dependencies from `useLoadDocuments` to `useLoadFromEndpoint` (and their other functions), so only one way.

tsconfig.json
klinzo: I also added `skipLibCheck` and extended the `types` array with `styled-components-react-native`

klinzo: But yes, mostly just deleted comments.

src/hooks/useLoadDocuments.ts
SteffiStuffel: yes, got this, but it gets passed to it and is then executed. My point is more about the execution order and not the dependencies.

klinzo: A I see. Yeah not sure about a better way. If we just pass the formatting method it won't be possible to use the fetch logic without hook.

Taggotty (approved):

Taggotty:
Tested on Android emulator. No issues encountered: 👍🏼

PR #183, opened 2021-11-14
LUN-141: Remove exercise results and sessions from async storage

klinzo:

### DIFF
--- a/src/components/ConfirmationModal.tsx
+++ b/src/components/ConfirmationModal.tsx
@@ -9,7 +9,6 @@ import { CloseIcon } from '../../assets/images'
 import { BUTTONS_THEME } from '../constants/data'
 import labels from '../constants/labels.json'
 import { RoutesParamsType } from '../navigation/NavigationTypes'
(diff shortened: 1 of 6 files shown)


SteffiStuffel (approved):
Tested in Android. Works perfectly fine.

Taggotty (approved):
What will happen with the storage data that was already created?

And furthermore the color of the label of the `Eingabe Überprüfen` button seems off

klinzo:
> What will happen with the storage data that was already created?

Good question, right now it will just be kept. I think that is fine as it shouldn't be a huge amount of data and does not hurt. Other opinions?
 
> And furthermore the color of the label of the `Eingabe Überprüfen` button seems off

Is this something new? There is already an issue for this (if I understood you correctly): https://issues.tuerantuer.org/browse/LUN-199

SteffiStuffel (approved):

klinzo:
@Taggotty if you think this should be handled differently/be removed, please open an issue :)

PR #186, opened 2021-11-17
LUN-161: Test for confirmation modal

SteffiStuffel:


klinzo (approved):

src/components/__tests__/ConfirmationModal.spec.tsx
```diff
@@ -1,45 +1,43 @@
...
-      const modalProps: ConfirmationModalPropsType = {
-        ...defaultModalProps,
-        visible: true
-      }
+    it('should close on cancel button click', () => {
+      const { getByText } = render(<ConfirmationModal {...defaultModalProps} />, { wrapper: wrapWithTheme })
+      const cancelButton = getByText('cancel')
+      fireEvent.press(cancelButton)
+      expect(setVisible).toBeCalledWith(false)
+      expect(confirmationAction).not.toBeCalled()
```
klinzo, line 33: I think you could assert that `navigation.navigate` was called as well with the right params.

SteffiStuffel: Navigation is not called anymore. The `ConfirmationModal` is now more generic and can be used for different purposes, e.g. also the confirmation when deleting a module, where no navigation is needed. Therefore the confirmation-action is now passed as a prop and can (but needn't) trigger a navigation

klinzo:
Still one review needed :)

SteffiStuffel:
@Taggotty

Taggotty (approved):

PR #187, opened 2021-11-17
LUN-133: Wrong words should be repeated

SteffiStuffel:


release-notes/unreleased/main.yml
```diff
@@ -0,0 +1,6 @@
+issue_key:
+show_in_stores: true
+platforms:
+  - android
+  - ios
+de: Wiederholungsmechanismus bei fehlerhaften Eingaben für alle Aufgabentypen
```
klinzo, line 6: `Wiederholungsmechanismus` is quite a difficult word, maybe:
```suggestion
de: Falsche Eingaben werden nun bei allen Übungen wiederholt
```

src/routes/choice-exercises/__tests__/ArticleChoiceExerciseScreen.spec.tsx
```diff
@@ -57,29 +62,52 @@ describe('ArticleChoiceExerciseScreen', () => {
   it('should allow to skip an exercise and try it out later', () => {
     mockUseLoadFromEndpointWitData(testDocuments)
 
-    const { getByText } = render(<ArticleChoiceExerciseScreen route={route} navigation={navigation} />, {
+    const { getByText, getAllByText } = render(<ArticleChoiceExerciseScreen route={route} navigation={navigation} />, {
       wrapper: wrapWithTheme
     })
 
+    expect(getAllByText('Helm')[0]).toBeDefined()
```
klinzo, line 69: Why do you use `getAllByText` here and assert just on the first element anyway? Same below

src/routes/choice-exercises/__tests__/WordChoiceExerciseScreen.spec.tsx
```diff
@@ -79,41 +90,26 @@ describe('WordChoiceExerciseScreen', () => {
...
-  it('should not allow to skip last document', () => {
-    mockUseLoadFromEndpointWitData(testDocuments)
-    const { queryByText, getByText } = render(<WordChoiceExerciseScreen route={route} navigation={navigation} />, {
-      wrapper: wrapWithTheme
-    })
-
-    expect(queryByText(labels.exercises.tryLater)).not.toBeNull()
-    fireEvent(getByText('Arbeitshose'), 'pressOut')
+    let correctAnswer = getByText('Helm')
+    fireEvent(correctAnswer, 'pressOut')
```
klinzo, line 101:
```suggestion
    fireEvent.press(correctAnswer)
```

src/routes/choice-exercises/__tests__/WordChoiceExerciseScreen.spec.tsx
```diff
@@ -79,41 +90,26 @@ describe('WordChoiceExerciseScreen', () => {
...
 
-  it('should not allow to skip last document', () => {
-    mockUseLoadFromEndpointWitData(testDocuments)
-    const { queryByText, getByText } = render(<WordChoiceExerciseScreen route={route} navigation={navigation} />, {
-      wrapper: wrapWithTheme
-    })
-
-    expect(queryByText(labels.exercises.tryLater)).not.toBeNull()
-    fireEvent(getByText('Arbeitshose'), 'pressOut')
+    let correctAnswer = getByText('Helm')
```
klinzo, line 100: You could just inline it to avoid defining a constant at all

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
```diff
@@ -107,36 +107,63 @@ const ChoiceExerciseScreen = ({
...
+      updateResult(result)
     }
     setTimeout(() => {
       setDelayPassed(true)
     }, correctAnswerDelay)
   }
 
+  const updateResult = (result: DocumentResultType): void => {
+    const indexOfCurrentResult = results.findIndex(result => result.id === currentDocument?.id)
+    indexOfCurrentResult !== -1 ? (results[indexOfCurrentResult] = result) : setResults([...results, result])
```
klinzo, line 140: You could use a map for this as well:
```suggestion
    setResults(results => results.map(result => result.id === newResult.id ? newResult : result))
```

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
```diff
@@ -107,36 +107,63 @@ const ChoiceExerciseScreen = ({
...
+  }
+
+  const getRetryInfoOfCurrent = (): { nthRetry: number; needsToBeRepeated: boolean } => {
+    const indexOfCurrent = results.findIndex(result => result.id === currentDocument?.id)
+    const nthRetry = indexOfCurrent === -1 ? 0 : results[indexOfCurrent].numberOfTries
+    const needsToBeRepeated =
+      nthRetry < 3 && (indexOfCurrent === -1 || results[indexOfCurrent].result === SIMPLE_RESULTS.incorrect)
+    return { nthRetry, needsToBeRepeated }
```
klinzo, line 148: Something should work as well:
```suggestion
   return results.find(result => result.id === currentDocument?.id).map(({ numberOfTries, result }) => ({ numberOfTries, needsToBeRepeated: numberOfTries < 3 && result === SIMPLE_RESULTS.incorrect }))
```

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
```diff
@@ -163,7 +190,11 @@ const ChoiceExerciseScreen = ({
             <ButtonContainer>
               {selectedAnswer !== null ? (
                 <Button onPress={onFinishWord} buttonTheme={BUTTONS_THEME.dark}>
-                  <LightLabelInput>{lastWord ? labels.exercises.showResults : labels.exercises.next}</LightLabelInput>
+                  <LightLabelInput>
+                    {lastWord && !getRetryInfoOfCurrent().needsToBeRepeated
```
klinzo, line 194:
```suggestion
                    {lastWord && !needsToBeRepeated
```

src/routes/write-exercise/components/FeedbackSection.tsx
```diff
@@ -48,7 +49,9 @@ const Feedback = ({ result, document, submission }: FeedbackPropsType): ReactEle
   } else {
     Icon = IncorrectFeedbackIcon
     background = incorrect_background
-    message = `${labels.exercises.write.feedback.wrong} „${document.article.value} ${document.word}“`
+    message = needsToBeRepeated
+      ? labels.exercises.write.feedback.wrongForRetry
```
klinzo, line 53: What does `wrongForRetry` mean?

src/routes/write-exercise/components/__tests__/Actions.spec.tsx
```diff
@@ -29,9 +33,8 @@ describe('Components', () => {
         isFinished: true
       }
 
-      const component = shallow(<Action {...actionProps} />)
-      expect(component.find('Button')).toHaveLength(1)
-      expect(component.find('[testID="check-out"]')).toHaveLength(1)
+      const { getByText } = render(<Action {...actionProps} />, { wrapper: wrapWithTheme })
+      expect(getByText(labels.exercises.showResults)).toBeDefined()
```
klinzo, line 37: :+1: Seems like you really have the hang of testing now and you even like it kind of? :D Good job!

src/routes/write-exercise/components/__tests__/FeedbackSection.spec.tsx
```diff
@@ -40,9 +46,12 @@ describe('Feedback section', () => {
 
   it('should render incorrect feedback', () => {
```
klinzo, line 47: Perhaps you could add a short test with `needsToBeRepeated` set to true?

src/routes/write-exercise/components/WriteExercise.tsx
```diff
@@ -130,23 +127,49 @@ const WriteExercise = ({
...
+      setResults([...results, result])
+    }
+  }
+
+  const getRetryInfoOfCurrent = (): { nthRetry: number; needsToBeRepeated: boolean } => {
+    const indexOfCurrent = results.findIndex(result => result.id === document?.id)
+    const nthRetry = indexOfCurrent === -1 ? 0 : results[indexOfCurrent].numberOfTries
+    const needsToBeRepeated =
+      nthRetry < 3 && (indexOfCurrent === -1 || results[indexOfCurrent].result !== SIMPLE_RESULTS.correct)
+    return { nthRetry, needsToBeRepeated }
```
klinzo, line 172: Same as in other exercise, refactor and execute once on top of file. Could you extract `3` to a constant?

src/routes/write-exercise/components/WriteExercise.tsx
```diff
@@ -130,23 +127,49 @@ const WriteExercise = ({
...
+      getRetryInfoOfCurrent().needsToBeRepeated ? tryLater() : setCurrentDocumentNumber(currentDocumentNumber + 1)
     }
   }
 
-  const storeResult = (score: SimpleResultType): void => {
-    setResults([...results, { ...document, result: score }])
+  const storeResult = (score: SimpleResultType, noFurtherRetries: boolean = false): void => {
+    const nthRetry = noFurtherRetries ? 3 : getRetryInfoOfCurrent().nthRetry
+
+    const indexOfCurrentResult = results.findIndex(result => result.id === document?.id)
```
klinzo, line 148: Just use `find` and `map` instead

klinzo (changes requested):

src/routes/choice-exercises/__tests__/ArticleChoiceExerciseScreen.spec.tsx
SteffiStuffel: Answer solutions for ArticleChoice exercises are:
Der Helm
Die Helm
Das Helm
Die Helm (Plural)
To check if the correct word, is asked/prompted it is enough to just check the first is the word we are looking for, all are the same but getByText fails if the text exists more than one time.

src/routes/choice-exercises/__tests__/WordChoiceExerciseScreen.spec.tsx
SteffiStuffel: Won't work as the SingleChoice exercises use the `onPressIn` and `onPressOut` handlers


`
  const onPressIn = (): void => {
    setPressed(true)
  }

  const onPressOut = (): void => {
    setPressed(false)
    onClick(answer)
  }

  return (
    <Container
      activeOpacity={1}
      correct={showCorrect}
      selected={selected}
      onPressIn={onPressIn}
      onPressOut={onPressOut}
      pressed={pressed}
`

src/routes/choice-exercises/__tests__/WordChoiceExerciseScreen.spec.tsx
SteffiStuffel: yeah indeed, but i think the variable makes the test better understandable, cause it is clear always the correctAnswer is clicked.

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
SteffiStuffel: Not sure if a map would work, tested you solution, which does not work. 
We need to add a element to the array, if a certain condition is true, a map-function is not appropriate for this use case.

src/routes/write-exercise/components/FeedbackSection.tsx
SteffiStuffel: If you enter the solution wrong for the first time you get the answer:
"That was wrong"
If you enter the solution wrong for the third time you get the answer:
"That was wrong, correct would be: der Hammer"

So if retrying you don't get the solution, you just can retry.

src/routes/write-exercise/components/__tests__/FeedbackSection.spec.tsx
SteffiStuffel: done

src/routes/write-exercise/components/WriteExercise.tsx
SteffiStuffel: Moved 3 to a constant. Did not understand what you mean with the first part.

src/routes/write-exercise/components/WriteExercise.tsx
SteffiStuffel: Same as above, using map for adding a new entry does not work. Also we need to copy the array so useState notices the update, just mutating the array is not recommended as changes can be overwritten.

src/routes/choice-exercises/__tests__/ArticleChoiceExerciseScreen.spec.tsx
klinzo: Then I think you should assert `expect(getAllByText('Helm')).toHaveLength(4)`

src/routes/write-exercise/components/FeedbackSection.tsx
klinzo: Then I'd prefer some naming like `wrong` and `wrongWithSolution` or similar, that you are able to retry is not related to the label.

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
klinzo: You are right, missed that. I don't think that you should modify `results` directly like this though: `results[indexOfCurrentResult] = result`

src/routes/write-exercise/components/WriteExercise.tsx
klinzo: This does not need to be part of a function. Just refactor using `find` as suggested in the other exercise and move it to the top.
I.e:
```
const { nthRetry, needsToBeRepeated } = results.find...
```

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
SteffiStuffel: fixed this.

src/routes/choice-exercises/__tests__/ArticleChoiceExerciseScreen.spec.tsx
SteffiStuffel: changed it, but as this is not what the test tests i think it really does not matter.

src/routes/write-exercise/components/WriteExercise.tsx
klinzo: Hmmm yeah. I think the code is really complicated and hard to understand though. Some ways I could think that perhaps could make it easier:
* Use a map (`{ [id]: result }`) instead of an array to store the results
* Initialize the results array for all documents

src/routes/write-exercise/components/FeedbackSection.tsx
SteffiStuffel: changed it.

src/routes/choice-exercises/__tests__/ArticleChoiceExerciseScreen.spec.tsx
klinzo: Why is the length `3`?

SteffiStuffel: Because `Helm (Plural)` does not match `Helm`. getByText is always a full-match.

klinzo: Then please use `getAllByText(/Helm/)`

src/routes/write-exercise/components/WriteExercise.tsx
SteffiStuffel: I though about this a little and i think it makes sense to split `WriteExercise` up into two components.
One holds knows all the stuff relevant for all words of the exercise, e.g. which is the currentWord, what are all the results and documents, how may words are in this exercise
One knows only stuff about the current word: Which is the current word, what was submitted and so on.

Are you fine with this? If so, I will do it in a separate PR.

klinzo: Sounds really reasonable, thanks!

src/routes/write-exercise/components/WriteExercise.tsx
SteffiStuffel: will be done in LUN-219

release-notes/unreleased/main.yml
```diff
@@ -0,0 +1,6 @@
+issue_key:
```
klinzo, line 1: Missing issue key. The name of the release note is also wrong (`main.yml`)

klinzo (commented):
I know you will refactor the WriteExercise in a separate PR, however I'd really like it if you could apply the two suggestions not yet resolved (using find and executing once on top) before merging. If you want I can also do that for you (here or in a separate PR).

klinzo:
Tested on android emulator, works fine for choice exercises but afaics not for write exercise.

release-notes/unreleased/main.yml
SteffiStuffel: done.

SteffiStuffel:
> I know you will refactor the WriteExercise in a separate PR, however I'd really like it if you could apply the two suggestions not yet resolved (using find and executing once on top) before merging. If you want I can also do that for you (here or in a separate PR).

done.

src/routes/write-exercise/components/WriteExercise.tsx
```diff
@@ -135,44 +139,40 @@ const WriteExercise = ({
...
-    const newArr = results
     if (indexOfCurrentResult !== -1) {
-      newArr[indexOfCurrentResult] = result
       if (results[indexOfCurrentResult].result === 'similar') {
         newArr[indexOfCurrentResult] = {
           ...document,
-          result: nthRetry >= numberOfMaxRetries - 1 ? 'incorrect' : 'similar',
-          numberOfTries: nthRetry + 1
+          result: numberOfTries >= numberOfMaxRetries ? 'incorrect' : score,
+          numberOfTries: 5
```
klinzo, line 163: I guess this is not intended

klinzo (commented):
Actually I meant that you should apply the suggestions for the SingleChoiceExercise as there is no separate issue for it.

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
```diff
@@ -70,6 +70,10 @@ const ChoiceExerciseScreen = ({
   const documents = newDocuments ?? data
   const currentDocument = documents ? documents[currentWord] : null
 
+  const result = results.find(result => result.id === currentDocument?.id)
+  const nthRetry = !result ? 0 : result.numberOfTries
```
klinzo, line 74:
```suggestion
  const nthRetry = result?.numberOfTries ?? 0
```

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
```diff
@@ -70,6 +70,10 @@ const ChoiceExerciseScreen = ({
...
+  const needsToBeRepeated = nthRetry < numberOfMaxRetries && (!result || result.result === SIMPLE_RESULTS.incorrect)
```
klinzo, line 75: Shouldn't it be repeated if it is nearly correct? Or is this handled somewhere else?

klinzo (approved):
Probably intended: If I click on `Lösung anzeigen` words will not be repeated in the write exercise. I think this is fine though.
Tested on android.

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
SteffiStuffel: Nearly correct is not possible for SingleChoiceExercises

PR #190, opened 2021-11-29
LUN-124: Check for enough module size

Saraharar98:


klinzo (changes requested):
Thanks for your first PR! :tada: 
Some things that would be nice:
* Please don't make changes to Podfile, package.json or project.pbxproj files (and revert these now)
* Make sure to run prettier before committing/pushing. You can enable the git hook to do so automatically on commit. Edit: @SteffiStuffel seems like we don't have this hook in this project yet, right? This is the hook in the Integreat project: https://github.com/Integreat/integreat-app/blob/main/.github/hooks/pre-commit
* PRs should be named `LUN-<number>: Your description of what was done`
* Commits should be named `LUN-<number>: Your description of what was done`

SteffiStuffel:
> Thanks for your first PR! tada Some things that would be nice:
> 
> * Please don't make changes to Podfile, package.json or project.pbxproj files (and revert these now)
> * Make sure to run prettier before committing/pushing. You can enable the git hook to do so automatically on commit. Edit: @SteffiStuffel seems like we don't have this hook in this project yet, right? This is the hook in the Integreat project: https://github.com/Integreat/integreat-app/blob/main/.github/hooks/pre-commit
> * PRs should be named `LUN-<number>: Your description of what was done`
> * Commits should be named `LUN-<number>: Your description of what was done`

Commits should be named `LUN-<number>-your-description-of-what-was-done`

About the pre-commit hook: not sure about this. We have the intellij config for running prettier on save, but obviously this is not helpful when working on a mac. If you suggest/prefer having also a pre-commit hook, feel free to create a ticket, everybody using intellij can disable it if being annoyed by the "double check"

PR #192, opened 2021-11-29
LUN-223: Add pre commit hook

klinzo:


.github/hooks/pre-commit
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
```
klinzo, line 1: Copied from https://github.com/Integreat/integreat-app/pull/521

.github/hooks/pre-commit
```diff
@@ -0,0 +1,33 @@
...
+
+#
+# Install this using:
+# ln -s ../../.github/hooks/pre-commit .git/hooks/pre-commit
+#
+
+# Check if something changed in .circleci directory such that the config.yml needs to be updated
+if [[ $(git status --short | grep .circleci) ]]; then
```
SteffiStuffel, line 9: Think this does not really make sense. Remember the discussion we had on Integreat about this?
https://github.com/Integreat/integreat-app/pull/514/files

I would appreciate if, we copy this when the issue (IGAPP-855) is fixed or if you fix it now for lunes and integreat. We can also mark LUN-223-ticket as blocked by IGAPP-855.

SteffiStuffel (changes requested):

.github/hooks/pre-commit
SteffiStuffel: forget about this, too early in the morning. Everything fine.

SteffiStuffel (approved):

Taggotty (approved):

PR #193, opened 2021-11-29
LUN-217: Rename modules

klinzo:

I did tidy up a little bit:
* Removed redundant object `extraParams` in route params
* Extracted children description logic
* Unified Discipline and DisciplineType types
* Fixed delete custom discipline labels


src/components/__tests__/CustomDisciplineMenuItem.spec.tsx
```diff
@@ -21,6 +21,7 @@ describe('Components', () => {
       icon: 'none',
       numberOfChildren: 1,
       isLeaf: false,
+      isFirstLevel: true,
```
SteffiStuffel, line 24: I'm not happy with this naming. As we have `isLeaf` i would suggest to call it `isRoot`. Otherwise we have to name it `isFirstLevel` and `isLastLevel`, but as we have a classic tree i would strongly prefer the first.

src/components/__tests__/CustomDisciplineMenuItem.spec.tsx
```diff
@@ -41,7 +42,7 @@ describe('Components', () => {
       mockUseLoadAsyncWithData(mockData)
       const { getByText } = renderItem()
       expect(getByText('Custom Discipline')).toBeDefined()
-      expect(getByText(`1 ${labels.home.unit}`)).toBeDefined()
+      expect(getByText(`1 ${labels.general.firstLevelDiscipline}`)).toBeDefined()
```
SteffiStuffel, line 45: same here.

src/constants/endpoints.ts
```diff
@@ -7,7 +7,9 @@ export interface DisciplineType {
   icon: string
   numberOfChildren: number
   isLeaf: boolean
+  isFirstLevel: boolean
   apiKey?: string
+  total_training_sets?: number
```
SteffiStuffel, line 12: why have you added this? We have the logic to map the total_training_sets-Number to the numberOfChildren.

src/constants/labels.json
```diff
@@ -89,6 +84,13 @@
     "error": {
       "noWifi": "Es konnte keine Verbindung mit dem Internet hergestellt werden.",
       "retryButton": "Erneut versuchen"
-    }
+    },
+    "firstLevelDiscipline": "Beruf",
```
SteffiStuffel, line 88: please adjust to `rootDiscipline`

src/hooks/__tests__/useLoadDiscipline.spec.ts
```diff
@@ -93,13 +99,13 @@ describe('loadDiscipline', () => {
...
 
-    it('has no leaf parent without an api key', async () => {
+    it('has parent without an api key', async () => {
       await loadDisciplines({ ...parent, isLeaf: false })
       expect(getFromEndpoint).toHaveBeenCalledWith('disciplines_by_level/1234', undefined)
     })
 
-    it('has leaf parent', async () => {
-      await loadDisciplines(parent)
+    it('has training sets as children', async () => {
```
SteffiStuffel, line 107: please do not reproduce the training-sets naming from the api, as we want to get rid of it. 
`has training sets as children` is same as `isLeaf`

src/navigation/NavigationTypes.ts
```diff
@@ -11,16 +11,8 @@ export type CountsType = {
...
-  id: number
-  title: string
-  numberOfChildren: number
-  isLeaf: boolean
-  apiKey?: string
-}
-
 interface ResultScreenData {
-  discipline: Discipline
+  discipline: DisciplineType
```
SteffiStuffel, line 15: Actually we planned to change to type/interface-naming without the `Type`-suffix, but as it is already inconsistent, we can do it later.

src/hooks/useLoadDisciplines.ts
```diff
@@ -13,8 +12,8 @@ export interface ServerResponse {
   total_documents: number
 }
 
-const getEndpoint = (parent: Discipline | null): string => {
-  if (parent?.isLeaf) {
+const getEndpoint = (parent: DisciplineType | null): string => {
+  if (parent?.total_training_sets && parent.total_training_sets > 0) {
```
SteffiStuffel, line 16: Does this really make sense?
With my requested changes in /endpoints.ts this can be reverted. If you want to have this extra check you can add it in `formatServerResponse`-function.

src/hooks/useLoadDisciplines.ts
```diff
@@ -23,19 +22,20 @@ const getEndpoint = (parent: Discipline | null): string => {
...
 }
 
-const formatServerResponse = (serverResponse: ServerResponse[], apiKey?: string): DisciplineType[] =>
+const formatServerResponse = (serverResponse: ServerResponse[], parent: DisciplineType | null): DisciplineType[] =>
   serverResponse.map(item => ({
     ...item,
     numberOfChildren: item.total_discipline_children || item.total_training_sets || item.total_documents,
-    isLeaf: item.total_discipline_children === 0,
-    apiKey: apiKey
+    isLeaf: item.total_documents !== undefined,
```
SteffiStuffel, line 29: Why?

SteffiStuffel (changes requested):

SteffiStuffel:
Tested, works fine.

src/hooks/__tests__/useLoadDiscipline.spec.ts
klinzo: It is not the same. `isLeaf` should mean that there are no more disciplines as children but documents, until now it just meant that we need another endpoint (training sets instead of disciplines) for which the naming is just wrong and really confusing.

klinzo: With the previous logic the second last level disciplines had `isLeaf` set to true while the last level had `isLeaf` set to false. Therefore it was not possible to decide which wording (`Wort` vs `Modul`) to use only depending on the discipline.

src/components/__tests__/CustomDisciplineMenuItem.spec.tsx
klinzo: I am fine with it. The reason I decided against it was that for me the root was the home screen which has multiple `Bereiche`. But I guess `isRoot` is way more intuitive :+1:

src/constants/endpoints.ts
klinzo: See below. If you like that better I could also introduce a new property `hasTrainingSets` or `needsTrainingSetsEndpoint` or similar

src/hooks/useLoadDisciplines.ts
klinzo: See explanation above. I am open to introduce a different property which makes it clearer

src/hooks/useLoadDisciplines.ts
klinzo: Because the previous check did not check if it is a leaf :)

src/hooks/__tests__/useLoadDiscipline.spec.ts
SteffiStuffel: No, isLeaf means it is the last discipline in the tree, which is a tree of disciplines.
Disciplines are groups, disciplines and training-sets. Documents are no disciplines, therefore words/documents are not part of the tree.

Distinction between Word and Module are fine (and currently working) with isLeaf-property, as if a discipline isLeaf it has words as children. In the ExersiceOverviewScreen we can always just use "words" as unit.

klinzo: Yes, exactly what I am saying. With the previous implementation training sets where NOT leafs while the discipline having training sets as children were leafs

src/constants/endpoints.ts
SteffiStuffel: i see, why the previous usage of `isLeaf` is confusing (and inconsistent with our appview of disciplines), but i'm really unhappy to add this property. 
Would it be a better solution to add a property `childrenType`, which can be `'discipline' || 'training_set || 'documents'`, just a thought, not sure if this would make anything better...
But at least i would prefer one of your suggested boolean-variable, as there is no need to pass the numberOfChilderen twice.

Taggotty (commented):
Hmm. I don't quite get the changes with disciplines, tranings_sets, etc. ... maybe we can shortly discuss this in the Lunes Call?

src/constants/endpoints.ts
SteffiStuffel: can you please rename? I would prefer `needsTrainsSetsEndpoint: boolean`

src/routes/DisciplineSelectionScreen.tsx
```diff
@@ -71,42 +71,34 @@ const DisciplineSelectionScreen = ({ route, navigation }: DisciplineSelectionScr
...
-    const descriptionForWord = item.numberOfChildren === 1 ? labels.home.word : labels.home.words
-    const descriptionForUnit = item.numberOfChildren === 1 ? labels.home.unit : labels.home.units
-    const description = discipline.isLeaf ? descriptionForWord : descriptionForUnit
 
     return (
       <MenuItem selected={selected} item={item} onPress={() => handleNavigation(item)}>
         <ItemText>
           <BadgeLabel selected={selected}>{item.numberOfChildren}</BadgeLabel>
-          <Description selected={selected}>{description}</Description>
+          <Description selected={selected}>{childrenDescription(item)}</Description>
```
Taggotty, line 79: I think this should be `childrenLabel` otherwise the number of words is displayed twic
discipline_selection_num_words_twice

Taggotty (approved):
Tested on android emulator. I also have the feeling that we are relying a lot on several boolean expressions here. Maybe some other time it makes sense to rethink the curren Discipline structure on the client side, regarding better Types, that would distinct the Level of Discipline we currently have or sth.

src/routes/DisciplineSelectionScreen.tsx
klinzo: Good catch!

src/constants/endpoints.ts
klinzo: Done

SteffiStuffel (approved):

PR #194, opened 2021-12-01
LUN-164: Add Unit Test for ImageCarousel

Taggotty:


src/components/__tests__/ImageCarousel.spec.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  const getUri = (image: ReactTestInstance): string => image.props.source[0].uri
+
+  it('should render images', async () => {
+    const { getByTestId, findAllByRole } = renderImageCarousel()
+
+    const row = getByTestId('Swipeable')
+    expect(row).toBeTruthy()
+    const swipeable = row.children[0] as ReactTestInstance
+    swipeable.instance.loadImage(1) // preload the second image
```
Taggotty, line 44: The image rendering library we use renders all images next to each other. Images that haven't been rendered yet are only represented by an empty `View`, meaning they are only displayed after loadImage is called for that image. Therefore after the second image was loaded, there are two images, As the first one doesn't get 'unloaded' or sth. 
Therefore testing swiping is rather difficult here, so I just tested wether there are two images.

klinzo: Perhaps you could add a check that before `loadImage` only one image is rendered? I think a short comment in the code to explain this would also be nice.

klinzo (approved):

src/components/ImageCarousel.tsx
```diff
@@ -55,7 +55,7 @@ const ImageCarousel = ({ images }: ImageCarouselPropsType): ReactElement => {
   }
 
   return (
-    <ImageView>
+    <ImageView testID={'Swipeable'}>
```
SteffiStuffel, line 58: We already have the testID "Swipeable" used in HomeScreen, not sure if it matters if duplicated.

SteffiStuffel (approved):
Nice :tada:

src/components/ImageCarousel.tsx
Taggotty: should not lead to problems as those components don't share a screen it is possible to have multiple testIds with the same name.

PR #201, opened 2021-12-15
LUN-227: Fix cut off header text

Saraharar98:


klinzo (changes requested):
This solution probably only works for titles of some length. Could you try to solve this using flex or similar, i.e. such that the title uses all available space?

PR #202, opened 2021-12-30
LUN-229: Fix retry exercise button

SteffiStuffel:


src/routes/ResultsOverviewScreen.tsx
```diff
@@ -117,6 +117,7 @@ interface ResultOverviewScreenPropsType {
 
 const ResultsOverview = ({ navigation, route }: ResultOverviewScreenPropsType): ReactElement => {
   const { exercise, results, discipline } = route.params.result
+  const { repeatExercise } = route.params
```
klinzo, line 120: AFAIK functions should not be passed as route params, better to just pass the necessary values or just duplicate the function. This is also indicated by the warning: `Non-serializable values were found in the navigation state.` While it works, it is not good practice. See https://reactnavigation.org/docs/troubleshooting/#i-get-the-warning-non-serializable-values-were-found-in-the-navigation-state

klinzo:
Tested on android

src/routes/ResultsOverviewScreen.tsx
SteffiStuffel: I see, thank you. Wanted to avoid copying the function, but will do, in this case.

klinzo (approved):

f1sh1918 (approved):
Tested on android (emulator) & ios native :)

PR #203, opened 2022-01-03
LUN-199: Color of check input button wrong

Saraharar98:
…o lunesWhite


SteffiStuffel (changes requested):
The button should change it's color depending if the button is disabled. The button is disabled when the input is empty.

Best solution would be to:
1) rename `styledInput` to `disabled` and make it of type `boolean`
2) pass `!input` as a value
3) use it for the styling `color: ${props => (props.disabled ? props.theme.colors.lunesBlackLight : props.theme.colors.lunesWhite)};`

SteffiStuffel (approved):
Tested on Android emulator :tada:

klinzo (approved):
Tested on Android

PR #205, opened 2022-01-10
LUN-115: Adjust rgb values by using generic rgb profile values

f1sh1918:
Had to convert the rgb values to fit for ios splashscreen using srgb.
Easiest way is to adjust the "RGB Slider" settings to srgb profile
Further Info:
https://github.com/expo/expo/issues/826#issuecomment-494283319


SteffiStuffel (approved):
Not tested.

Taggotty (approved):
Also not tested

PR #206, opened 2022-01-12
LUN-227: Fix cut off header text

f1sh1918:
Fixed header text cut off.

**Test:**
Gastronomie, Lebensmittel -> Bäcker

Tested on android & ios (native and emulator)

### DIFF
--- a/src/navigation/Navigator.tsx
+++ b/src/navigation/Navigator.tsx
@@ -1,8 +1,8 @@
 import { NavigationContainer } from '@react-navigation/native'
 import { createStackNavigator, TransitionPresets } from '@react-navigation/stack'
 import React from 'react'
-import { Text, TouchableOpacity, StyleSheet } from 'react-native'
-import { widthPercentageToDP as wp } from 'react-native-responsive-screen'
+import { TouchableOpacity, StyleSheet } from 'react-native'
+import styled from 'styled-components/native'
 
 import { BackButton, CloseButton, BackArrowPressed, HomeButtonPressed, Home } from '../../assets/images'
 import labels from '../constants/labels.json'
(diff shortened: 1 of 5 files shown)


src/navigation/Navigator.tsx
```diff
@@ -28,26 +28,32 @@ export const styles = StyleSheet.create({
...
-    zIndex: 100
-  },
-  headerRight: {
-    paddingRight: 15
+  headerLeftContainer: {
+    flexGrow: 3
   }
 })
 
+export const NavigationTitle = styled.Text`
```
f1sh1918, line 40: Should we put styles in separate file?
I think in this case it make sense to reuse it f.e. in ExerciseHeader

klinzo: I think it makes sense if it is actually used in multiple places. Then just create a `NavigationTitle` component file

src/navigation/Navigator.tsx
```diff
@@ -140,7 +146,7 @@ const Navigator = (): JSX.Element => {
           options={{
             headerLeft: () => null,
             headerTitle: ' ',
```
klinzo, line 148:
```suggestion
            headerTitle: '',
```

klinzo (approved):
Tested on iOS simulator

src/navigation/Navigator.tsx
f1sh1918: hm the problem is that this is not a real component but a style wrapper for a component that i want to share.
i thought about sth like this f.e. having a folder like "test" but "styles" and create Navigator inside with its styles
https://blog.cloudboost.io/separate-your-code-with-styled-components-ec4fd1ee3ef8

klinzo: Its usable as a component though (and technically is a react component), isn't it? So no problem creating a separate file for it :)

SteffiStuffel (approved):
Looks good. Tested on ios emulator and android emulator.

PR #207, opened 2022-01-12
LUN-235: Fix back navigation in resultscreen

f1sh1918:
I think just use the goBack functionality should do the trick. Don't know why we need here an explicit screen.
I hope i haven't overseen anything

### DIFF
--- a/release-notes/unreleased/LUN-235-fix-back-navigation-in-resultscreen.yml
+++ b/release-notes/unreleased/LUN-235-fix-back-navigation-in-resultscreen.yml
@@ -1,0 +1,6 @@
+issue_key: LUN-235
+show_in_stores: true
+platforms:
+  - android
+  - ios
+de: Navigationsbug behoben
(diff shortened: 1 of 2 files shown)


klinzo (approved):
Tested on Android

release-notes/unreleased/LUN-235-fix-back-navigation-in-resultscreen.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: LUN-235
+show_in_stores: true
+platforms:
+  - android
+  - ios
+de: Navigation zurück zur Ergebnisübersicht behoben
```
SteffiStuffel, line 6: Navigation behoben? eher "Navigation gefixed" oder "Navigationsbug behoben"

SteffiStuffel (approved):
PR name has wrong prefix. Tested on android emulator.

PR #212, opened 2022-01-19
LUN-230: Adjust tests for popover

f1sh1918:


src/routes/write-exercise/components/__tests__/MissingArticlePopover.spec.tsx
```diff
@@ -5,21 +5,24 @@ import labels from '../../../../constants/labels.json'
...
   })
-
+  // Unfortunately jest renders the popover(modal) even if it's set to invisible
   it('should not show Popover when invisible', () => {
-    const { queryByText } = render(<MissingArticlePopover setIsPopoverVisible={jest.fn()} isVisible={false} />, {
+    const { getByTestId } = render(<MissingArticlePopover setIsPopoverVisible={jest.fn()} isVisible={false} />, {
       wrapper: wrapWithTheme
     })
-    expect(queryByText(labels.exercises.write.feedback.articleMissing)).toBeNull()
+    expect(getByTestId('popover').props.isVisible).toBeFalsy()
```
f1sh1918, line 26: I know this isn't a good test, but seems like jest renders the modal in any case. you can check with `debug()`
Spent lots of time in it and couldn't find a better solution

src/routes/write-exercise/components/__tests__/MissingArticlePopover.spec.tsx
```diff
@@ -5,21 +5,24 @@ import labels from '../../../../constants/labels.json'
...
+  default: 'Popover',
+  PopoverPlacement: 'top'
+}))
+
+describe('MissingArticlePopover', () => {
   it('should show Popover when visible', () => {
     const { getByText } = render(<MissingArticlePopover setIsPopoverVisible={jest.fn()} isVisible={true} />, {
       wrapper: wrapWithTheme
     })
     expect(getByText(labels.exercises.write.feedback.articleMissing)).toBeDefined()
```
SteffiStuffel, line 19: but wouldn't his be always be defined, also if popover is invisible, as you wrote on the test below. So maybe we should check here also for the visible prop?

f1sh1918: you're right. @Taggotty and me tried some stuff but didn't find a better solution so far

klinzo (approved):
I don't think we should spend too much time on this so fine for me :)

SteffiStuffel (approved):

src/routes/write-exercise/components/__tests__/InteractionSection.spec.tsx
```diff
@@ -53,17 +63,15 @@ describe('InteractionSection', () => {
...
+    const { getByText, getByPlaceholderText, getByTestId } = renderInteractionSection(
       { ...document, result: null, numberOfTries: 0 },
       false
     )
     const inputField = getByPlaceholderText(labels.exercises.write.insertAnswer)
     fireEvent.changeText(inputField, 'Spachtel')
     const button = getByText(labels.exercises.write.checkInput)
     fireEvent.press(button)
-    await waitFor(() => expect(getByText(labels.exercises.write.feedback.articleMissing)).toBeDefined())
+    await waitFor(() => expect(getByTestId('popover').props.isVisible).toBeTruthy())
```
Taggotty, line 75: just as a hint. This is the same as `await findByTestId('...')`

Taggotty (approved):

PR #217, opened 2022-01-24
LUN-228: Remove Enzyme

Taggotty:


package.json
```diff
@@ -7,7 +7,7 @@
     "android:production": "react-native run-android --no-packager --variant=release",
     "ios": "react-native run-ios",
     "start": "react-native start",
-    "test": "jest --silent --verbose --maxWorkers=1",
+    "test": "jest --verbose --maxWorkers=1",
```
Taggotty, line 10: Any reason why we had --silent set?

src/routes/__tests__/AddCustomDisciplineScreen.spec.tsx
```diff
@@ -67,19 +67,18 @@ describe('AddCustomDisciplineScreen', () => {
...
+    expect(await findByText(labels.addCustomDiscipline.error.alreadyAdded)).not.toBeNull()
   })
 
   it('should show wrong-code-error', async () => {
     const { findByText, findByPlaceholderText } = render(<AddCustomDisciplineScreen navigation={navigation} />, {
       wrapper: wrapWithTheme
     })
-    mocked(loadGroupInfo).mockRejectedValueOnce({ code: 403 })
-
+    mocked(loadGroupInfo).mockRejectedValueOnce({ response: { status: 403 }})
```
Taggotty, line 77: Not sure why this test worked before ^^

jest.setup.ts
```diff
@@ -1,7 +1,10 @@
 import mockAsyncStorage from '@react-native-async-storage/async-storage/jest/async-storage-mock'
 import '@testing-library/jest-native/extend-expect'
 import 'react-native-gesture-handler/jestSetup'
+// @ts-ignore
```
klinzo, line 4: Why do we need this here? Perhaps add a comment

jest.setup.ts
```diff
@@ -1,7 +1,10 @@
...
+import mockSafeAreaContext from 'react-native-safe-area-context/jest/mock';
 
+jest.mock('react-native-safe-area-context', () => mockSafeAreaContext);
```
klinzo, line 7:
```suggestion
import mockSafeAreaContext from 'react-native-safe-area-context/jest/mock'

jest.mock('react-native-safe-area-context', () => mockSafeAreaContext)
```

src/routes/__tests__/AddCustomDisciplineScreen.spec.tsx
klinzo: Probably because a promise is not `null`. Could we use `queryByText` instead?

src/routes/write-exercise/__tests__/WriteExerciseScreen.spec.tsx
```diff
@@ -69,12 +69,13 @@ describe('WriteExerciseScreen', () => {
...
-    mockUseLoadAsyncWithError('Network Error')
-    const { findByText } = render(<WriteExerciseScreen route={route} navigation={navigation} />, {
+  it('should show network error', async () => {
+    const errorMessage = 'Network Error'
+    mockUseLoadAsyncWithError(errorMessage)
+    const { findByText, debug } = render(<WriteExerciseScreen route={route} navigation={navigation} />, {
       wrapper: wrapWithTheme
     })
-    expect(findByText(labels.general.error.noWifi)).toBeDefined()
+    expect(await findByText(`${labels.general.error.noWifi} (${errorMessage})`)).toBeDefined()
```
klinzo, line 78: Do we really want to display the error message as well?

klinzo (approved):

package.json
SteffiStuffel: I don't know why, so I think it is fine to remove.

SteffiStuffel (approved):
Nice :tada:

src/routes/__tests__/AddCustomDisciplineScreen.spec.tsx
Taggotty: Only in combination with wait for. However, according to the documentation findByText is the same as waitWor -> getBy, therfore this should not make any difference here

src/routes/write-exercise/__tests__/WriteExerciseScreen.spec.tsx
Taggotty: Hmm. That's how it's implemented currently. It's basically the same as the other comment. This test shouln't have worked beforehand.

PR #220, opened 2022-01-26
LUN-224: Remove type suffix

Taggotty:


Taggotty:
The only "Types" left are `ResultType` and `ArticleType`, but as they are ambiguous with `Result` and `Article` I kept them as they were. Also `ResultType` and `ArticleType` are in fact the type of result/article. So I think it is fine.

klinzo (approved):
Not tested

SteffiStuffel (approved):
Tested on Android emulator.

PR #222, opened 2022-01-26
LUN-212: Write exercise scrollable screen

f1sh1918:
Tested on android emulator & ios native


src/components/ImageCarousel.tsx
```diff
@@ -1,37 +1,51 @@
...
 import { COLORS } from '../constants/theme/colors'
 
-const ImageView = styled.View`
-  height: 35%;
+const ImageView = styled.View<{ height: number }>`
+  height: ${({ height }) => height}px;
 `
-const StyledImage = styled.Image`
-  height: 100%;
+const StyledImage = styled.Image<{ height: number; minimized: boolean }>`
```
f1sh1918, line 13: Due to being in this imageViewer lib, i couldn't find a better way to resize the image within the resized container. Percentage values didn't work for the image

src/components/ImageCarousel.tsx
```diff
@@ -1,37 +1,51 @@
...
+  height: ${({ height }) => height}px;
+  ${({ minimized }) =>
+    minimized &&
+    `
+    aspect-ratio: 1.5;
```
f1sh1918, line 18: `resizeMode='center'` did only work on android properly, so i used this way to center and set a nice aspectRadio

src/routes/write-exercise/components/WriteExercise.tsx
```diff
@@ -41,14 +42,23 @@ const WriteExercise = ({ documents, route, navigation }: WriteExercisePropType):
...
+  const keyboard = useKeyboard()
   const current = documentsWithResults[currentIndex]
   const nthRetry = current.numberOfTries ?? 0
   const needsToBeRepeated = nthRetry < numberOfMaxRetries && current.result !== SIMPLE_RESULTS.correct
 
   const tryLater = useCallback(() => {
-    setDocumentsWithResults(moveToEnd(documentsWithResults, currentIndex))
-    Keyboard.dismiss()
-  }, [documentsWithResults, currentIndex])
+    // Wait until keyboard is hidden before go next
```
f1sh1918, line 51: ImageViewer component does not resize properly if we do not wait for the keyboard to be dismissed before switching to the next question

src/components/ImageCarousel.tsx
```diff
@@ -1,37 +1,51 @@
...
+
 interface ImageUrlType {
   url: string
 }
 
-const ImageCarousel = ({ images }: ImageCarouselPropsType): ReactElement => {
+const ImageCarousel = ({ images, minimized = false }: ImageCarouselPropsType): ReactElement => {
+  const { height: deviceHeight } = useWindowDimensions()
+  const heightPercent = minimized ? 17.5 : 35
+  const viewerHeight = (deviceHeight * heightPercent) / 100
```
klinzo, line 47: Maybe a comment would make sense here. Also avoid magic numbers (perhaps better to write `17.5` as `35 / 2`)

src/components/ImageCarousel.tsx
klinzo: Perhaps you could add a comment `Center image` if that is the purpose

src/components/ImageCarousel.tsx
```diff
@@ -40,7 +54,7 @@ const ImageCarousel = ({ images }: ImageCarouselPropsType): ReactElement => {
     return !currentIndex || !allSize ? (
       <></>
     ) : (
-      <PaginationView>
+      <PaginationView topPosition={minimized ? -10 : 10}>
```
klinzo, line 57: Better to just pass `minimized` here imo and do the styling in the styled component

src/routes/write-exercise/components/WriteExercise.tsx
klinzo: Why do we only need to handle this for `tryLater` and not the other buttons like check and give up? Also comments should better tell WHY something is done than WHAT is done, the what should be clear most of the time by looking at the code (oc not always possible).
```suggestion
    // ImageViewer is not resized correctly if keyboard is not dismissed before going to next document
```

src/routes/write-exercise/components/WriteExercise.tsx
```diff
@@ -41,14 +42,23 @@ const WriteExercise = ({ documents, route, navigation }: WriteExercisePropType):
...
+    if (keyboard.keyboardShown) {
+      const onKeyboardHideSubscription = Keyboard.addListener('keyboardDidHide', () => {
+        setDocumentsWithResults(moveToEnd(documentsWithResults, currentIndex))
+        onKeyboardHideSubscription.remove()
+      })
+      Keyboard.dismiss()
+    } else {
+      setDocumentsWithResults(moveToEnd(documentsWithResults, currentIndex))
+    }
```
klinzo, line 60: Why not just:
```suggestion
    if (keyboard.keyboardShown) {
      Keyboard.dismiss()
    }
    setDocumentsWithResults(moveToEnd(documentsWithResults, currentIndex))
```

src/routes/write-exercise/components/WriteExercise.tsx
```diff
@@ -41,14 +42,23 @@ const WriteExercise = ({ documents, route, navigation }: WriteExercisePropType):
...
-    setDocumentsWithResults(moveToEnd(documentsWithResults, currentIndex))
-    Keyboard.dismiss()
-  }, [documentsWithResults, currentIndex])
+    // Wait until keyboard is hidden before go next
+    if (keyboard.keyboardShown) {
+      const onKeyboardHideSubscription = Keyboard.addListener('keyboardDidHide', () => {
+        setDocumentsWithResults(moveToEnd(documentsWithResults, currentIndex))
+        onKeyboardHideSubscription.remove()
+      })
+      Keyboard.dismiss()
```
klinzo, line 57: Why do you use both the hook and the import from react-native? Do we need the additional dependency?

klinzo (commented):
Writing tests for this is probably too difficult/does not make sense as it is mostly styling, right?

Tested on Android, I like it. In the long term a smooth animation could make sense but I think that is not important right now.

src/routes/write-exercise/components/WriteExercise.tsx
f1sh1918: because the dismiss is not fast enough and the next screen gets rerendered with minimized picture and then the keyboard closes

src/routes/write-exercise/components/WriteExercise.tsx
f1sh1918: should i add the regular lifecycle function or create my own hook? sth like this?
```js
const [isKeyboardVisible, setKeyboardVisible] = useState(false);

 useEffect(() => {
    const keyboardDidShowListener = Keyboard.addListener(
      'keyboardDidShow',
      () => {
        setKeyboardVisible(true); // or some other action
      }
    );
    const keyboardDidHideListener = Keyboard.addListener(
      'keyboardDidHide',
      () => {
        setKeyboardVisible(false); // or some other action
      }
    );

    return () => {
      keyboardDidHideListener.remove();
      keyboardDidShowListener.remove();
    };
  }, []);
```

yarn.lock
```diff
@@ -1872,6 +1872,11 @@
   resolved "https://registry.yarnpkg.com/@react-native-community/eslint-plugin/-/eslint-plugin-1.1.0.tgz#e42b1bef12d2415411519fd528e64b593b1363dc"
   integrity sha512-W/J0fNYVO01tioHjvYWQ9m6RgndVtbElzYozBq1ZPrHO/iCzlqoySHl4gO/fpCl9QEFjvJfjPgtPMTMlsoq5DQ==
 
+"@react-native-community/hooks@^2.8.1":
```
SteffiStuffel, line 1875: Why have you added this?

src/routes/write-exercise/components/WriteExercise.tsx
```diff
@@ -41,14 +42,23 @@ const WriteExercise = ({ documents, route, navigation }: WriteExercisePropType):
...
   const nthRetry = current.numberOfTries ?? 0
   const needsToBeRepeated = nthRetry < numberOfMaxRetries && current.result !== SIMPLE_RESULTS.correct
 
   const tryLater = useCallback(() => {
-    setDocumentsWithResults(moveToEnd(documentsWithResults, currentIndex))
-    Keyboard.dismiss()
-  }, [documentsWithResults, currentIndex])
+    // ImageViewer is not resized correctly if keyboard is not dismissed before going to next document
+    if (isKeyboardShown) {
+      const onKeyboardHideSubscription = Keyboard.addListener('keyboardDidHide', () => {
```
SteffiStuffel, line 53: Why is this subscription needed?

SteffiStuffel (approved):
Tested on ios and android (browserstack).

Works fine, but when scrolling down, there ist now to much white space (both on android and ios)

LUN-212

src/hooks/useKeyboard.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect, useState } from 'react'
+import { Keyboard } from 'react-native'
+
+// This hook listens to the keyboard and returns boolean whether keyboard is opened or closed
+
+export const useKeyboard = (): boolean => {
```
klinzo, line 6:
```suggestion
export const useIsKeyboardVisible = (): boolean => {
```

src/components/ImageCarousel.tsx
```diff
@@ -1,37 +1,53 @@
...
 }
+
 interface ImageUrlType {
   url: string
 }
 
-const ImageCarousel = ({ images }: ImageCarouselPropsType): ReactElement => {
+const ImageCarousel = ({ images, minimized = false }: ImageCarouselPropsType): ReactElement => {
+  const { height: deviceHeight } = useWindowDimensions()
+  // Since ImageViewer Lib doesn't resize image correct on container size change, we set the current height here depending minimized value
```
klinzo, line 47:
```suggestion
  // Manually resize ImageViewer since it doesn't happen automatically on container size changes
```

src/routes/write-exercise/components/WriteExercise.tsx
f1sh1918: because we have to wait until the keyboard is hidden to go to the next word, so you have to listen to this event

yarn.lock
f1sh1918: thx probably forgot to push new yarn lock. gonna update this

src/hooks/useKeyboard.ts
f1sh1918: refactored manually!

f1sh1918:
> Tested on ios and android (browserstack).
> 
> Works fine, but when scrolling down, there ist now to much white space (both on android and ios)
> 
Just fixed the scroll space issue with last commit. Would be nice if you could test it on android native. Rest was tested by me

src/routes/write-exercise/components/WriteExercise.tsx
SteffiStuffel: Oke but why? What would happen if we don't wait?

f1sh1918: check comment above :) ImageViewwer doesn't resize correctly

SteffiStuffel: oke, i see. thanks

src/hooks/useIsKeyboardVisible.ts
```diff
@@ -3,7 +3,7 @@ import { Keyboard } from 'react-native'
 
 // This hook listens to the keyboard and returns boolean whether keyboard is opened or closed
```
klinzo, line 5: Optional: I think this comment is kind of redunant/describes implementation details and could therefore be removed.

klinzo (approved):
Tested on android again, nice work!

PR #225, opened 2022-02-01
LUN-197: Resize app screen for tablets

f1sh1918:

Finally we can all use lunes on our iPads /play sexyback

image


klinzo (approved):
Not tested

SteffiStuffel (approved):
Tested on browserstack (iPad Pro 12)

Looks quite shitty, as some elements resize and some don't, but this would be a new ticket.

PR #228, opened 2022-02-02
LUN-240: Escape slashes for browserstack upload

SteffiStuffel:


klinzo (commented):
What is `////-` doing? Do you have a link where this behavior is described?

SteffiStuffel:
> What is `////-` doing? Do you have a link where this behavior is described?

first `//` meas all occurrences are matched `//` is the escaped slash which should be replaced `-` is the char we want it wo replace with.
https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion

Taggotty (approved):
Nice 👍🏼  tested replacement locally in cmd

klinzo (approved):
Ah regex magic :+1:

PR #230, opened 2022-02-02
LUN-242: Header text could use more space

klinzo:


src/navigation/Navigator.tsx
```diff
@@ -37,11 +37,15 @@ export const styles = StyleSheet.create({
     borderBottomWidth: 1
   },
   headerRightContainer: {
-    paddingRight: 15,
-    flexShrink: 1
+    paddingHorizontal: 15,
+    maxWidth: 60
```
f1sh1918, line 41: setting a maxWidth will unfortunately not do the trick thats why spent so much time on it.
The right container has to stay also flexible. Thats why i worked with proportion or you use flex-basis

f1sh1918 (changes requested):
Tested on android. Right header container is too small in some exercises
You could add in `ExerciseHeader.tsx` a separate style, so Navigation Header will not be affected only Exercise Header:
` headerRightContainerStyle: {maxWidth:100, paddingRight:15}`

image

src/navigation/Navigator.tsx
f1sh1918: yes because this container gonna be used in different cases

src/components/ExerciseHeader.tsx
```diff
@@ -32,6 +32,7 @@ interface ExerciseHeaderProps {
 
 const ExerciseHeader = ({ navigation, route, currentWord, numberOfWords }: ExerciseHeaderProps): JSX.Element => {
   const [isModalVisible, setIsModalVisible] = useState(false)
+  const progressText = numberOfWords !== 0 ? `${currentWord + 1} ${labels.general.header.of} ${numberOfWords}` : ''
```
klinzo, line 35: Is this change fine @SteffiStuffel ?

f1sh1918 (approved):
Tested on android emulator and ios native. nice!

SteffiStuffel (approved):
Tested on browserstack android and ios.

src/components/ExerciseHeader.tsx
klinzo: May have been wanted :)

PR #233, opened 2022-02-14
LUN-202: Add axios cache interceptor

Taggotty:

This will keep all requests in cache for 5 minutes in memory, but we could also increase the time to live here. As we currently don't have a refresh you need to wait for 5 minutes or close and reopen the app the get the most recent responses.


SteffiStuffel:
Good solution. Tested on android emulator.

SteffiStuffel (approved):

klinzo (approved):

f1sh1918 (approved):

PR #234, opened 2022-02-15
LUN-232: Add qr code scanner

SteffiStuffel:

Can be tested with this valid QR-Code:
https://api.qrserver.com/v1/create-qr-code/?size=150x150&data=rXTIfCwz.eOYXwwn486fR5PlKFW9KagLtIzwwBCrE


src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
```diff
@@ -0,0 +1,133 @@
...
+          )}
+        </Icon>
+        <Camera
+          captureAudio={false}
+          onBarCodeRead={onBarCodeRead}
+          testID='camera'
+          onStatusChange={status => setPermissionDenied(status.cameraStatus === 'NOT_AUTHORIZED')}
+          onCameraReady={() => setPermissionDenied(false)}>
+          {() =>
+            permissionDenied && !requestPermissions ? (
```
f1sh1918, line 107: i think here is no need for tenary expression and render a fragment just use `&&` and remove  `: <></>`

src/routes/add-custom-discipline/AddCustomDisciplineScreen.tsx
```diff
@@ -30,22 +33,31 @@ const Description = styled.Text`
...
 `
 
+const StyledTextInput = styled.TextInput`
+  flex-grow: 2;
+  font-size: ${props => props.theme.fonts.largeFontSize};
+  font-weight: ${props => props.theme.fonts.lightFontWeight};
+  letter-spacing: 0.11px;
+  font-family: ${props => props.theme.fonts.contentFontRegular};
+  color: ${prop => prop.theme.colors.lunesBlack};
+  width: 80%;
```
f1sh1918, line 58:
```suggestion
  padding-right: 15px
```

android/app/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,7 @@
   package="app.lunes">
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.VIBRATE"/>
```
klinzo, line 5: Don't we need the camera permission as well? https://react-native-camera.github.io/react-native-camera/docs/installation.html#android---other-required-steps

src/routes/add-custom-discipline/AddCustomDisciplineScreen.tsx
```diff
@@ -30,22 +33,31 @@ const Description = styled.Text`
   padding: 10px 0;
 `
 
-const StyledTextInput = styled.TextInput<{ errorMessage: string }>`
-  font-size: ${props => props.theme.fonts.largeFontSize};
-  font-weight: ${props => props.theme.fonts.lightFontWeight};
-  letter-spacing: 0.11px;
-  font-family: ${props => props.theme.fonts.contentFontRegular};
-  color: ${prop => prop.theme.colors.lunesBlack};
+const InputContainer = styled.View<{ errorMessage: string }>`
```
klinzo, line 36:
```suggestion
const InputContainer = styled.View<{ error: boolean }>`
```

f1sh1918:
Really nice! Works like a charm. If someone gives another approve i could start with ios part

f1sh1918 (approved):
Tested on android

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+const NotAuthorisedView = styled.View`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 100%;
+  background-color: ${props => props.theme.colors.lunesWhite};
+`
+
+const NoAuthDescription = styled.Text`
```
klinzo, line 38:
```suggestion
const NotAuthorisedDescription = styled.Text`
```

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
```diff
@@ -0,0 +1,133 @@
...
+  font-family: ${props => props.theme.fonts.contentFontRegular};
+  font-size: ${props => props.theme.fonts.defaultFontSize};
+  color: ${props => props.theme.colors.lunesGreyMedium};
+  padding-bottom: 20px;
+`
+
+interface Props {
+  setVisible: (input: boolean) => void
+  setCode: (input: string) => void
```
klinzo, line 47:
```suggestion
  setVisible: (visible: boolean) => void
  setCode: (code: string) => void
```

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
```diff
@@ -0,0 +1,133 @@
...
+}
+
+const AddCustomDisciplineScreen = ({ setVisible, setCode }: Props): ReactElement => {
+  const appState = useRef(AppState.currentState)
+
+  const [isPressed, setIsPressed] = useState<boolean>(false)
+  const [requestPermissions, setRequestPermissions] = useState<boolean>(false)
```
klinzo, line 54: Maybe:
```suggestion
  const [permissionRequested, setPermissionRequested] = useState<boolean>(false)
```

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
```diff
@@ -0,0 +1,133 @@
...
+    if (requestPermissions) {
+      PermissionsAndroid.request(PermissionsAndroid.PERMISSIONS.CAMERA)
+        .then(() => setRequestPermissions(false))
+        .catch(e => console.error(e))
+    }
+  }, [requestPermissions])
+
+  useEffect(() => {
+    const subscription = AppState.addEventListener('change', nextAppState => {
+      if (appState.current.match(/inactive|background/) && nextAppState === 'active') {
```
klinzo, line 72: I know this is just copied from the docs but I think we should not use a regex here as regexes are error prone and remove type safety.
```suggestion
      const inactive = appState.current === 'inactive' || appState.current === 'background'
      if (inactive && nextAppState === 'active') {
```
Not sure if we could also change the check to
```suggestion
      if (appState.current !== 'active' && nextAppState === 'active') {
```

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
```diff
@@ -0,0 +1,133 @@
...
+        setRequestPermissions(true)
+      }
+      appState.current = nextAppState
+    })
+    return () => {
+      subscription.remove()
+    }
```
klinzo, line 79:
```suggestion
    return subscription.remove
```

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
```diff
@@ -0,0 +1,133 @@
...
+  }, [])
+
+  const refresh = () => {
```
klinzo, line 82:
```suggestion
  const openSettings = () => {
```

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
```diff
@@ -0,0 +1,133 @@
...
+                  onPress={() => setVisible(false)}
+                  label={labels.addCustomDiscipline.qrCodeScanner.noAuthorization.back}
+                  buttonTheme={BUTTONS_THEME.outlined}
+                />
+                <Button
+                  onPress={() => refresh()}
+                  label={labels.addCustomDiscipline.qrCodeScanner.noAuthorization.grant}
+                  buttonTheme={BUTTONS_THEME.contained}
+                />
+              </NotAuthorisedView>
```
klinzo, line 122: Could we extract this to a separate component?

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
```diff
@@ -0,0 +1,133 @@
...
+  padding-bottom: 20px;
+`
+
+interface Props {
+  setVisible: (input: boolean) => void
+  setCode: (input: string) => void
+}
+
+const AddCustomDisciplineScreen = ({ setVisible, setCode }: Props): ReactElement => {
+  const appState = useRef(AppState.currentState)
```
klinzo, line 51: Optional: To avoid having to use `.current` all the time you could just `useState` (which is interchangeable here) and `useRef` is only syntactic sugar:
```suggestion
  const [appState, setAppState] = useState(AppState.currentState)
```

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const [requestPermissions, setRequestPermissions] = useState<boolean>(false)
+  const [permissionDenied, setPermissionDenied] = useState<boolean>(false)
+
+  const onBarCodeRead = (scanResult: BarCodeReadEvent) => {
+    setCode(scanResult.data)
+    setVisible(false)
+  }
+
+  useEffect(() => {
+    if (requestPermissions) {
```
klinzo, line 63: Could you explain this? AFAICS `requestPermissions` is only ever set to `true` if the app comes from background to foreground while this is rendered. So if I open the app and navigate here without putting it in the background, the permission is never requested? What am I missing here? :see_no_evil:

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
```diff
@@ -0,0 +1,133 @@
...
+          ) : (
+            <CloseCircleIconWhite testID='close-circle-icon-white' />
+          )}
+        </Icon>
+        <Camera
+          captureAudio={false}
+          onBarCodeRead={onBarCodeRead}
+          testID='camera'
+          onStatusChange={status => setPermissionDenied(status.cameraStatus === 'NOT_AUTHORIZED')}
+          onCameraReady={() => setPermissionDenied(false)}>
```
klinzo, line 105: Do we need both?

src/constants/labels.json
```diff
@@ -19,6 +19,13 @@
       "alreadyAdded": "Dieser Code wurde bereits erfolgreich hinzufügt.",
       "wrongCode": "Dieser Code existiert nicht (ungültiger Code).",
       "technical": "Ein technischer Fehler ist aufgetreten. Prüfen Sie Ihre Internetverbindung."
+    },
+    "qrCodeScanner": {
+      "noAuthorization": {
+        "description": "Feature ist ohne Kamerazugriff nicht verfügbar",
```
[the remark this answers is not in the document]
klinzo: Perhaps something like the firefox text makes sense:
signal-2022-02-15-115521
```suggestion
        "description": "Kamerazugriff zum Scannen von QR-Codes erforderlich. Öffnen Sie die Android-Einstellungen, tippen Sie auf Berechtigungen und tippen Sie auf Erlauben.",
```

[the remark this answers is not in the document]
klinzo: Text for button should be changed as well then to `Einstellungen` or similar

klinzo (commented):
Tested on android, works fine!
Minor notes:
* If permissions are not granted and you click on the qr code icon, the screen goes black for a really short time and flickers.
* The white header looks a bit weird imo
* Back button in camera screen does nothing

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
SteffiStuffel: You are right, though because it is a faac it is needed, but it is not. Removed it.

src/routes/add-custom-discipline/AddCustomDisciplineScreen.tsx
SteffiStuffel: This would not work, as when entering a long code, the icon is moved to the right. Adjusted it nevertheless to width 90% and removed flex.

src/routes/add-custom-discipline/AddCustomDisciplineScreen.tsx
```diff
@@ -30,22 +33,31 @@ const Description = styled.Text`
...
       props.errorMessage ? props.theme.colors.lunesFunctionalIncorrectDark : props.theme.colors.lunesGreyDark};
   border-radius: 4px;
   margin-top: ${hp('8%')}px;
-  padding-left: 15px;
+  padding: 0 15px;
   height: ${hp('8%')}px;
 `
 
+const StyledTextInput = styled.TextInput`
+  flex-grow: 2;
```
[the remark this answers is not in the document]
SteffiStuffel: removed completely, not needed.

android/app/src/main/AndroidManifest.xml
SteffiStuffel: weird, why is this working then?

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
SteffiStuffel: We can, but it think it is not really needed. This component is only < 150 lines long and only has one responsibility: handling the camera and its permissions. So imo it is totally fine to keep it here.

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
SteffiStuffel: Tested, does not work. I guess because setAppState causes a rerender then.

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
SteffiStuffel: Yeah maybe i should add a comment here.
This is only needed when you got to the page, deny camera permissions, click then on settings. When then navigating back from settings this code is executed, leading to a second request for permission. The first one is triggered automatically by RNCamera, therefore no need for us to handle.

Not sure if this is the best solution, but i couldn't work out a better on. There is a `onStatusChange` function of RNCamera, but it does not get called (correctly), i guess because there is nor rerender, when navigating back from settings. I tried a lot of different approaches, but either i got stuck in a infinite rerender cycle or the sencond prompt for permissions was not fired. So I ended up with this solution.

SteffiStuffel: oke maybe this is important to mention: it is only needed when navigate to settings and then select "ask every time", when selecting "allow while using the app" it would work fine without this.

android/app/src/main/AndroidManifest.xml
klinzo: No idea. I think we don't need the vibrate permission though, do we?

SteffiStuffel:
> Tested on android, works fine! Minor notes:
> 
> * If permissions are not granted and you click on the qr code icon, the screen goes black for a really short time and flickers.

Arg, i though i fixed this. Will have another look at it.

> * Back button in camera screen does nothing

Cannot reproduce? You mean the button above the "Setting / Allow"-Button? Works fine for me.

android/app/src/main/AndroidManifest.xml
SteffiStuffel: No, i removed it.

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
klinzo: I guess it comes down to personal preference, yes. For me this would not be about the general length of the component (I agree here, < 150 LOC is perfectly fine) but the length of the jsx. I think 40 LOC is really long for jsx and makes it hard to review/work with (or at least for me). As this is just a suggestion, feel free to ignore it. Other opinions welcome as well. I just think we should strive for some modularity.

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
klinzo: Can't we just always request permissions in a useFocusEffect/app state change? Doesn't hurt to request permissions twice, right?

klinzo:
> > 
> > * If permissions are not granted and you click on the qr code icon, the screen goes black for a really short time and flickers.
> 
> Arg, i though i fixed this. Will have another look at it.

You could always request permissions manually on render/focus effect and show nothing until this is completed.

> > * Back button in camera screen does nothing
> 
> Cannot reproduce? You mean the button above the "Setting / Allow"-Button? Works fine for me.

No, I meant the "hardware" back button, not the one in the app.

src/routes/add-custom-discipline/__tests__/AddCustomDisciplineScreen.spec.tsx
```diff
@@ -84,4 +89,14 @@ describe('AddCustomDisciplineScreen', () => {
     fireEvent.press(submitButton)
     expect(await findByText(labels.addCustomDiscipline.error.wrongCode)).not.toBeNull()
   })
+
+  it('should open qr code scanner', async () => {
+    const { findByLabelText } = render(<AddCustomDisciplineScreen navigation={navigation} />, {
+      wrapper: wrapWithTheme
+    })
+    const QRCodeIcon = await findByLabelText('qr-code-scanner')
```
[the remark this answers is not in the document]
SteffiStuffel: All tests in this file are async, because this component is not rendered until `useReadCustomDisciplines` returns the customDisciplines from AsyncStorage.

SteffiStuffel:
> > > * If permissions are not granted and you click on the qr code icon, the screen goes black for a really short time and flickers.
> > 
> > 
> > Arg, i though i fixed this. Will have another look at it.
> 
> You could always request permissions manually on render/focus effect and show nothing until this is completed.
> 
> > > * Back button in camera screen does nothing
> > 
> > 
> > Cannot reproduce? You mean the button above the "Setting / Allow"-Button? Works fine for me.
> 
> No, I meant the "hardware" back button, not the one in the app.

I see, added handling.

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
SteffiStuffel: oke convinced me, moved it.

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
SteffiStuffel: I could not get this working, always leading to a loop, if you can make a working solution with useFocusEffect or anything, I would be happy to see it. For me it always leaded into an endless loop of active / focused and opening and closing the request-dialog.

klinzo: will try and let you know :)

src/routes/add-custom-discipline/components/NotAuthorisedView.tsx
```diff
@@ -0,0 +1,46 @@
...
+  font-family: ${props => props.theme.fonts.contentFontRegular};
+  font-size: ${props => props.theme.fonts.defaultFontSize};
+  color: ${props => props.theme.colors.lunesGreyMedium};
+  padding: 0 20px 20px;
+  text-align: center;
+`
+
+interface Props {
+  setVisible: (visible: boolean) => void
+  openSettings: () => void
```
klinzo, line 27: Optional: Could be moved here.

klinzo (commented):
Hint: Some comments are hidden and not yet addressed.

android/app/src/main/AndroidManifest.xml
klinzo: Maybe because it's a debug build but not sure.

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
SteffiStuffel: oke more or less just renaming. Thought it would be cool to get rid of the useEffect or listening to the AppState.

src/routes/add-custom-discipline/components/NotAuthorisedView.tsx
SteffiStuffel: What could be moved here?

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
```diff
@@ -44,32 +44,32 @@ const AddCustomDisciplineScreen = ({ setVisible, setCode }: Props): ReactElement
...
+  }, [permissionRequested])
 
   useEffect(() => {
     const subscription = AppState.addEventListener('change', nextAppState => {
-      if (appState.current.match(/inactive|background/) && nextAppState === 'active') {
-        setRequestPermissions(true)
+      if ((appState.current === 'inactive' || appState.current === 'background') && nextAppState === 'active') {
+        setPermissionRequested(false)
       }
+      console.log('app state changed')
```
klinzo, line 61:
```suggestion
```

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
```diff
@@ -44,32 +44,32 @@ const AddCustomDisciplineScreen = ({ setVisible, setCode }: Props): ReactElement
...
-  const openSettings = () => {
-    Linking.openSettings().catch(() => console.error('Unable to open Settings'))
-  }
-
   return (
     <Modal visible transparent animationType='fade' onRequestClose={() => setVisible(false)}>
       <Container>
+        <Text>
+          {permissionRequested} {permissionGranted}
+        </Text>
```
klinzo, line 72:
```suggestion
```

src/routes/add-custom-discipline/components/QRCodeReaderOverlay.tsx
```diff
@@ -44,32 +44,32 @@ const AddCustomDisciplineScreen = ({ setVisible, setCode }: Props): ReactElement
 
   // Needed when navigating back from settings, when users selected "ask every time" as camera permission option
   useEffect(() => {
-    if (requestPermissions) {
+    if (!permissionRequested) {
+      console.log('useEffect')
```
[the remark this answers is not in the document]
SteffiStuffel: arrg, sorry about that

klinzo (approved):
Nice!

PR #235, opened 2022-02-15
LUN-253: Change discipline item onPress behavior

klinzo:

### DIFF
--- a/src/components/DisciplineItem.tsx
+++ b/src/components/DisciplineItem.tsx
@@ -4,7 +4,6 @@ import { widthPercentageToDP as wp } from 'react-native-responsive-screen'
 import styled from 'styled-components/native'
 
 import { ChevronRight } from '../../assets/images'
-import { Discipline } from '../constants/endpoints'
 import { COLORS } from '../constants/theme/colors'
 
 const Container = styled(Pressable)<{ selected: boolean }>`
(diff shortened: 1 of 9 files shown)


src/components/DisciplineItem.tsx
```diff
@@ -1,60 +1,97 @@
...
+const DescriptionContainer = styled.View`
+  flex-direction: row;
+  align-items: center;
+`
+
+const Description = styled.Text<{ pressed: boolean }>`
+  font-size: ${props => props.theme.fonts.defaultFontSize};
+  font-weight: ${props => props.theme.fonts.lightFontWeight};
+  font-family: ${props => props.theme.fonts.contentFontRegular};
+  color: ${props => (props.pressed ? props.theme.colors.lunesWhite : props.theme.colors.lunesGreyMedium)};
```
klinzo, line 51: The color is now `lunesWhite` (`white` before). Not sure if that is correct but in this screen it is that way: https://app.zeplin.io/project/5fc7c98a435298126bcfd4ee/screen/5fd7ad2c3cfec77f5c68f10c

src/components/__tests__/Button.spec.tsx
```diff
@@ -11,81 +11,79 @@ import resetAllMocks = jest.resetAllMocks
 
 type ButtonProps = ComponentProps<typeof Button>
 
-describe('Components', () => {
-  describe('Button', () => {
-    const onPressMock = jest.fn()
+describe('Button', () => {
```
klinzo, line 14: Only removed the wrapping in `describe('Components', () => {...})`, same for other tests (except from `DisciplineItem.spec.tsx`).

klinzo:
Will do this for exercises and similar in a separate PR.

f1sh1918 (approved):
i wonder whether there is a way to disable `pressIn` while scrolling to avoid changing background of the item but didn't find a fast solution for that.
But apart of that it works as expected.

Tested on android

klinzo:
Please test this if this feels smooth for you.

SteffiStuffel (commented):
Please rename branch (wrong prefix).

Tested, does not work as expected. I though the color should change when to blue "onPressIn" and the navigation should start when "onPressOut"?

src/components/DisciplineItem.tsx
```diff
@@ -1,60 +1,101 @@
...
   onPress: () => void
 }
 
-const DisciplineItem = ({ selected, onPress, item, children }: DisciplineItemProps): JSX.Element => {
-  const { icon, title } = item
+const DisciplineItem = ({ onPress, icon, title, description, badgeLabel }: DisciplineItemProps): ReactElement => {
+  const [pressed, setPressed] = useState<boolean>(false)
+  const onItemPress = () => {
+    setPressed(true)
+    setTimeout(() => setPressed(false), PRESS_ANIMATION_DURATION)
```
SteffiStuffel, line 82: It think we should use onPressOut here.

klinzo:
Wrong branch name, will reopen.

PR #239, opened 2022-02-16
LUN-251: Render with Theme function

Taggotty:


klinzo (approved):

SteffiStuffel (approved):
Have you thought about moving `wrapWithTheme` into src/testing/render.ts, so it cannot be imported/used in future tests?

PR #240, opened 2022-02-16
LUN-253: Change discipline item onPress behavior

klinzo:

Reopened #235


src/components/DisciplineItem.tsx
```diff
@@ -1,60 +1,101 @@
...
+const DescriptionContainer = styled.View`
+  flex-direction: row;
+  align-items: center;
+`
+
+const Description = styled.Text<{ pressed: boolean }>`
+  font-size: ${props => props.theme.fonts.defaultFontSize};
+  font-weight: ${props => props.theme.fonts.lightFontWeight};
+  font-family: ${props => props.theme.fonts.contentFontRegular};
+  color: ${props => (props.pressed ? props.theme.colors.lunesWhite : props.theme.colors.lunesGreyMedium)};
```
klinzo, line 51: The color is now `lunesWhite` (`white` before). Not sure if that is correct but in this screen it is that way and was already implemented like this for `DisciplineSelectionScreen`: https://app.zeplin.io/project/5fc7c98a435298126bcfd4ee/screen/5fd7ad2c3cfec77f5c68f10c

Apart from that styles are only copied.

klinzo:
@SteffiStuffel 
> Tested, does not work as expected. I though the color should change when to blue "onPressIn" and the navigation should start when "onPressOut"?

I chose this behavior after all because the behavior when attempting to scroll was really weird. If you liked the other behavior better, I could just revert my last commit. Please try both and let me know.

SteffiStuffel:
Check out this: 
`
    <Container
      onLongPress={() => setPressed(true)}
      delayLongPress={20}
      pressed={pressed}
      onPressOut={() => {
        if (pressed) {
          setPressed(false)
          onPress()
        }
      }}>
`

I think this is what we want.

klinzo:
> Check out this:
> 
> ```
>     <Container
>       onLongPress={() => setPressed(true)}
>       delayLongPress={20}
>       pressed={pressed}
>       onPressOut={() => {
>         if (pressed) {
>           setPressed(false)
>           onPress()
>         }
>       }}>
> ```
> 
> I think this is what we want.

Does not really work for me :( Scrolling triggers a navigation now and very short presses are not handled.

SteffiStuffel:
> > Check out this:
> > ```
> >     <Container
> >       onLongPress={() => setPressed(true)}
> >       delayLongPress={20}
> >       pressed={pressed}
> >       onPressOut={() => {
> >         if (pressed) {
> >           setPressed(false)
> >           onPress()
> >         }
> >       }}>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think this is what we want.
> 
> Does not really work for me :( Scrolling triggers a navigation now and very short presses are not handled.

If that is not enough we need to listen to scroll event and cancel click event if scrolled.

src/components/DisciplineItem.tsx
```diff
@@ -1,60 +1,101 @@
...
+`
+
+const BadgeLabel = styled.Text<{ pressed: boolean }>`
+  font-family: ${props => props.theme.fonts.contentFontBold};
+  font-weight: ${props => props.theme.fonts.defaultFontWeight};
+  min-width: ${wp('6%')}px;
+  height: ${wp('4%')}px;
```
f1sh1918, line 58: you shouldn't set a height on a text. It should depend on font-size and paddings.
Since you didn't implement that its optional

f1sh1918 (approved):

SteffiStuffel (changes requested):
I think if we change the scroll/click behaviour we should do it, as it should be and no "quick fix"

klinzo:
> I think if we change the scroll/click behaviour we should do it, as it should be and no "quick fix"

Sure, agreed, but please tell me how it should be then because honestly I don't know. The current solution is imo the best I tried yet and not really a hack. However, I am open to try different things if you tell me what exactly you want, but right now I have no idea how to make it better. There is also the solution without my last commit if you like that better. Your proposal above does not work because of several things as mentioned above. Things to consider in a solution:
* Scrolling should continue to work, i.e. it should not navigate to a discipline (with or without triggering a color change)
* Short presses should continue to work, no matter how short the press duration is
* Long presses can or can not be handled depending on preference

klinzo:
If it is okay for you I'd also do adjustments in a separate PR to avoid more merge conflicts...

SteffiStuffel:
> > I think if we change the scroll/click behaviour we should do it, as it should be and no "quick fix"
> 
> Sure, agreed, but please tell me how it should be then because honestly I don't know. The current solution is imo the best I tried yet and not really a hack. However, I am open to try different things if you tell me what exactly you want, but right now I have no idea how to make it better. There is also the solution without my last commit if you like that better. Your proposal above does not work because of several things as mentioned above. Things to consider in a solution:
> 
> * Scrolling should continue to work, i.e. it should not navigate to a discipline (with or without triggering a color change)
> * Short presses should continue to work, no matter how short the press duration is
> * Long presses can or can not be handled depending on preference

Isn't on the current main just the behaviour we want?

SteffiStuffel:
Here is a working solution for the HomeScreen, nothing finished to merge, just to show how we can achieve the desired behaviour: 
https://github.com/digitalfabrik/lunes-app/pull/247

SteffiStuffel:
> If it is okay for you I'd also do adjustments in a separate PR to avoid more merge conflicts

I don't think it makes sense to merge this, as it is not the changes / behaviour we want.

klinzo:
@SteffiStuffel just tried your approach and while it works fine, the only difference I am experiencing in behavior is that long presses do not lead to a changing of the background color in my solution and the timeout is different. Am I missing something here?
If these two things are your reason for requesting changes here I can change both of these things. Since your approach is much more complicated imo (using refs, handling behavior in parent component/scroll view, using multiple callbacks) I'd like to stick to mine. Just let me know what exactly my solution is lacking and I will try to come up with a solution. Just saying it "is not the behavior we want" is not very clear :see_no_evil: 

Apart from that, I think we should discuss the long press behavior again since I am not really sure how much intuitive it is to give the user visual feedback on long press and then do nothing.

src/components/DisciplineItem.tsx
klinzo: Not changed by me, will keep the current implementation.

SteffiStuffel:
Sorry for not being clear enough here, I thought the previous discussions make this clear.

The wanted behaviour is actually what many other apps also do, e.g. Android system settings or youtube app. This is also the behaviour the `TouchableHighlight` or `TouchableOpacity` of RN have for changing styling.

1. When clicking on an item = onClickIn: Color changes
2. When release the click = onClickOut: The navigation starts and the item stays dark until the navigation is finished.
3. While the item is clicked/touched the color is changed = dark
4. When clicking on an item (color should change) and scroll afterwards (color should change back to light)
5. When scrolling the color should not change, but it is not always possible, as in the moment a users clicks on an item, we do not know if this is to start a scroll or if it stays a simple onClick-event. Therefore we should set a timeout which is a good middle course.


Currently the solution does not fulfill 1 and 3. 
The other (previous) does not fulfill 2 and 5.

I'm fine with other (more simple) solutions as mine, also this was only for demonstration purpose, no clean code that could/should be merged.

src/routes/DisciplineSelectionScreen.tsx
```diff
@@ -78,29 +40,25 @@ const DisciplineSelectionScreen = ({ route, navigation }: DisciplineSelectionScr
...
-    return (
-      <DisciplineItem selected={selected} item={item} onPress={() => handleNavigation(item)}>
-        <ItemText>
-          <BadgeLabel selected={selected}>{item.numberOfChildren}</BadgeLabel>
-          <Description selected={selected}>{childrenLabel(item)}</Description>
-        </ItemText>
-      </DisciplineItem>
-    )
-  }
+  const relevantDisciplines = disciplines?.filter(it => it.numberOfChildren > 0)
```
SteffiStuffel, line 53: (optional) maybe naming can be improved here, e.g. nonEmptyDisciplines

SteffiStuffel (approved):
Tested. Works nice, way better than before :)

PR #242, opened 2022-02-16
LUN-249: Fix jest warning and fix open handles

Taggotty:

Before there were 15 open Handles detected. 
All calls to setTimeout should be mocked with `jest.useFakeTimer`. Alternatively you can also mock the component, that calls `setTimeout`, such as in the `react-native-popover-view`.


.eslintrc.js
```diff
@@ -86,6 +86,7 @@ module.exports = {
         'no-magic-numbers': 'off',
         '@typescript-eslint/no-var-requires': 'off',
         '@typescript-eslint/no-non-null-assertion': 'off',
+        '@typescript-eslint/no-floating-promises': 'off',
```
Taggotty, line 89: jest will warn, that a sync call to act should not be awaited. So I turned this rule off in tests

src/components/__tests__/App.spec.tsx
```diff
@@ -10,16 +10,14 @@ jest.mock('../../navigation/Navigator', () => {
   const Text = require('react-native').Text
   return () => <Text>Navigator</Text>
 })
-jest.useFakeTimers('modern')
+jest.useFakeTimers()
```
klinzo, line 13: Can't we just move this in `jest.setup`?

klinzo (approved):

f1sh1918 (approved):

SteffiStuffel (approved):
Nice.

PR #243, opened 2022-02-16
LUN-256: Added QR code scanner for ios

f1sh1918:

Tested on ios and android

### DIFF
--- a/ios/Lunes/Info.plist
+++ b/ios/Lunes/Info.plist
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+    <key>NSCameraUsageDescription</key>
+    <string>Um Barcodes zu scannen, benötigt LUNES Zugriff auf ihre Kamera</string>
 	<key>UIUserInterfaceStyle</key>
 	<string>Light</string>
 	<key>CFBundleDevelopmentRegion</key>
@@ -25,7 +27,7 @@
 	<key>CFBundleVersion</key>
 	<string>$(CURRENT_PROJECT_VERSION)</string>
 	<key>ITSAppUsesNonExemptEncryption</key>
-    <false/>
+	<false/>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
 	<key>NSAppTransportSecurity</key>
--- a/ios/Lunes/Info.plist
+++ b/ios/Lunes/Info.plist
@@ -3,7 +3,7 @@
 <plist version="1.0">
 <dict>
     <key>NSCameraUsageDescription</key>
-    <string>Zum Scannen von QR-Codes benötigt LUNES Zugriff auf Ihre Kamera.</string>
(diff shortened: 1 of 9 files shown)


ios/Lunes/Info.plist
```diff
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+    <key>NSCameraUsageDescription</key>
+    <string>Um Barcodes zu scannen, benötigt LUNES Zugriff auf ihre Kamera</string>
```
klinzo, line 6:
```suggestion
    <string>Zum Scannen von QR-Codes benötigt LUNES Zugriff auf Ihre Kamera.</string>
```

src/constants/labels.json
```diff
@@ -22,7 +22,7 @@
     },
     "qrCodeScanner": {
       "noAuthorization": {
-        "description": "Kamerazugriff zum Scannen von QR-Codes erforderlich. Öffnen Sie die Android-Einstellungen, tippen Sie auf Berechtigungen und tippen Sie auf Erlauben.",
+        "description": "Kamerazugriff zum Scannen von QR-Codes erforderlich. Öffnen Sie die Einstellungen, tippen Sie auf Berechtigungen und tippen Sie auf Erlauben.",
```
klinzo, line 25: I guess this only applies for android, not sure how you can grant permissions in iOS.

f1sh1918: i mean we could adjust these text but there is a `Einstellungen` Button below the text and then you reach the settings section on each platform

ios/Lunes/Info.plist
```diff
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+    <key>NSCameraUsageDescription</key>
+    <string>Zum Scannen von QR-Codes benötigt LUNES Zugriff auf Ihre Kamera.</string>
```
SteffiStuffel, line 6: Lunes uses informal form of adress. So should be `deine Kamera`

SteffiStuffel (approved):
Tested on browserstack ios: Is not working, but may no be due to you changes, but because of browserstack.
Tested on android real devices: still works fine.

src/constants/labels.json
klinzo: Even though this is incorrect on iOS, it is fine for me. There is no such thing as `Berechtigungen` or `Erlauben`. You just toggle the switch `Kamera`. If you open the settings directly and not from the button in lunes, you just need to select `Lunes` and toggle the switch.

f1sh1918:
> Tested on browserstack ios: Is not working, but may no be due to you changes, but because of browserstack. Tested on android real devices: still works fine.

works on real ios devices. browserstack also crashes for me

f1sh1918:
> Tested on iOS simulator, permissions are handled correctly. I could not scan a QR code obviously but I think that should work just fine. At least on the simulator (iPhone 12 Pro Max) the close button on the QR code scanner is cut and not pressable/not working.

fixed this by adding `SafeAreaView` and adjusted Icon positioning

klinzo (approved):

PR #244, opened 2022-02-17
LUN-246: Improve responsiveness for tablets

f1sh1918:

For this issue i had to introduce spacing classes and refactor a lot. Please test properly on tablet and mobile phone
There is still some space for improvements on tablets but i think we don't have that huge focus on it

More style refactoring gonna be done after style definitions
https://issues.tuerantuer.org/browse/LUN-259


src/components/Button.tsx
```diff
@@ -25,13 +25,13 @@ const ThemedButton = styled.TouchableOpacity<ThemedButtonProps>`
...
     `};
   flex-direction: row;
-  padding: 10px 20px;
+  padding: ${props => `${wp('3%')}px ${props.theme.spacings.sm}`};
+
   width: ${wp('70%')}px;
   align-items: center;
   border-radius: ${hp('7%')}px;
   justify-content: center;
-  height: ${hp('7%')}px;
```
f1sh1918, line 33: you shouldn't set height on a button it should depend on font-size & padding

src/constants/theme/spacings.ts
```diff
@@ -0,0 +1,12 @@
...
+// These pixels are calculated for a 400px width mobile screen
+export const SPACINGS = {
+  xxs: `${wp('1%')}px`, // 4px
+  xs: `${wp('2%')}px`, // 8px
+  sm: `${wp('4%')}px`, // 16px
+  md: `${wp('6%')}px`, // 24px
+  lg: `${wp('8%')}px`, // 32px
+  xl: `${wp('10%')}px`, // 40px
+  xxl: `${wp('12%')}px` // 48px
+}
```
f1sh1918, line 12: Maybe there are some spacings missing, maybe we should discuss with ulrike and nikola but i would like to avoid using custom spacings as far as possible

src/navigation/Navigator.tsx
```diff
@@ -34,7 +35,8 @@ const styles = StyleSheet.create({
     shadowOpacity: 0,
     elevation: 0,
     borderBottomColor: COLORS.disabled,
-    borderBottomWidth: 1
+    borderBottomWidth: 1,
+    height: wp('15%')
```
SteffiStuffel, line 39: height depending on the width. Wouldn't `hp(..)` make more sense here?

SteffiStuffel (approved):
Tested on Android. Still looks good on smartphone-size.

src/navigation/Navigator.tsx
f1sh1918: the problem is that font, spacings and everything depend on the wp, so for tablets its increasing a lot and doesn't fit with hp for tablet and mobile

image

klinzo (approved):
Header looks rather small here:
image

src/components/Button.tsx
```diff
@@ -25,13 +25,13 @@ const ThemedButton = styled.TouchableOpacity<ThemedButtonProps>`
       border-width: 1px;
     `};
   flex-direction: row;
-  padding: 10px 20px;
+  padding: ${props => `${wp('3%')}px ${props.theme.spacings.sm}`};
```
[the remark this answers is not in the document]
f1sh1918: looks weird with 2% or 4% unfortunately i would create a refactor ticket if we have all the sizes from design. Maybe we get another spacing with 3%

f1sh1918:
> 

check last commit

f1sh1918:
> Header looks rather small here: image

check last commit

PR #250, opened 2022-03-01
LUN-33: Setup Sentry

SteffiStuffel:


src/services/sentry.ts
```diff
@@ -0,0 +1,47 @@
+/* eslint-disable no-console */
```
SteffiStuffel, line 1: Copied from Integreat.

tools/sentry-release.sh
```diff
@@ -0,0 +1,151 @@
+#!/usr/bin/env bash
```
SteffiStuffel, line 1: Copied from Integreat.

.circleci/config.yml
```diff
@@ -157,6 +157,16 @@ jobs:
                 command: bundle exec fastlane build version_name:${NEW_VERSION_NAME} version_code:${NEW_VERSION_CODE}
                 name: '[FL] Build Android'
                 working_directory: android
+            - run:
+                command: |
+                    mkdir -p sourcemaps/
+                    cp android/app/build/generated/sourcemaps/react/release/index.android.bundle.map sourcemaps/
+                    cp android/app/build/generated/assets/react/release/index.android.bundle sourcemaps/
```
maxammann, line 164:
```suggestion
                    cp android/app/build/generated/sourcemaps/react/release/index.android.bundle* sourcemaps/
```

more similar to below

src/components/Header.tsx
```diff
@@ -48,25 +49,38 @@ const SmileIconStyle = styled.View`
...
+const Header = (): JSX.Element => {
+  const [counter, setCounter] = useState<number>(0)
+  const CLICKS_TO_ENABLE_SENTRY = 20
+
+  const onPress = () => {
+    if (counter > CLICKS_TO_ENABLE_SENTRY) {
+      reportError('Error for testing Sentry')
+    }
+    setCounter(oldVal => oldVal + 1)
+  }
```
maxammann, line 61: Could be extracted to a separate component

src/services/sentry.ts
```diff
@@ -0,0 +1,47 @@
...
+      case Sentry.Severity.Log:
+        console.log(message)
+        break
+      case Sentry.Severity.Info:
+        console.info(message)
+        break
+      case Sentry.Severity.Debug:
+        console.debug(message)
+        break
+    }
```
maxammann, line 32: Is there no function from sentry which accepts a log level?

Also shouldn't the `level` variable have a proper type? I would suspect something like 'debug' | 'error'

tools/sentry-release.sh
maxammann: Minor: Could be refactored to a .ts file

maxammann (approved):
Looks good!

src/navigation/Navigator.tsx
```diff
@@ -59,6 +60,8 @@ const Navigator = (): JSX.Element => {
   const [isPressed, setIsPressed] = React.useState<boolean>(false)
   const [isHomeButtonPressed, setIsHomeButtonPressed] = React.useState<boolean>(false)
 
+  initSentry()
```
klinzo, line 63: I think a better place for this would be App.tsx. In Integreat we just have this in the navigator because it was customizable in the intro slides whether ti should/should not be enabled.

src/services/sentry.ts
klinzo: Copied from integreat which was implemented by me. I struggled with the correct typing. But there probably really is a better way to do this.

klinzo (approved):
Tested on android, error report got sent to sentry (without release but that is probably because I used a development build).

src/services/sentry.ts
SteffiStuffel: As it is working/fine for Integreat and this is just the setup i will keep it, if there is a need to change i can do in a separate ticket.

.circleci/config.yml
SteffiStuffel: Sorry i don't get it. Once the path is `android/app/build/generated/**assets**/...` and once it is `android/app/build/generated/**sourcemaps**/..`. So i cannot just replace line 163 and line 164 with the solution you suggested, can I?

src/components/SentryTestPressable.tsx
```diff
@@ -0,0 +1,24 @@
...
+  children: ReactElement
+}
+
+const SentryTestPressable = ({ children }: Props): ReactElement => {
+  const [counter, setCounter] = useState<number>(0)
+  const CLICKS_TO_ENABLE_SENTRY = 20
+
+  const onPress = () => {
+    if (counter > CLICKS_TO_ENABLE_SENTRY) {
+      reportError('Error for testing Sentry')
```
klinzo, line 16: Perhaps it makes sense to reset it to zero after sending one report. Otherwise if you don't count the clicks you'll send multiple reports.

.circleci/config.yml
maxammann: oh sorry, did not see it!

src/components/SentryTestPressable.tsx
maxammann: @SteffiStuffel 

Oh sorry, I did oversee this. I would suggest to throw here a native (Android/Swift) or JS exception instead. Else it is not possible to test the whole stack.

So the app should really crash in this example. If the app actually stops, then crash reports get sent with the next start of the app. This should also be testable.

I'm not sure how we do it in Integreat, but somehow we are really crashing the Integreat app.

maxammann: Something like this: https://github.com/digitalfabrik/integreat-app/blob/f21b485038305188d7041f03f380ce3a592f6e43/native/src/utils/createSettingsSections.ts#L149

src/components/SentryTestPressable.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import { reportError } from '../services/sentry'
+
+interface Props {
+  children: ReactElement
+}
+
+const SentryTestPressable = ({ children }: Props): ReactElement => {
+  const [counter, setCounter] = useState<number>(0)
+  const CLICKS_TO_ENABLE_SENTRY = 20
```
maxammann, line 12: Shouldn't this be named `CLICKS_TO_TEST_SENTRY`?

src/components/SentryTestPressable.tsx
SteffiStuffel: Added it.

PR #255, opened 2022-03-09
LUN-252-2: Fix failing build android step

SteffiStuffel:

### DIFF
--- a/android/settings.gradle
+++ b/android/settings.gradle
@@ -2,17 +2,4 @@ rootProject.name = 'lunes-app-android'
 
 apply from: file("../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesSettingsGradle(settings)
 
-include ':@react-native-community_async-storage'
-project(':@react-native-community_async-storage').projectDir = new File(rootProject.projectDir, '../node_modules/@react-native-community/async-storage/android')
-include ':react-native-sound-player'
-project(':react-native-sound-player').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-sound-player/android')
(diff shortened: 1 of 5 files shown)


android/settings.gradle
```diff
@@ -2,17 +2,4 @@ rootProject.name = 'lunes-app-android'
 
 apply from: file("../node_modules/@react-native-community/cli-platform-android/native_modules.gradle"); applyNativeModulesSettingsGradle(settings)
 
-include ':@react-native-community_async-storage'
```
klinzo, line 5: Why did you remove those?

tsconfig.json
```diff
@@ -17,7 +17,5 @@
     "types": ["jest", "node", "react", "styled-components-react-native"],
     "skipLibCheck": true
   },
-  "exclude": [
-    "node_modules", "babel.config.js", "metro.config.js", "jest.config.js", "tools"
-  ]
+  "include": ["src/**/*", "./jest.setup.ts", "index.js", "jest.config.ts", "metro.config.ci.js", "metro.config.js"]
```
klinzo, line 20: This e.g. excludes `App.tsx`, right? Why not exclude like we did before?

klinzo (approved):
Not tested

android/settings.gradle
SteffiStuffel: Because it is not needed.

tsconfig.json
SteffiStuffel: No, `**` matches zero or more directories. Changed it so it is more similar to integreat.

f1sh1918 (approved):

PR #257, opened 2022-03-14
LUN-30: Generate mattermost notifications

SteffiStuffel:


.circleci/src/commands/notify.yml
```diff
@@ -0,0 +1,79 @@
+description: Send a notification (to Mattermost) at the end of a job, based on success or failure. Must be the last step in a job.
+parameters:
+  success_message:
+    default: ''
+    type: string
+  failure_mentions:
+    default: '@max.ammann @stefanie.metzger @sarah.sporck @steffen.kleinle'
```
f1sh1918, line 7: i think steffen added me recently to integreat app, i think it would make sense here, too

f1sh1918 (approved):
Looks good so far even i probably can not test it

.circleci/src/jobs/check.yml
```diff
@@ -4,6 +4,7 @@ steps:
   - checkout
   - check_config
   - restore_yarn_cache
+  - restore_yarn_tools_cache
```
klinzo, line 7: Why do we need tools for checks?

.circleci/src/jobs/bump_version.yml
```diff
@@ -12,15 +12,19 @@ steps:
   - restore_yarn_tools_cache
```
klinzo, line 12: Perhaps it make sense to adjust this accordingly to integreat (change enum `delivery` to boolean `prepare_delivery` param

klinzo (approved):
Obviously not tested but it looks like you already did that. Only reviewed very superficial since most was probably copied from integreat.

.circleci/src/commands/notify.yml
```diff
@@ -0,0 +1,79 @@
+description: Send a notification (to Mattermost) at the end of a job, based on success or failure. Must be the last step in a job.
```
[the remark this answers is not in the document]
SteffiStuffel: yes, mostly.

.circleci/src/jobs/check.yml
SteffiStuffel: hmh, though without I got an error, will remove and test again, shouldn't be needed :thinking:

.circleci/src/jobs/bump_version.yml
SteffiStuffel: As this wasn't changed in integreat, when I was starting to work on this ticket, I would say out of scope, but if you think this change is important, feel free to open a ticket.

.circleci/src/jobs/check.yml
SteffiStuffel: Oke, this is needed as ts:check also checks the tools-folder.

PR #262, opened 2022-03-22
LUN-272: Incorrect names for additional areas

SteffiStuffel:

I fixed several things here:
1) Wrong names of multiple custom disciplines: axios-cache did handle them the same way as they only differ in the Auth-header.
2) Jumping of custom disciplines, because container while loading has different size/margins
3) Error message above Title in a module, when loading fails which looks bad, see here https://issues.tuerantuer.org/browse/LUN-271

api-key for testing:
rXTIfCwz.eOYXwwn486fR5PlKFW9KagLtIzwwBCrE
qYA.YYyZben

### DIFF
--- a/release-notes/unreleased/LUN-272-incorrect-names-for-additional-areas.yml
+++ b/release-notes/unreleased/LUN-272-incorrect-names-for-additional-areas.yml
@@ -1,5 +1,5 @@
 issue_key: LUN-272
-show_in_stores: true
+show_in_stores: false
 platforms:
   - android
   - ios
--- a/release-notes/unreleased/LUN-272-incorrect-names-for-additional-areas.yml
+++ b/release-notes/unreleased/LUN-272-incorrect-names-for-additional-areas.yml
@@ -1,0 +1,6 @@
+issue_key: LUN-272
+show_in_stores: true
+platforms:
+  - android
+  - ios
+de: Fehlerhalfte Titel von individuellen Bereichen behoben
--- a/src/components/ListItem.tsx
+++ b/src/components/ListItem.tsx
@@ -1,25 +1,29 @@
 import React, { ReactElement, useCallback, useState } from 'react'
-import { GestureResponderEvent, Pressable } from 'react-native'
+import { GestureResponderEvent } from 'react-native'
 import { heightPercentageToDP as hp, widthPercentageToDP as wp } from 'react-native-responsive-screen'
 import styled, { useTheme } from 'styled-components/native'
 
 import { ChevronRight } from '../../assets/images'
 import { ContentSecondaryLight } from './text/Content'
 
-const Container = styled(Pressable)<{ pressed: boolean }>`
-  min-height: ${hp('12%')}px;
+export const GenericListContainer = styled.Pressable`
   width: ${wp('90%')}px;
   margin-bottom: ${props => props.theme.spacings.xxs};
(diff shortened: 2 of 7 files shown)


release-notes/unreleased/LUN-272-incorrect-names-for-additional-areas.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: LUN-272
+show_in_stores: true
+platforms:
+  - android
+  - ios
+de: Fehlerhalfte Titel von individuellen Bereichen behoben
```
klinzo, line 6: Was this on production? Otherwise I'd set `show_in_stores: false`

src/components/ListItem.tsx
```diff
@@ -1,25 +1,29 @@
...
+import { GestureResponderEvent } from 'react-native'
 import { heightPercentageToDP as hp, widthPercentageToDP as wp } from 'react-native-responsive-screen'
 import styled, { useTheme } from 'styled-components/native'
 
 import { ChevronRight } from '../../assets/images'
 import { ContentSecondaryLight } from './text/Content'
 
-const Container = styled(Pressable)<{ pressed: boolean }>`
-  min-height: ${hp('12%')}px;
+export const GenericListContainer = styled.Pressable`
```
klinzo, line 9:
```suggestion
export const GenericListItemContainer = styled.Pressable`
```

src/services/axios.ts
```diff
@@ -1,11 +1,21 @@
...
 import { addTrailingSlashToUrl } from './url'
 
 const baseURL = __DEV__ ? 'https://lunes-test.tuerantuer.org/api' : 'https://lunes.tuerantuer.org/api'
 
-setupCache(axios)
+const keyGenerator = buildKeyGenerator(true, ({ headers, baseURL = '', url = '', method = 'get', params, data }) => ({
+  url: baseURL + (baseURL && url ? '/' : '') + url,
+  headers: headers?.Authorization ?? 'not-set',
+  method,
+  params: params as unknown,
```
klinzo, line 12: Why `as unknown`? Is it not possible to correctly type this?

release-notes/unreleased/LUN-272-incorrect-names-for-additional-areas.yml
SteffiStuffel: yes it was.

src/services/axios.ts
SteffiStuffel: No I guess not. You can not know what is going to be passed for params. It's the same code as the defaultKeyGenerator form axios-cache-interceptor uses.

release-notes/unreleased/LUN-272-incorrect-names-for-additional-areas.yml
klinzo: You sure? axios-cache-interceptor was only added in februrary, but at least the version from the play store is from january:
image
image

klinzo (approved):
Tested on android

release-notes/unreleased/LUN-272-incorrect-names-for-additional-areas.yml
SteffiStuffel: hmh, you are right :thinking: 
Thought Julia found this on live app. Will remove.

PR #266, opened 2022-03-29
LUN-274: Select profession on initial start

SteffiStuffel:


src/navigation/Navigator.tsx
```diff
@@ -56,13 +60,25 @@ const styles = (headerHeight?: number) =>
...
 
+  useEffect(() => {
+    const initialize = async () => {
+      const professions = await AsyncStorage.getSelectedProfessions()
+      setIntroShown(professions !== null)
+    }
+    initialize()
+      .then(() => setWaitingForInit(false))
+      .catch(error => reportError(error))
+  }, [])
```
klinzo, line 80: Why do we need a custom useEffect here instead of just using `useReadSelectedProfession`? Then you can just do `const introShown = professions !== null`

src/navigation/Navigator.tsx
```diff
@@ -56,13 +60,25 @@ const styles = (headerHeight?: number) =>
...
+  const [waitingForInit, setWaitingForInit] = useState<boolean>(true)
+  const [introShown, setIntroShown] = useState<boolean>(false)
 
   // Set only height for tablets since header doesn't scale auto
   const headerHeight = useTabletHeaderHeight(wp('15%'))
 
+  useEffect(() => {
+    const initialize = async () => {
+      const professions = await AsyncStorage.getSelectedProfessions()
+      setIntroShown(professions !== null)
```
klinzo, line 75: Is it required that a profession gets selected? Maybe it would be more explicit if we saved a separate value `introShown` to the async storage.

I also think that if I don't want to select a profession I should not have to click on `Überspringen` each time I open the app (this fails the alternate course of the issue).

src/routes/intro/IntroScreen.tsx
```diff
@@ -0,0 +1,83 @@
...
+}
+
+const IntroScreen = ({ navigation }: IntroScreenProps): JSX.Element => {
+  const { data: disciplines, error, loading, refresh } = useLoadDisciplines(null)
+  const { data: selectedProfessions, refresh: refreshSelectedProfessions } = useReadSelectedProfessions()
+
+  useFocusEffect(
+    React.useCallback(() => {
+      refreshSelectedProfessions()
+    }, [refreshSelectedProfessions])
```
klinzo, line 44: The wrapping in callback should not be needed as it is already wrapped in a callback in `useLoadAsync`

src/routes/intro/IntroScreen.tsx
```diff
@@ -0,0 +1,83 @@
...
+  margin-bottom: ${props => props.theme.spacings.lg};
+`
+
+interface IntroScreenProps {
+  navigation: StackNavigationProp<RoutesParams, 'Intro'>
+}
+
+const IntroScreen = ({ navigation }: IntroScreenProps): JSX.Element => {
+  const { data: disciplines, error, loading, refresh } = useLoadDisciplines(null)
+  const { data: selectedProfessions, refresh: refreshSelectedProfessions } = useReadSelectedProfessions()
```
klinzo, line 39: I think we have some inconsistent naming again now. Profession vs Discipline. Don't we use that the same way?

klinzo, line 39: Why don't you handle error and loading the same way as those from `useLoadDiscipline`?

src/services/AsyncStorage.ts
```diff
@@ -1,5 +1,41 @@
...
+export const getSelectedProfessions = async (): Promise<Discipline[] | null> => {
+  const professions = await AsyncStorage.getItem('selectedProfessions')
+  return professions ? JSON.parse(professions) : null
+}
+
+export const setSelectedProfessions = async (selectedProfessions: Discipline[]): Promise<void> => {
+  await AsyncStorage.setItem('selectedProfessions', JSON.stringify(selectedProfessions))
+}
+
+export const pushSelectedProfession = async (profession: Discipline): Promise<Discipline[]> => {
```
klinzo, line 14: Some unit tests would be nice here

src/routes/intro/__tests__/IntroScreen.spec.tsx
```diff
@@ -0,0 +1,55 @@
...
+import IntroScreen from '../IntroScreen'
+
+jest.mock('@react-navigation/native')
+jest.mock('../../../hooks/useLoadDisciplines')
+jest.mock('../../../hooks/useReadSelectedProfessions')
+
+describe('IntroScreen', () => {
+  const navigation = createNavigationMock<'Intro'>()
+
+  it('should navigate to discipline', async () => {
```
klinzo, line 21:
```suggestion
  it('should navigate to profession selection of discipline', async () => {
```

src/routes/intro/__tests__/IntroScreen.spec.tsx
```diff
@@ -0,0 +1,55 @@
...
+    mocked(useLoadDisciplines).mockReturnValueOnce(getReturnOf(mockDisciplines))
+    mocked(useReadSelectedProfessions).mockReturnValueOnce(getReturnOf(null))
+
+    const { findByText } = render(<IntroScreen navigation={navigation} />)
+    const firstDiscipline = await findByText('First Discipline')
```
klinzo, line 26: No need for `await findByText`, `getByText` is enough

src/routes/intro/__tests__/ProfessionSelectionScreen.spec.tsx
```diff
@@ -0,0 +1,73 @@
...
+    const spyOnPush = jest.spyOn(AsyncStorageService, 'pushSelectedProfession')
+
+    const { findByText, queryAllByTestId } = render(<ProfessionSelectionScreen route={route} navigation={navigation} />)
+    expect(await findByText(labels.intro.skipSelection)).toBeDefined()
+    const profession = await findByText(mockDisciplines[0].title)
+    expect(profession).toBeDefined()
+    expect(queryAllByTestId('check-icon')).toHaveLength(0)
+    fireEvent.press(profession)
+    expect(spyOnPush).toHaveBeenCalledTimes(1)
+    expect(spyOnPush).toHaveBeenCalledWith(mockDisciplines[0])
```
klinzo, line 53: Optional to avoid spies:
```suggestion
    await waitFor(async () => {
      const selectedProfessions = await AsyncStorageService.getSelectedProfessions()
      expect(selectedProfessions).toEqual([mockDisciplines[0]])
    })

```

src/routes/intro/ProfessionSelectionScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+              <ButtonContainer>
+                <Button
+                  onPress={navigateToHomeScreen}
+                  label={labels.intro.skipSelection}
+                  buttonTheme={BUTTONS_THEME.contained}
+                />
+              </ButtonContainer>
+            ) : (
+              <></>
+            )
```
klinzo, line 99: I think it would be more intuitive (and easier to implement) if the skip button would be implemented the same way as the confirm selection button. Otherwise the skip button is mostly not visible and the user has to scroll down first. In combination with a very unobtrusive/small text that the user should select their professions I think it is kind of hard to grasp what the user has to do.

klinzo (changes requested):

src/navigation/Navigator.tsx
SteffiStuffel: I see your point here, but not sure if it is possible to do. Profession is null, either when a) intro was not shown before or b) when profession is not yet loaded (when using useReadSelectedProfession). Therefore we need to check for the Promise to return, to set waitingForInti(false) correctly.

src/hooks/useReadSelectedProfessions.ts
```diff
@@ -0,0 +1,9 @@
+import { Discipline } from '../constants/endpoints'
+import AsyncStorage from '../services/AsyncStorage'
+import useLoadAsync, { Return } from './useLoadAsync'
+
+const loadFromAsyncStorage = async (): Promise<Discipline[] | null> => AsyncStorage.getSelectedProfessions()
```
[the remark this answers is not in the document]
SteffiStuffel: Done, also for useReadCustomDisicplines.

src/hooks/useReadSelectedProfessions.ts
```diff
@@ -0,0 +1,9 @@
...
+
+const useReadFromAsyncStorage = (): Return<Discipline[] | null> => useLoadAsync(loadFromAsyncStorage, null)
```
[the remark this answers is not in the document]
SteffiStuffel: I think as it is/ can be used in different parts of the code it is good to have it in a separate file.

src/navigation/Navigator.tsx
SteffiStuffel: Yes, currently profession is null, when the value was never set. If you skip the selection, an empty array should be saved, therefore the alternate course should work. But I forgot to set this. 

If you think the clearness of a separate value will outweigh that one more slow reading from AsyncStorage we need then, I will add it. Let me know what you prefer.

src/routes/intro/__tests__/IntroScreen.spec.tsx
SteffiStuffel: Why? Because I mocked every async stuff and it is no longer asnyc then?

src/routes/intro/ProfessionSelectionScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { childrenDescription } from '../../services/helpers'
+
+const Root = styled.View`
+  background-color: ${props => props.theme.colors.background};
+  height: 100%;
+`
+
+const StyledList = styled(FlatList)`
+  width: 100%;
+` as ComponentType as new () => FlatList<Discipline>
```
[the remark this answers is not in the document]
SteffiStuffel: hmh, I think you are right, it is not needed. :thinking:

src/routes/intro/ProfessionSelectionScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  const Item = ({ item }: { item: Discipline }): JSX.Element => (
+    <DisciplineListItem
+      item={item}
+      onPress={() => selectDiscipline(item)}
+      hasBadge
+      rightChildren={
+        selectedProfessions?.includes(item) ? (
+          <IconContainer>
+            <CheckCircleIconGreen testID='check-icon' />
```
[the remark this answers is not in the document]
SteffiStuffel: Fixed it.

src/routes/intro/__tests__/ProfessionSelectionScreen.spec.tsx
SteffiStuffel: Done.
I decided to use a spy, as I thought it makes sense for a unit test, to just check if "outgoing" call is valid and not also if the AsyncStorageService is working properly. Why is it bad to use spies?
Best test case, would be to wait for green-check-icon to appear / disapear, but I cannot get this working.

src/routes/intro/ProfessionSelectionScreen.tsx
SteffiStuffel: It was on purpose the "skip" button is kind of "hidden", so the users get encouraged to select a profession. Would keep it for now like this.

f1sh1918 (approved):
Tested on ios. Button "Starten" could need some margin-bottom to avoid accidently clicking on the ios bottom sheet. See suggestion (optional)
image

src/routes/intro/ProfessionSelectionScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+import AsyncStorage from '../../services/AsyncStorage'
+import { childrenDescription } from '../../services/helpers'
+
+const Root = styled.View`
+  background-color: ${props => props.theme.colors.background};
+  height: 100%;
+`
+
+const ButtonContainer = styled.View`
+  padding-top: ${props => props.theme.spacings.md}
```
f1sh1918, line 27:
```suggestion
  padding: ${props => `${props.theme.spacings.md} 0`}
```

src/routes/intro/ProfessionSelectionScreen.tsx
```diff
@@ -0,0 +1,121 @@
...
+            !isProfessionSelected ? (
+              <ButtonContainer>
+                <Button
+                  onPress={navigateToHomeScreen}
+                  label={labels.intro.skipSelection}
+                  buttonTheme={BUTTONS_THEME.contained}
+                />
+              </ButtonContainer>
+            ) : (
+              <></>
```
f1sh1918, line 98: no need to render a fragment if there is no content in it. Instead return `: null`

src/navigation/Navigator.tsx
klinzo: You can just use `loading` property returned by the hook instead of `waitingForInit` or am I missing something?

src/navigation/Navigator.tsx
klinzo: As discussed the current way is fine.

src/routes/intro/__tests__/IntroScreen.spec.tsx
klinzo: yes, exactly.

src/routes/intro/__tests__/ProfessionSelectionScreen.spec.tsx
klinzo: I don't know if it is bad to use spies but I think it kind of tests implementation details which could be avoided here checking if the async storage has the right value afterwards. E.g. it could happen that while the spy is called correctly, another async storage function is called somewhere else overwriting it in the async storage.

src/routes/intro/ProfessionSelectionScreen.tsx
klinzo: Okay, then LGTM.

src/routes/intro/ProfessionSelectionScreen.tsx
```diff
@@ -48,9 +44,10 @@ const ProfessionSelectionScreen = ({ route, navigation }: ProfessionSelectionScr
   const { discipline } = route.params
   const { data: disciplines, error, loading, refresh } = useLoadDisciplines(discipline)
   const { data: selectedProfessions, refresh: refreshSelectedProfessions } = useReadSelectedProfessions()
+  const isProfessionSelected = selectedProfessions && selectedProfessions.length > 0
```
klinzo, line 47: Optional: Imo naming should be plural, otherwise I'd think it is just whether a currently displayed profession is selected and not whether any profession is selected.

src/routes/intro/ProfessionSelectionScreen.tsx
```diff
@@ -74,19 +71,22 @@ const ProfessionSelectionScreen = ({ route, navigation }: ProfessionSelectionScr
     />
   )
 
-  const navigateToHomeScreen = () => {
+  const navigateToHomeScreen = async () => {
+    if (!isProfessionSelected) {
```
klinzo, line 75: Same as in IntroScreen

src/routes/intro/IntroScreen.tsx
```diff
@@ -37,21 +38,21 @@ interface IntroScreenProps {
...
   const navigateToDiscipline = (item: Discipline): void => {
     navigation.navigate('ProfessionSelection', {
       discipline: item
     })
   }
 
-  const navigateToHomeScreen = (): void => {
-    navigation.navigate('Home')
+  const navigateToHomeScreen = async () => {
+    if (!isProfessionSelected) {
```
klinzo, line 52: I'd just check `selectedProfessions === null`, no need to set the selected professions to `[]` if it is already `[]`

src/routes/intro/IntroScreen.tsx
```diff
@@ -37,21 +38,21 @@ interface IntroScreenProps {
...
+      await AsyncStorage.setSelectedProfessions([])
+    }
+    navigation.push('Home')
```
klinzo, line 55: Why push instead of navigate? Same for ProfessionSelectionScreen

klinzo (commented):
There are still some comments that are hidden and where not addressed

klinzo (approved):
Looking good. Please have a look at my comments, you don't have to do everything (some hidden comments not yet addressed from my first review).

src/navigation/Navigator.tsx
SteffiStuffel: Nope, you are right, this is working.

src/routes/intro/IntroScreen.tsx
SteffiStuffel: Yes there is a reason: professions are disciplines that can be selected (it is only possible for second level disciplines of lunes standard vocabulary)

SteffiStuffel: If `useLoadDiscipline` is still loading, we have no data to display, whereas if the `useReadSelectedProfession` is still loading, we can display the data, but only the green-check-icon is missing. Not sure if the reading from AsyncStorage can fail :thinking: At least it is far less common than a failing api-request.

If you want I can open a ticket, where we implement an appropriate error handling.

src/services/AsyncStorage.ts
SteffiStuffel: done.

src/routes/intro/ProfessionSelectionScreen.tsx
SteffiStuffel: done

src/routes/intro/IntroScreen.tsx
SteffiStuffel: Changed to navigate

src/routes/intro/IntroScreen.tsx
SteffiStuffel: you are right. done

SteffiStuffel:
> There are still some comments that are hidden and are not addressed yet (maybe also on purpose, not sure)

I think I addressed all comments. Also fixed the bottom margin @f1sh1918 mentioned.

src/routes/intro/IntroScreen.tsx
klinzo: No, sounds good to me :+1:

klinzo (approved):

PR #274, opened 2022-04-06
LUN-282: Error unable to open url mailto

f1sh1918:

Both fixes are only necessary for emulators to avoid errors. On real devices everything works as expected.

`tel` & `mailto` can only be tested on real ios devices. Just added handling for not supported devices like emulators
Impressum-> Click on mailaddress

### DIFF
--- a/src/services/url.ts
+++ b/src/services/url.ts
@@ -7,7 +7,7 @@ export const openExternalUrl = (url: string): Promise<void> =>
   Linking.canOpenURL(url)
     .then(supported => {
       if (!supported) {
-        // emulators do not support certain schemes like mailto or tel
+        // iOS Simulators do not support certain schemes like mailto or tel
         console.log('Cannot handle url', url)
         return null
       }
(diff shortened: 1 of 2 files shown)


src/services/url.ts
```diff
@@ -5,5 +5,12 @@ export const addTrailingSlashToUrl = (url: string): string => (url.endsWith('/')
 
 export const openExternalUrl = (url: string): Promise<void> =>
   Linking.canOpenURL(url)
-    .then(() => Linking.openURL(url))
+    .then(supported => {
+      if (!supported) {
+        // emulators do not support certain schemes like mailto or tel
```
klinzo, line 10:
```suggestion
        // iOS Simulators do not support certain schemes like mailto or tel
```

src/services/url.ts
```diff
@@ -5,5 +5,12 @@ export const addTrailingSlashToUrl = (url: string): string => (url.endsWith('/')
...
+        console.log('Cannot handle url', url)
```
klinzo, line 11: Our internal logging function should be used (`log` from sentry.ts)

klinzo (approved):

SteffiStuffel (approved):
I think we had this issue on a real device (on claras phone), not sure which on this was. Do you know, @klinzo ?

klinzo:
Not sure anymore, I think the issue on claras phone was something different. Lets see if this happens again in the future

PR #275, opened 2022-04-07
LUN-305: Swipe gesture to delete individual area disfunctional

f1sh1918:

Check Steps to reproduce in https://issues.tuerantuer.org/browse/LUN-305

Since this happens mainly on ios pls check on ios emulator or real device

@klinzo i'm not 100% sure if the test adjustments make sense :)

### DIFF
--- a/src/components/__tests__/ListItem.spec.tsx
+++ b/src/components/__tests__/ListItem.spec.tsx
@@ -32,8 +32,8 @@ describe('ListItem', () => {
     expect(arrowIcon.props.fill).toBe(COLORS.primary)
     expect(getByText(title).instance.props.style[0].color).toBe(COLORS.text)
 
-    fireEvent(arrowIcon, 'pressIn', { nativeEvent: { pageY: 123 } })
-    fireEvent(arrowIcon, 'pressOut', { nativeEvent: { pageY: 123 } })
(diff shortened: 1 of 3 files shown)


src/components/__tests__/ListItem.spec.tsx
```diff
@@ -67,14 +67,14 @@ describe('ListItem', () => {
     })
   })
 
-  it('should not call on press callback if scrolling', async () => {
+  it('should not call on press callback if scrolling or swiping', async () => {
```
klinzo, line 70: I think two separate tests would make sense here, one that scrolling (only change y position) and one that swiping(only change x position) does not trigger on click. On the other hand the swipe to delete functionality is removed in #270 anyway. But I think just copying and adjusting would be easy and won't hurt.

f1sh1918: hm ok i overseen the fact that it gonna be removed. it we don't make any hotfix release. this pr is maybe not needed at all. So i won't put any more effort in tests to be honest

klinzo (approved):
Tested on browserstack

SteffiStuffel (changes requested):
We don't need this changes. Next release will contain the individualized home screen and there won't be this swipe-to-delete component anymore.

f1sh1918:
> We don't need this changes. Next release will contain the individualized home screen and there won't be this swipe-to-delete component anymore.

yes i realized this way too late. gonna close the pr

PR #276, opened 2022-04-08
LUN-280: Show correct solution if wrong

f1sh1918:

Check how to Test in [LUN-280](https://issues.tuerantuer.org/browse/LUN-280)


klinzo (approved):
Tested on android

f1sh1918:
> Tested. Works.
> 
> (optional): Padding of the text on the right is too small, but was like this before.

update paddings. even to fit better to the input field above (article word)

PR #278, opened 2022-04-12
LUN-132: Tidy up

klinzo:

This PR renames `ExercisesScreens` to `ExercisesScreen` (no trailing `s`) and merges `WriteExercise` into `WriteExerciseScreen`.


src/routes/write-exercise/WriteExerciseScreen.tsx
```diff
@@ -1,20 +1,154 @@
 import { RouteProp } from '@react-navigation/native'
```
klinzo, line 1: No need to review, just merged WriteExercise in

SteffiStuffel (approved):
This makes sense. The merging of WriteExercise and WriteExerciseScreen is now possible as the loading of documents is not done there anymore. So it would be the same for other exercises, maybe you can have a look (or already had) if a refactoring would also make sense there, too?

klinzo:
> This makes sense. The merging of WriteExercise and WriteExerciseScreen is now possible as the loading of documents is not done there anymore. So it would be the same for other exercises, maybe you can have a look (or already had) if a refactoring would also make sense there, too?

I did, but I think it only makes sense for the WriteExerciseScreen since for the choice exercises the code is shared and the VocabularyListScreen is way bigger then the WriteExerciseScreen already.

f1sh1918 (approved):

PR #280, opened 2022-04-14
LUN-276: Manage Selection Screen

SteffiStuffel:

Handling of loading and errors will be done in LUN-301.
I not also did LUN-276 (adding of a profession) here, but implemented the hole ManageSelectionScreen, so I also did LUN-277 (removing of a profession). These user-story-tickets as a developer task does not really make sense, so I think it is good, we decided to move the "functional"(=fachlich) user stories into the epic and create developer tasks that make sense for developing, e.g. in this case it should have been "Implement Manage Selection Screen"


src/routes/home/HomeScreen.tsx
```diff
@@ -36,13 +36,14 @@ interface HomeScreenProps {
...
+  const { data: selectedProfessions, refresh: refreshSelectedProfessions } = useReadSelectedProfessions()
   const isCustomDisciplineEmpty = !customDisciplines || customDisciplines.length <= 0
 
   useFocusEffect(
     React.useCallback(() => {
       refreshCustomDisciplines()
-    }, [refreshCustomDisciplines])
+      refreshSelectedProfessions()
+    }, [refreshCustomDisciplines, refreshSelectedProfessions])
   )
```
klinzo, line 47: Optional since not added in this PR: No useCallback should be needed since refreshes are already wrapped in one in the hooks.

src/routes/__tests__/ProfessionSelectionScreen.spec.tsx
```diff
@@ -0,0 +1,102 @@
...
+        icon: 'none',
+        numberOfChildren: 2,
+        isLeaf: false,
+        parentTitle: null,
+        needsTrainingSetEndpoint: false
+      }
+    }
+  })
+
+  it('should select profession when clicked', async () => {
```
klinzo, line 42: Minor: I think it would be better to stick to the term `pressed`

src/routes/manage-disciplines/ManageDisciplinesScreen.tsx
```diff
@@ -0,0 +1,119 @@
...
+const ManageDisciplinesScreen = ({ navigation }: Props): ReactElement => {
+  const {
+    data: selectedProfessions,
+    loading,
+    error,
+    refresh: refreshSelectedProfessions
+  } = useReadSelectedProfessions()
+  const { data: customDisciplines, refresh: refreshCustomDisciplines } = useReadCustomDisciplines()
+
+  const refresh = useCallback(() => {
```
klinzo, line 49: No callback should be needed.

src/routes/manage-disciplines/ManageDisciplinesScreen.tsx
```diff
@@ -0,0 +1,119 @@
...
+    refreshSelectedProfessions()
+  }, [refreshCustomDisciplines, refreshSelectedProfessions])
+
+  React.useEffect(() => {
+    refresh()
+    const willFocusSubscription = navigation.addListener('focus', () => {
+      refresh()
+    })
+    return willFocusSubscription
+  }, [navigation, refresh])
```
klinzo, line 60: Any reason why you are not using the focus effect from reactnavigation?
Minor: Unnecessary wrapping of refresh in the listener and no need for a const, can be returned directly.

src/routes/manage-disciplines/ManageDisciplinesScreen.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  const Item = ({ item }: { item: Discipline }): JSX.Element => {
+    const unselectProfessionAndRefresh = (item: Discipline) => {
+      removeSelectedProfession(item)
+        .then(refreshSelectedProfessions)
+        .catch(err => reportError(err))
```
klinzo, line 66:
```suggestion
        .catch(reportError)
```

src/routes/manage-disciplines/ManageDisciplinesScreen.tsx
```diff
@@ -0,0 +1,119 @@
...
+          <Pressable onPress={() => unselectProfessionAndRefresh(item)} testID='delete-icon'>
+            <CloseIconRed />
+          </Pressable>
+        }
+      />
+    )
+  }
+
+  const navigateToProfessionSelection = () => {
+    navigation.push('Intro', { initialSelection: false })
```
klinzo, line 82: Same below in `navigateToAddCustomDiscipline`
```suggestion
    navigation.navigate('Intro', { initialSelection: false })
```

klinzo (changes requested):
Tested on android. Works perfectly, really nice work again!

Things noticed:
1. The touch target for the header menu seems a little bit small
2. `Beruf hinzufügen` is centered while `Bereich Hinzufügen` seems to be left aligned
3. The cards on the home screen are quite large such that only the first two are visible on my (not really small phone), maybe we could reduce some spacings here?

SteffiStuffel:
> 1. The touch target for the header menu seems a little bit small

Fix it.

> 2. `Beruf hinzufügen` is centered while `Bereich Hinzufügen` seems to be left aligned

Good catch, wasn't like in the designs. Fixed it.

> 3. The cards on the home screen are quite large such that only the first two are visible on my (not really small phone), maybe we could reduce some spacings here?

Removed the spacing a bit here, if we want it smaller, I think we should talk to Nikola first.

src/components/AddElement.tsx
```diff
@@ -0,0 +1,49 @@
...
+  margin: ${props => props.theme.spacings.sm} auto;
+  padding: ${props => props.theme.spacings.xs};
+`
+
+const AddCustomDisciplineText = styled(SubheadingPrimary)`
+  text-transform: uppercase;
+`
+
+const Explanation = styled(ContentSecondary)`
+  padding: 0 ${props => props.theme.spacings.xxl} 0 15%;
```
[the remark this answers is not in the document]
SteffiStuffel: Yes right. I was really racking my brain here, as we need a padding that has the width of the icon plus the padding. In css i would have used grid, but this is not supported in react native, also css-calc function is not supported. Therefore I decided this is the best solution, although I am also not really happy with it. If you have a better solution, let me know. Otherwise I think it is fine to keep it.

src/components/AddElement.tsx
```diff
@@ -0,0 +1,49 @@
...
+  onPress: () => void
+  label: string
+  explanation?: string
+}
+
+const AddElement = ({ onPress, label, explanation }: Props): ReactElement => (
+  <Root onPress={onPress}>
+    <FlexRow>
+      <Icon width={wp('8%')} height={wp('8%')} testID='add-icon' />
+      <AddCustomDisciplineText>{label}</AddCustomDisciplineText>
```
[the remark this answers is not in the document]
SteffiStuffel: Though it would make sense here, to check if the component shows the icon, but not really needed. Removed it and replaced it with label.

src/components/HorizontalLine.tsx
```diff
@@ -0,0 +1,13 @@
...
+const Line = styled.View`
+  width: 100%;
+  height: 1px;
+  margin: ${props => props.theme.spacings.xs} auto;
+  border: solid 1px ${props => props.theme.colors.disabled};
+`
+
+const HorizontalLine = (): ReactElement => <Line />
+
+export default HorizontalLine
```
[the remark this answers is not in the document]
SteffiStuffel: No, no need. Changed it.

src/components/ListItem.tsx
```diff
@@ -138,17 +146,19 @@ const ListItem = ({
...
   return (
     <Container
+      disabled={disabled || !onPress}
       onPressIn={onPressIn}
       onPressOut={onPressOut}
       onLongPress={() => setPressed(true)}
       pressed={pressed}
-      delayLongPress={200}>
+      delayLongPress={200}
+      testID='list-item'>
```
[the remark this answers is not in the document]
SteffiStuffel: hmh, I was orienting myself on the tests above. I added to the test, that if disabled, the onPress-function is not called, but kept the test for correct background-color, as I am not really sure, if we should remove it.

src/navigation/Navigator.tsx
```diff
@@ -118,13 +120,18 @@ const Navigator = (): JSX.Element | null => {
...
         initialRouteName={professions !== null ? 'Home' : 'Intro'}
-        screenOptions={TransitionPresets.SlideFromRightIOS}>
+        screenOptions={{ cardStyle: { backgroundColor: theme.colors.background } }}>
         <Stack.Screen options={{ headerShown: false }} name='Home' component={HomeScreen} />
-        <Stack.Screen options={{ headerShown: false }} name='Intro' component={IntroScreen} />
+        <Stack.Screen
+          options={{ headerShown: false }}
+          name='Intro'
+          component={IntroScreen}
+          initialParams={{ initialSelection: true }}
```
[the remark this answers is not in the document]
SteffiStuffel: Something like Scope- or SectorSelection?

src/routes/ProfessionSelectionScreen.tsx
```diff
@@ -50,26 +50,36 @@ const ProfessionSelectionScreen = ({ route, navigation }: ProfessionSelectionScr
     if (selectedProfessions?.some(profession => profession.id === selectedItem.id)) {
       await AsyncStorage.removeSelectedProfession(selectedItem).then(refreshSelectedProfessions)
     } else {
-      await AsyncStorage.pushSelectedProfession(selectedItem).then(refreshSelectedProfessions)
+      await AsyncStorage.pushSelectedProfession(selectedItem).then(() => {
+        if (!initialSelection) {
+          navigation.navigate('ManageDisciplines')
+        } else {
+          refreshSelectedProfessions()
+        }
```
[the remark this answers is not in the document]
SteffiStuffel: Yes please.

src/routes/home/HomeScreen.tsx
SteffiStuffel: Without the callback we get infinite calls. `Please report: Excessive number of pending callbacks: 501. Some pending callbacks that might have leaked by never being called from native code`

src/routes/add-custom-discipline/AddCustomDisciplineScreen.tsx
```diff
@@ -90,7 +90,7 @@ const AddCustomDiscipline = ({ navigation }: AddCustomDisciplineScreenProps): JS
     setLoading(true)
     loadGroupInfo(code)
       .then(async () => AsyncStorage.setCustomDisciplines([...customDisciplines, code]))
-      .then(() => navigation.navigate('Home'))
+      .then(() => navigation.goBack())
```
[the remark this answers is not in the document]
SteffiStuffel: yes please

src/routes/manage-disciplines/ManageDisciplinesScreen.tsx
SteffiStuffel: not sure, why I did this. Tried with FocusEffect and seems to work fine also.

src/components/ListItem.tsx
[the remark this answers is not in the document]
klinzo: Yeah then it's fine, lets keep it :)

src/routes/home/HomeScreen.tsx
klinzo: Perhaps since we have to wrap it in a new arrow function to be able to call both refreshes? Splitting it in two `useFocusEffect`s might help (did not actually test it so not sure). Anyway, fine as it is.

src/navigation/Navigator.tsx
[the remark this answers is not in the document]
klinzo: We already have a `DisciplineSelection`, right? TBH I don't have an overview which is used when anymore and about the different implementations. Maybe we would be able to merge some stuff? Of course not in the scope of this PR. I am fine `ScopeSelection` I guess, not really nice but okay.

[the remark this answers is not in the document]
SteffiStuffel: I already created a ticket for it [LUN-300](https://issues.tuerantuer.org/browse/LUN-300).

We have ScopeSelection and ProfessionSelection for selecting a profession that should be learnt. And we have DisciplineSeleciton for selection which discipline of a profession should be learnt.

src/components/AddElement.tsx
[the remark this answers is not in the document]
klinzo: I definitely don't have a better solution, css and styling is not really my sahneseite :D maybe @f1sh1918 has one though

src/navigation/Navigator.tsx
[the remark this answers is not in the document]
klinzo: Works for me, renaming can be done there than as well :+1:

src/components/AddElement.tsx
```diff
@@ -0,0 +1,44 @@
...
+  margin: ${props => props.theme.spacings.sm} 0;
+`
+
+const AddCustomDisciplineText = styled(SubheadingPrimary)`
+  text-transform: uppercase;
+  padding-left: ${props => props.theme.spacings.xs};
+`
+
+const Explanation = styled(ContentSecondary)`
+  padding: 0 ${props => props.theme.spacings.xxl} 0 11%;
```
f1sh1918, line 19: can't we use either xl(10%) or xxl(12%) ? even if the design differs we should use defined spacings

src/components/ListItem.tsx
```diff
@@ -138,17 +146,19 @@ const ListItem = ({
...
       {iconToRender}
       <FlexContainer>
         {titleToRender}
         <DescriptionContainer>
           {badgeLabel && <BadgeLabel pressed={pressed}>{badgeLabel}</BadgeLabel>}
-          <Description pressed={pressed}>{description}</Description>
+          {description && description.length > 0 && <Description pressed={pressed}>{description}</Description>}
```
f1sh1918, line 161:
```suggestion
           {description?.length > 0 && <Description pressed={pressed}>{description}</Description>}
```
you can use shorthand

src/components/ListItem.tsx
```diff
@@ -138,17 +146,19 @@ const ListItem = ({
...
       onLongPress={() => setPressed(true)}
       pressed={pressed}
-      delayLongPress={200}>
+      delayLongPress={200}
+      testID='list-item'>
       {iconToRender}
       <FlexContainer>
         {titleToRender}
         <DescriptionContainer>
           {badgeLabel && <BadgeLabel pressed={pressed}>{badgeLabel}</BadgeLabel>}
```
f1sh1918, line 160:
```suggestion
           {badgeLabel && badgeLabel.length > 0 && <BadgeLabel pressed={pressed}>{badgeLabel}</BadgeLabel>}
```
optional since you didn't change that line

src/components/AddElement.tsx
SteffiStuffel: this padding needs to be the same as the width of the icon plus the padding of the text, so these text is exactly beneath the other one. For better understanding have a look at the designs, which makes it clear.

src/components/ListItem.tsx
```diff
@@ -16,9 +15,14 @@ export const GenericListItemContainer = styled.Pressable`
...
   min-height: ${hp('12%')}px;
-  background-color: ${prop => (prop.pressed ? prop.theme.colors.primary : prop.theme.colors.backgroundAccent)};
+  background-color: ${prop => {
+    if (prop.disabled) {
+      return prop.theme.colors.disabled
+    }
+    return prop.pressed ? prop.theme.colors.primary : prop.theme.colors.backgroundAccent
+  }};
   border: 1px solid ${prop => (prop.pressed ? prop.theme.colors.primary : prop.theme.colors.disabled)};
   padding: ${props =>
```
f1sh1918, line 27: optional because not touched:
`padding: ${props => props.theme.spacings.sm};`
not sure why padding left & right are not equal

src/routes/manage-selections/ManageSelectionsScreen.tsx
```diff
@@ -0,0 +1,112 @@
...
+          <CustomDisciplineItem key={customDiscipline} refresh={refreshCustomDisciplines} apiKey={customDiscipline} />
+        ))}
+      </ServerResponseHandler>
+
+      <AddElement
+        onPress={navigateToAddCustomDiscipline}
+        label={labels.home.addCustomDiscipline}
+        explanation={labels.manageDisciplines.descriptionAddCustomDiscipline}
+      />
+      <Padding />
```
f1sh1918, line 107: can't find any difference without this view container. whats the purpose?

src/components/HeaderWithMenu.tsx
```diff
@@ -22,25 +26,44 @@ const SmileIconStyle = styled.Pressable`
   top: 20px;
 `
 
-const MenuIconContainer = styled.Pressable`
+const HeaderButtonsContainer = styled.View`
```
f1sh1918, line 29: you can just use flexbox instead of positioning absolute

```js
const HeaderButtonsContainer = styled.View`
  align-self: flex-end;
`
```

f1sh1918 (commented):
Works fine on iOS just found some minor styling issues/improvements

- Red circle of delete icon in `ManageSelectionScreen` is missing
- I think we could remove the disabled color from ListItem (check code comment)
- Design of `ManageSelectionScreen`does not fit with the implementation

src/components/AddElement.tsx
f1sh1918: alright keep it like that.

src/routes/manage-selections/ManageSelectionsScreen.tsx
SteffiStuffel: If you add 5 professions the last element is close to the bottom.

src/components/HeaderWithMenu.tsx
SteffiStuffel: nice

SteffiStuffel:
> Works fine on iOS just found some minor styling issues/improvements
> 
> * Red circle of delete icon in `ManageSelectionScreen` is missing
> * I think we could remove the disabled color from ListItem (check code comment)
> * Design of `ManageSelectionScreen`does not fit with the implementation

You can see the version, I implemented here, won't do the changes on the screen now, but fixed the red circle icon
https://app.zeplin.io/project/5fc7c98a435298126bcfd4ee/screen/62419215eaccb511191ee013/version/62419215eaccb511191ee014

Fixed disable color.

src/components/ListItem.tsx
SteffiStuffel: Because we use this list item also for the selection of a module and there we have this ">" icon, which should be more on the right. https://app.zeplin.io/project/5fc7c98a435298126bcfd4ee/screen/5fd7ad2c3cfec77f5c68f10c

src/components/ListItem.tsx
```diff
@@ -98,7 +98,9 @@ const ListItem = ({
   disabled = false
 }: ListItemProps): ReactElement => {
   const [pressInY, setPressInY] = useState<number | null>(null)
-  const [pressed, setPressed] = useState<boolean>(false)
+  const [pressed, setPressedState] = useState<boolean>(false)
+  const setPressed = (pressed: boolean): void => onPress && setPressedState(pressed)
```
klinzo, line 102: Please find another name for current `setPressed` and rename the `useState` method to `setPressed`. I think this is a convention we should stick to that `useState` value and setter are always the same. How about or similar:
```suggestion
  const [pressed, setPressed] = useState<boolean>(false)
  const updatePressed = (pressed: boolean): void => onPress && setPressedState(pressed)
```
I also think it should depend on `disabled` instead of on `onPress`. If the button is not disabled it should react to user presses even if there is no `onPress` method. At least that is what I'd expect.

klinzo (approved):

f1sh1918 (changes requested):
image

In `ManageSelectionsScreen`and `CustomDisciplineItem` different paddings for CloseIconRed Pressable Container are used

src/components/ListItem.tsx
SteffiStuffel: My thoughts about this were:
1) you don't have to think that setPress needs different handling and just call it as normal and side-case handling is done for you
2) disabling changes the color too, and we what three types of behaviour: a) disabled = grey and no action on click b) disabled = not-grey but no action on click c) not disabled = not-grey and action on click.

f1sh1918 (approved):

PR #267, opened 2022-03-29
LUN-233: Create component for unlock exercises

bahaaTuffaha:
This pull request belongs to an issue on our [LUN-233](https://issues.tuerantuer.org/browse/LUN-233)

PR #281, opened 2022-04-19
Lun 264 create locking lane for unlocking exercises

abdullahOde-117:

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "@sentry/react-native": "^3.2.13",
     "axios": "^0.24.0",
     "axios-cache-interceptor": "0.9.1",
+    "pods": "^2.0.0",
     "react": "17.0.2",
     "react-native": "0.66.4",
     "react-native-camera": "^4.2.1",
@@ -48,7 +49,7 @@
     "react-native-svg": "^12.1.1",
     "react-native-svg-transformer": "^0.14.3",
     "react-native-tts": "4.1.0",
-    "react-native-vector-icons": "^9.0.0",
+    "react-native-vector-icons": "^9.1.0",
     "string-similarity": "^4.0.4",
     "styled-components": "^5.3.3"
   },
--- a/src/routes/ExercisesScreens.tsx
+++ b/src/routes/ExercisesScreens.tsx
@@ -3,7 +3,7 @@ import { StackNavigationProp } from '@react-navigation/stack'
 import React,{useState} from 'react'
 import { FlatList, Alert } from 'react-native'
 import styled from 'styled-components/native'
-
+import AAsyncStorage from '@react-native-async-storage/async-storage';
 import ListItem from '../components/ListItem'
 import Title from '../components/Title'
 import Trophy from '../components/Trophy'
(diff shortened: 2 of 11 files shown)


package.json
```diff
@@ -29,6 +29,7 @@
     "@sentry/react-native": "^3.2.13",
     "axios": "^0.24.0",
     "axios-cache-interceptor": "0.9.1",
+    "pods": "^2.0.0",
```
SteffiStuffel, line 32: please undo

package.json
```diff
@@ -48,7 +49,7 @@
     "react-native-svg": "^12.1.1",
     "react-native-svg-transformer": "^0.14.3",
     "react-native-tts": "4.1.0",
-    "react-native-vector-icons": "^9.0.0",
+    "react-native-vector-icons": "^9.1.0",
```
SteffiStuffel, line 52: please undo

SteffiStuffel (changes requested):
Please remove these line-svgs. You should create Views which are only fex px wide and have a dynamic/appropriate height.

SteffiStuffel:
> Please remove these line-svgs. You should create Views which are only few px wide and have a dynamic/appropriate height. E.g. like here https://www.w3schools.com/howto/howto_css_vertical_line.asp

src/routes/ExercisesScreens.tsx
```diff
@@ -13,9 +13,15 @@ import { RoutesParams } from '../navigation/NavigationTypes'
 import { childrenDescription } from '../services/helpers'
 import { MIN_WORDS } from './choice-exercises/WordChoiceExerciseScreen'
 
+
```
f1sh1918, line 16: rm empty lines

src/routes/ExercisesScreens.tsx
```diff
@@ -13,9 +13,15 @@ import { RoutesParams } from '../navigation/NavigationTypes'
...
+
+
+
 const Root = styled.View`
   background-color: ${prop => prop.theme.colors.background};
   height: 100%;
+  display:flex;
+  flexDirection:row;
```
f1sh1918, line 24: in styled components you can use normal css like
`flex-direction: row`

src/routes/ExercisesScreens.tsx
```diff
@@ -38,6 +47,19 @@ const ExercisesScreen = ({ route, navigation }: ExercisesScreenProps): JSX.Eleme
       })
     }
   }
+  
+useFocusEffect(
+  React.useCallback(()=>{
+    AAsyncStorage.getItem('level').then((value)=>{
```
f1sh1918, line 53: move this function to `AsyncStorage.ts` use async fkt with Promise

package.json
SteffiStuffel: If we do it in a separate ticket, Martin will know and can test, when we do it here, we now have to do research, if it breaks something and hope everything is fine. If there is no need, I would not do it here. @klinzo any opinions on this?

src/routes/ExercisesScreens.tsx
SteffiStuffel: No need for you @abdullahOde-117 to write AsyncStorage-Function just use a dummy variable and add a comment, that it will be done in LUN-131. See how bahaa did it here (line 38): https://github.com/digitalfabrik/lunes-app/pull/267/files/#diff-e32395b665dac2ea67c45f7e2de4b1f284a10026c8eaa71b5da1ea046567614f

package.json
klinzo: Agreeing with @SteffiStuffel , upgrades should only be done if there is a reason for it or if there is a task for it.

f1sh1918:
If you need some help by explaining things like css and how to create such a progress bar or async functions just contact me via mattermost

PR #282, opened 2022-04-19
LUN-265 Add menu bottom bar

bahaaTuffaha:
This pull request belongs to an issue on our [LUN-265](https://issues.tuerantuer.org/browse/LUN-265).

Notes:
I couldn’t hide the bottom navigator at certain screens , I have tried the following:
1- useNavigationState or useNavigation to get the current screen name and hide it accordingly but it always return the stack navigator screen names.
2- by passing a prop to the exercise screen ( a bit confused from typescript).
3 - maybe using global state like React Context (I think its a terrible idea?).
Any suggestions ??


SteffiStuffel:
@steffenkleinle I will close this PR, but not delete the branch.

PR #284, opened 2022-04-21
LUN-310: Save progess of exercises

SteffiStuffel:

For testing you need to add logging somewhere, to check if everything is saved correctly, as the saved progress is not yet displayed anywhere.


SteffiStuffel:
> This pull request belongs to an issue on our [bugtracker](https://issues.tuerantuer.org/secure/RapidBoard.jspa?rapidView=27&view=detail). You can find it there by looking for an issue with the key which is mentioned in the title of this pull request. It starts with the keyword _LUN_.

For testing you need to add logging somewhere, to check if everything is saved correctly, as the saved progress is not yet displayed anywhere.

src/services/helpers.ts
```diff
@@ -56,3 +58,12 @@ export const shuffleArray = <T>(array: T[]): T[] => {
   }
   return shuffled
 }
+
+export const saveExerciseProgress = async (
```
klinzo, line 62: I think it would be fine to just move this to the async storage as well.

src/services/__tests__/AsyncStorage.spec.ts
```diff
@@ -40,4 +41,35 @@ describe('AsyncStorage', () => {
     await AsyncStorage.pushSelectedProfession(mockDisciplines[1])
     await expect(AsyncStorage.getSelectedProfessions()).resolves.toHaveLength(2)
   })
+
+  describe('ExerciseProgress', () => {
+    const firstExercise = { disciplineId: 1, exerciseKey: ExerciseKeys.wordChoiceExercise, score: 0.5 }
+    const secondExercise = { disciplineId: 1, exerciseKey: ExerciseKeys.writeExercise, score: 0.5 }
+    const secondExerciseBetter = { disciplineId: 1, exerciseKey: ExerciseKeys.writeExercise, score: 0.75 }
+
+    beforeEach(() => AsyncStorage.clearExerciseProgress())
```
klinzo, line 50: Is the async storage not cleaned automatically between each test yet? I think that should definitely happen. Perhaps even a jest.clearAllMocks should be enough.

src/services/AsyncStorage.ts
```diff
@@ -56,6 +59,34 @@ export const removeCustomDiscipline = async (customDiscipline: string): Promise<
...
 }
 
+export const getExerciseProgress = async (): Promise<Array<ExerciseProgress>> => {
+  const progress = await AsyncStorage.getItem(exerciseProgress)
+  return progress ? JSON.parse(progress) : []
+}
+
+export const setExerciseProgress = async (progress: ExerciseProgress): Promise<void> => {
+  const savedProgress = await getExerciseProgress()
+  const index = savedProgress.findIndex(
```
klinzo, line 69: Can't we save a map `id -> ExerciseProgress` instead of an `ExerciseProgress[]` array? I think that would make mutation a lot easier. This logic seems to be quite complicated.

src/navigation/NavigationTypes.ts
```diff
@@ -10,6 +10,7 @@ export interface DocumentResult {
 }
 
 interface ExerciseParams {
+  disciplineId: number
```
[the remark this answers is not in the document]
SteffiStuffel: Yes you are right, I was confused about this ExerciseParams types, but now I get it.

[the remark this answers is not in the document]
SteffiStuffel: As discussed in the call, we will keep disciplineId and check for it before saving progress.

src/services/__tests__/AsyncStorage.spec.ts
SteffiStuffel: Created a ticket for this

src/services/AsyncStorage.ts
SteffiStuffel: Yes we can use a Map, but it is really ugly to save / read Maps from JSON. This would be the working code for Maps:
`
const replacer = (key: any, value: any) => {
  if (value instanceof Map) {
    return {
      dataType: 'Map',
      value: Array.from(value.entries()) // or with spread: value: [...value]
    }
  }
  return value
}

const reviver = (key: any, value: any) => {
  if (typeof value === 'object' && value !== null) {
    if (value.dataType === 'Map') {
      return new Map(value.value)
    }
  }
  return value
}

export const getExerciseProgress = async (): Promise<Progress> => {
  const progress = await AsyncStorage.getItem(progressKey)
  return progress ? JSON.parse(progress, reviver) : new Map()
}

const setExerciseProgress = async (disciplineId: number, exerciseKey: ExerciseKey, score: number): Promise<void> => {
  const savedProgress = await getExerciseProgress()
  const disciplineProgress = savedProgress.get(disciplineId)
  if (disciplineProgress) {
    const exerciseProgress = disciplineProgress.get(exerciseKey)
    if (!exerciseProgress || exerciseProgress <= score) {
      disciplineProgress.set(exerciseKey, score)
      savedProgress.set(disciplineId, disciplineProgress)
    }
  } else {
    const exerciseProgress: ExerciseProgress = new Map()
    exerciseProgress.set(exerciseKey, score)
    savedProgress.set(disciplineId, exerciseProgress)
  }
  await AsyncStorage.setItem(progressKey, JSON.stringify(savedProgress, replacer))
`

SteffiStuffel:
> i just tested functionality and printed out the progress from `AsyncStorage` `getProgress` i did two exercises like you can see in the screenshot and i see that the `exerciseKey: 1` -> "Word Choice" But how do i know which area it is f.e. "Bau und Gerüst" or "Kleben und Schreiben" since the `discliplineId` seems to be "Maler/in und Lackier/in"
> 

Fixed it. Good catch, wrong id was saved.

src/services/__tests__/helpers.spec.tsx
```diff
@@ -1,8 +1,7 @@
 import { SIMPLE_RESULTS } from '../../constants/data'
 import { DocumentResult } from '../../navigation/NavigationTypes'
 import DocumentBuilder from '../../testing/DocumentBuilder'
-import AsyncStorage from '../AsyncStorage'
-import { saveExerciseProgress } from '../helpers'
+import AsyncStorage, { saveExerciseProgress } from '../AsyncStorage'
 
 describe('helpers', () => {
   it('should calculate and save exercise progress correctly', async () => {
```
klinzo, line 7: This test is now in the wrong place I think. Fixed it in #289

klinzo (approved):
Created a proposal for simpler progress saving here: #289
Feel free to reject if it does not suit your needs but I think it makes things quite a bit easier. Sorry for being unclear what I meant with saving a map :see_no_evil:

src/services/AsyncStorage.ts
```diff
@@ -59,32 +60,34 @@ export const removeCustomDiscipline = async (customDiscipline: string): Promise<
...
+export const saveExerciseProgress = async (
+  disciplineId: number,
+  exerciseKey: ExerciseKey,
+  documentsWithResults: DocumentResult[]
+): Promise<void> => {
+  const score = documentsWithResults.filter(doc => doc.result === 'correct').length / documentsWithResults.length
+  await setExerciseProgress(disciplineId, exerciseKey, score)
 }
 
 export const getProgress = async (profession: Discipline): Promise<number> => {
```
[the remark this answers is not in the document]
SteffiStuffel: No this is the progress of a profession which needs to be calculated from progress of all - also nested - training sets. Tickets is open for this, but waiting for cms.

klinzo (changes requested):
Not sure if this is correct:
In the write exercise if I repeat on only the wrong answers, the score is (probably) wrong. E.g. I have 5 words and have 2 correct, the score is `0.4` (2 of 5). If I restart the exercise and have 2 correct, the score is `0.66` (2 of 3), even though it should probably be `0.8` (4 of 5) or should stay the same. Not sure what the correct behavior would be here but I think the current is not correct.

I am also not sure if it makes sense that scores can only increase and never decrease, but thats more of an UI/UX/project idea.

SteffiStuffel:
> Not sure if this is correct: In the write exercise if I repeat on only the wrong answers, the score is (probably) wrong. E.g. I have 5 words and have 2 correct, the score is `0.4` (2 of 5). If I restart the exercise and have 2 correct, the score is `0.66` (2 of 3), even though it should probably be `0.8` (4 of 5) or should stay the same. Not sure what the correct behavior would be here but I think the current is not correct.
> 

You are right, we should think about correct handling here. I will do this a separate ticket.

SteffiStuffel:
> Not sure if this is correct: In the write exercise if I repeat on only the wrong answers, the score is (probably) wrong. E.g. I have 5 words and have 2 correct, the score is `0.4` (2 of 5). If I restart the exercise and have 2 correct, the score is `0.66` (2 of 3), even though it should probably be `0.8` (4 of 5) or should stay the same. Not sure what the correct behavior would be here but I think the current is not correct.
> 
> I am also not sure if it makes sense that scores can only increase and never decrease, but thats more of an UI/UX/project idea.

Oke just thought about it again, and we won't need it, as the whole retry wrong words stuff is going to be removed.

klinzo (approved):

klinzo:
> Oke just thought about it again, and we won't need it, as the whole retry wrong words stuff is going to be removed.

Any specific reason for this? Is there already an issue?

SteffiStuffel:
> > Oke just thought about it again, and we won't need it, as the whole retry wrong words stuff is going to be removed.
> 
> Any specific reason for this? Is there already an issue?

Will be done with redesigning of the result screens LUN-231. Reason is better user flow and the repetition of wrong words, which lead to 100% correct answers most of the time.

PR #287, opened 2022-04-26
LUN-318: Exercises screen shows wrong list title

f1sh1918:

Choose a discipline and check if the exercise screen uses correct header for navigation and list

### DIFF
--- a/src/navigation/NavigationTypes.ts
+++ b/src/navigation/NavigationTypes.ts
@@ -16,8 +16,10 @@ interface ExerciseParams {
   closeExerciseAction: CommonNavigationAction
 }
 
-export interface ExercisesParams {
+export interface ExercisesParams extends Omit<ExerciseParams, 'documents' | 'closeExerciseAction'> {
   discipline: Discipline
+  documents: Document[] | null
+  closeExerciseAction?: CommonNavigationAction
 }
 
 type ResultParams = ExerciseParams & {
--- a/src/navigation/NavigationTypes.ts
+++ b/src/navigation/NavigationTypes.ts
@@ -9,16 +9,14 @@ export interface DocumentResult {
   numberOfTries: number
 }
 
-export interface ExerciseParams {
+interface ExerciseParams {
   disciplineTitle: string
   documents: Document[]
   closeExerciseAction: CommonNavigationAction
 }
 
 export interface ExercisesParams {
-  documents: null
   discipline: Discipline
-  disciplineTitle: string
 }
 
 type ResultParams = ExerciseParams & {
--- a/src/navigation/NavigationTypes.ts
+++ b/src/navigation/NavigationTypes.ts
@@ -9,19 +9,17 @@ export interface DocumentResult {
   numberOfTries: number
 }
 
-interface ExerciseParams {
+export interface ExerciseParams {
   disciplineTitle: string
   documents: Document[]
   closeExerciseAction: CommonNavigationAction
 }
 
-export type ExercisesParams =
-  | {
-      documents: null
-      discipline: Discipline
-      disciplineTitle: string
-    }
-  | ExerciseParams
+export interface ExercisesParams {
+  documents: null
+  discipline: Discipline
+  disciplineTitle: string
+}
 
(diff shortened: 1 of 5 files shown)


src/routes/DisciplineSelectionScreen.tsx
```diff
@@ -32,7 +32,11 @@ const DisciplineSelectionScreen = ({ route, navigation }: DisciplineSelectionScr
 
   const handleNavigation = (selectedItem: Discipline): void => {
     if (selectedItem.isLeaf) {
-      navigation.navigate('Exercises', { discipline: selectedItem, disciplineTitle: discipline.title, documents: null })
+      navigation.navigate('Exercises', {
+        discipline: selectedItem,
+        disciplineTitle: selectedItem.title,
```
klinzo, line 37: :+1:

src/navigation/NavigationTypes.ts
```diff
@@ -9,19 +9,17 @@ export interface DocumentResult {
...
 }
 
-export type ExercisesParams =
-  | {
-      documents: null
-      discipline: Discipline
-      disciplineTitle: string
-    }
-  | ExerciseParams
+export interface ExercisesParams {
```
klinzo, line 18: If you change this type please change the logic in `ExercisesScreen` as well with the suggestion below. Also the types should be cleaned up then: Remove `documents: null` and `disciplineTitle`. Otherwise leave it as it was and I will clean it up later.

src/routes/ExercisesScreen.tsx
```diff
@@ -29,7 +29,7 @@ const ExercisesScreen = ({ route, navigation }: ExercisesScreenProps): JSX.Eleme
   const { params } = route
   const { disciplineTitle } = params
 
-  const load = useCallback(async (params: ExercisesParams) => {
+  const load = useCallback(async (params: ExercisesParams | ExerciseParams) => {
```
klinzo, line 32: Line 32 to 38 should be replaced with if you keep the change of the navigation types:
``` ts
  const { data: documents, error, loading, refresh } = useLoadDocuments(discipline)
```

klinzo (changes requested):
Tested

src/routes/ExercisesScreen.tsx
klinzo: Done in https://github.com/digitalfabrik/lunes-app/pull/287/commits/13c7cf9bbed09646973c9eac373532f95d6987e2

src/navigation/NavigationTypes.ts
klinzo: Done in https://github.com/digitalfabrik/lunes-app/pull/287/commits/13c7cf9bbed09646973c9eac373532f95d6987e2

klinzo (approved):

src/navigation/NavigationTypes.ts
SteffiStuffel: It think it should still be possible to navigate to ExerciseScreen with passing documents instead of discipline, but I can add this later if wanted.

SteffiStuffel (approved):
Tested

src/navigation/NavigationTypes.ts
```diff
@@ -16,14 +16,11 @@ interface ExerciseParams {
...
 
-export type ExercisesParams =
-  | {
-      documents: null
-      discipline: Discipline
-      disciplineTitle: string
-      disciplineId: number
-    }
-  | ExerciseParams
+export interface ExercisesParams extends Omit<ExerciseParams, 'documents' | 'closeExerciseAction'> {
```
f1sh1918, line 19: @SteffiStuffel , @klinzo 
i'm not sure if this extend and omit is a good way or its better to create a separate `ExercisesParams` interface since there are some differences between both interface

SteffiStuffel: I don't have a strong opinion with this and i am fine with both options

f1sh1918: alright then i keep it as it is

PR #288, opened 2022-04-27
LUN-278: Start next exercise from home screen

SteffiStuffel:


src/hooks/useLoadDocuments.ts
```diff
@@ -29,11 +29,16 @@ const formatServerResponse = (documents: DocumentFromServer[]): Document[] =>
 
 export const loadDocuments = async (discipline: Discipline): Promise<Document[]> => {
   const url = ENDPOINTS.documents.replace(':id', `${discipline.id}`)
-
   const response = await getFromEndpoint<DocumentFromServer[]>(url, discipline.apiKey)
   return formatServerResponse(response)
 }
 
+export const loadDocumentsById = async (disciplineId: number): Promise<Document[]> => {
```
klinzo, line 36: This is quite redundant and duplicated. Should be better to refactor `loadDocuments` to only take `disciplineId` and optionally an `apiKey`. Then you can either change `useLoadDocuments` such that it also just takes an id + apiKey or just pass discipline.id and discipline.apiKey as params

src/services/__tests__/helpersNextEcercise.spec.tsx
```diff
@@ -0,0 +1,105 @@
+import { loadDisciplines } from '../../hooks/useLoadDisciplines'
```
klinzo, line 1: File should be called `helpers.spec.ts` and another `describe` block should be used instead.

src/routes/home/HomeScreen.tsx
```diff
@@ -56,6 +60,21 @@ const HomeScreen = ({ navigation }: HomeScreenProps): JSX.Element => {
     })
   }
 
+  const navigateToNextExercise = async (item: Discipline): Promise<void> => {
```
klinzo, line 63: This is currently not async since you are not awaiting anything

src/routes/home/HomeScreen.tsx
```diff
@@ -56,6 +60,21 @@ const HomeScreen = ({ navigation }: HomeScreenProps): JSX.Element => {
...
+    getNextExercise(item)
+      .then(async nextExerciseData => {
```
klinzo, line 65: I think it is not good to mix async/await inside then/catch blocks. Instead, you should either chain it correctly or only use async/await style syntax.

src/routes/home/HomeScreen.tsx
```diff
@@ -56,6 +60,21 @@ const HomeScreen = ({ navigation }: HomeScreenProps): JSX.Element => {
...
+        const [disciplineId, exerciseKey] = nextExerciseData
+        const documents = await loadDocumentsById(disciplineId)
+        navigation.navigate(EXERCISES[exerciseKey].nextScreen, {
```
klinzo, line 68: If you navigate to `EXERCISES[exerciseKey].nextScreen`, `getNextExercise` does not really return the next exercise, does it?

src/routes/home/HomeScreen.tsx
```diff
@@ -56,6 +60,21 @@ const HomeScreen = ({ navigation }: HomeScreenProps): JSX.Element => {
     })
   }
 
+  const navigateToNextExercise = async (item: Discipline): Promise<void> => {
+    getNextExercise(item)
+      .then(async nextExerciseData => {
+        const [disciplineId, exerciseKey] = nextExerciseData
```
klinzo, line 66: Better to use an object as return type, otherwise it is just `Promise<number, number>` and one does not know which numbers it returns.

src/routes/home/HomeScreen.tsx
```diff
@@ -56,6 +60,21 @@ const HomeScreen = ({ navigation }: HomeScreenProps): JSX.Element => {
...
+        const documents = await loadDocumentsById(disciplineId)
```
klinzo, line 67: This probably makes clicks/navigation seem really unresponsive. Maybe we need to move document loading into the exercises again/at least support both. This is really a pain in the a** :cry:

src/services/helpers.ts
```diff
@@ -56,3 +58,29 @@ export const shuffleArray = <T>(array: T[]): T[] => {
   }
   return shuffled
 }
+
+export const getNextExercise = async (profession: Discipline): Promise<[number, number]> => {
```
klinzo, line 62: Will review this once updated to base branch

klinzo (changes requested):

src/routes/home/HomeScreen.tsx
klinzo: Other ideas would of course be welcome...

klinzo: If you want I can implement a proposal

src/hooks/useLoadDocuments.ts
SteffiStuffel: You are right, i tried this, but I couldn't figure out how to pass multiple params to useLoadAsync.

`
const useLoadDocuments = (disciplineId: number, apiKey?: string): Return<Document[]> => useLoadAsync(loadDocuments, {discipline.id, discipline.apiKey})
`

I get `TS2345: Argument of type '(disciplineId: number, apiKey: string) => Promise ' is not assignable to parameter of type '(params: number) => Promise '`

src/services/__tests__/helpersNextEcercise.spec.tsx
SteffiStuffel: Also I tried this, but the test in helpers.spec.tsx needs unmocked/real version of of AsyncStorage, wheres the tests for NextExercise (in this file) need a mocked version of it. As far as I know (and tired out) it is not possible to mock stuff only for some tests in a file.

src/routes/home/HomeScreen.tsx
SteffiStuffel: getNextExercise returns the exerciseKey of the nextExercise? I think I do not get your question?

src/routes/home/HomeScreen.tsx
SteffiStuffel: We also could request documents when rendering home screen :thinking: 
But feel free to implement a proposal

src/hooks/useLoadDocuments.ts
klinzo: You just use an object as well, i.e.:
``` ts
const const useLoadDocuments = ({ id, apiKey }: { id: number, apiKey?: string }): Return<Document[]> => useLoadAsync(loadDocuments, { id, apiKey })
```

src/services/__tests__/helpersNextEcercise.spec.tsx
klinzo: It is somehow possible (I think by using doMock and restoreMock or similar), but I agree, it is a real pain. However, the test that is currently in `helpers.spec.ts` in your branch was moved in #284 to the async storage test so you can move it without problems. Also the "real" async storage in tests should already be mocked afaik.

src/routes/home/HomeScreen.tsx
klinzo: But why do we need `.nextScreen` then? Isn't it the next exercise of the exercise with exercise key?

SteffiStuffel: Ah, oke, I see where your confusion comes from. Naming here is really shitty, as "nextScreen" is just the screen of this exercise. I will refactor this in a separate PR.

src/services/__tests__/helpersNextEcercise.spec.tsx
SteffiStuffel: Tried doMock, but couldn't get it working, but as you mentioned no need any more as it was moved.

src/services/helpers.ts
```diff
@@ -56,3 +58,31 @@ export const shuffleArray = <T>(array: T[]): T[] => {
...
+  const disciplines = await loadDisciplines(profession) // TODO LUN-316 leaf disciplines must be loaded, also if nested
+  if (disciplines.length <= 0) {
+    throw new Error(`No Disciplines for id ${profession.id}`)
+  }
+  const progress = await AsyncStorage.getExerciseProgress()
+  const doneExercisesOfLeafDiscipline = (disciplineId: number): number => {
+    const progressOfDiscipline = progress[disciplineId]
+    return progressOfDiscipline
+      ? Object.keys(progressOfDiscipline).filter(item => progressOfDiscipline[item] !== undefined).length
+      : 0
```
klinzo, line 74: Isnt this just the same as `Object.keys.length` or `Object.values.length`?

src/services/helpers.ts
```diff
@@ -56,3 +58,31 @@ export const shuffleArray = <T>(array: T[]): T[] => {
...
+  }
+  const result = disciplines.find(discipline => doneExercisesOfLeafDiscipline(discipline.id) < exercisesWithProgress)
```
klinzo, line 76: Maybe you could rename `result` to what it actually contains (something like `unfinishedDiscipline` probably?

src/services/helpers.ts
```diff
@@ -56,3 +58,31 @@ export const shuffleArray = <T>(array: T[]): T[] => {
...
+  if (!result) {
+    return { disciplineId: disciplines[0].id, exerciseKey: exercisesWithoutProgress } // TODO LUN-319 show success that every exercise is done
+  }
+  const disciplineProgress = progress[result.id]
+  if (!disciplineProgress) {
+    return { disciplineId: result.id, exerciseKey: exercisesWithoutProgress }
```
klinzo, line 82: I don't really understand why you return `exercisesWithoutProgress` as `exerciseKey`. Isn't exerciseWithoutProgress just the number of exercises where no progress is saved (i.e. `1` because it is only the VocabularyList)?

src/services/helpers.ts
```diff
@@ -56,3 +58,31 @@ export const shuffleArray = <T>(array: T[]): T[] => {
...
+  }
+  const nextExerciseKey = EXERCISES.slice(exercisesWithoutProgress).find(
+    exercise => disciplineProgress[exercise.key] === undefined
+  )
+  return { disciplineId: result.id, exerciseKey: nextExerciseKey ? nextExerciseKey.key : exercisesWithoutProgress }
```
klinzo, line 87: Minor: You could do
```suggestion
  return { disciplineId: result.id, exerciseKey: nextExerciseKey?.key ?? exercisesWithoutProgress }
```

klinzo (approved):
Not tested

src/services/helpers.ts
SteffiStuffel: No, as the data structure allows a object like this:

```
200: {
   1: 0.2,
   2: undefined
}

And for this we need length = 1, therefore we need to filter the undefined values.

File: src/services/helpers.ts (hunk 10 above)
Comment (SteffiStuffel) [reply src/services/helpers.ts:76]: Renamed to firstUnfinishedDiscipline

File: src/services/helpers.ts (hunk 11 above)
Comment (SteffiStuffel) [reply src/services/helpers.ts:82]: Yes you are right. But the if-condition `!disciplineProgress` gets true, if there is not progress for this discipline, therefore we want to start the next exercise, that should be done, which is the SingleChoiceExercise currently, but in general it is the next exercise after skipping the exercises that have no progress.

File: src/services/helpers.ts [hunk 13]
```diff
@@ -56,3 +58,36 @@ export const shuffleArray = <T>(array: T[]): T[] => {
   }
   return shuffled
 }
+
+export const getNextExercise = async (
```
f1sh1918, line 62: i think due to the complexity of this fkt, it would be nice to get a short comment

src/routes/home/components/DisciplineCard.tsx
```diff
@@ -37,16 +40,35 @@ interface PropsType {
...
+          setDocuments(data)
+        })
+        .catch(reportError)
+    }
+  }, [nextExercise])
 
   const navigate = () => {
-    navigateToNextExercise(discipline)
+    if (documents !== null && nextExercise !== null) {
+      navigateToNextExercise(nextExercise.disciplineId, nextExercise.exerciseKey, '', documents) // TODO set discipline title correct LUN-320
```
f1sh1918, line 70: why do we need to pass the `disciplineTitle` here?
I see that according to design we need more information about the next exercise for extending DisciplineCard but i think that comes later with a separate task
https://app.zeplin.io/project/5fc7c98a435298126bcfd4ee/screen/624ac2cc6aacf3112ea90218

f1sh1918 (approved):
Really great work :) 
Tested on ios

src/routes/home/components/DisciplineCard.tsx
SteffiStuffel: Resolved in https://github.com/digitalfabrik/lunes-app/pull/290

PR #293, opened 2022-05-04
LUN-314: Exercise height should be 100 

f1sh1918:

Tested on ios and android


SteffiStuffel (approved):
Tested on ios browserstack. Can you please add a note for Martin in the ticket, that this needs to be tested on devices with different notches etc?

klinzo (approved):
Not tested

PR #294, opened 2022-05-05
LUN-301: Add loading and error handling

SteffiStuffel:

### DIFF
--- a/src/routes/manage-selections/components/ProfessionItem.tsx
+++ b/src/routes/manage-selections/components/ProfessionItem.tsx
@@ -1,0 +1,16 @@
+import React, { ReactElement } from 'react'
+
+import { useLoadDiscipline } from '../../../hooks/useLoadDiscipline'
+import SelectionItem from './SelectionItem'
+
+interface PropsType {
+  id: number
+  deleteItem: () => void
+}
+
+const ProfessionItem = ({ id, deleteItem }: PropsType): ReactElement => {
+  const discipline = useLoadDiscipline(id)
+  return <SelectionItem discipline={discipline} deleteItem={deleteItem} />
+}
+
+export default ProfessionItem
--- a/src/components/ErrorMessage.tsx
+++ b/src/components/ErrorMessage.tsx
@@ -3,13 +3,14 @@ import { widthPercentageToDP as wp } from 'react-native-responsive-screen'
 import styled from 'styled-components/native'
 
 import { BUTTONS_THEME } from '../constants/data'
+import { NetworkError } from '../constants/endpoints'
 import labels from '../constants/labels.json'
 import Button from './Button'
 
 const Container = styled.View`
-  width: ${wp('90%')}px;
+  width: ${wp('80%')}px;
   margin: auto;
-  margin-top: ${props => props.theme.spacings.xl};
+  margin-top: ${props => props.theme.spacings.md};
   text-align: center;
   display: flex;
   align-items: center;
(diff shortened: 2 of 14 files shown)


src/routes/manage-selections/components/ProfessionItem.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { ReactElement } from 'react'
```
SteffiStuffel, line 1: I created this ProfessionItem, which does only load a profession and then displays the Selection item. Same for the CustomDisciplineItem. 

We need this, because of non-conditional calling of hooks, but I am not sure, if this is the best solution. Other solution would be to create a hook, which load either a discipline or customDiscipline by evaluating if passed parameter is a string(=apikey) or number(=discipline).

Any arguments pro/con the solutions or even other solutions?

SteffiStuffel: @klinzo

klinzo:
Will test once the decision on conditional rendering/hook logic is made.

src/routes/manage-selections/components/ProfessionItem.tsx
SteffiStuffel: I did not like the idea of the hook, which checks for types, but you are right, just having a good Params-type and it is the better solution. 

I will not do this in this PR, as it is quite a lot of refactoring and has nothing to do with the real task of LUN-301. Created a new ticket for it (LUN-339) and will start working on it soon.

klinzo (commented):
Things I noticed:
* If you have e.g. no internet connection, each card shows the same error and you have to click on `ERNEUT VERSUCHEN` separately for each card. Could maybe be improved in a separate issue.
* If you go to `Bereiche verwalten`, the screen jumps a little bit because the items don't occupy space until loaded

klinzo (approved):
Tested on android

SteffiStuffel:
> Things I noticed:
> 
> * If you go to `Bereiche verwalten`, the screen jumps a little bit because the items don't occupy space until loaded
> 

It should show a correct loading box with correct height for loading information form the cms, but we will have jumping after loading id's from async storage, as we cannot know before loading ids from storage how many boxes we need. Is that what you mean or is the loading box for loading data from cms not working?

SteffiStuffel:
> Things I noticed:
> 
> * If you go to `Bereiche verwalten`, the screen jumps a little bit because the items don't occupy space until loaded
> 

It should show a correct loading box with correct height for loading information form the cms, but we will have jumping after loading id's from async storage, as we cannot know before loading ids from storage how many boxes we need. Is that what you mean or is the loading box for loading data from cms not working?

klinzo:
> > Things I noticed:
> > 
> > * If you go to `Bereiche verwalten`, the screen jumps a little bit because the items don't occupy space until loaded
> 
> It should show a correct loading box with correct height for loading information form the cms, but we will have jumping after loading id's from async storage, as we cannot know before loading ids from storage how many boxes we need. Is that what you mean or is the loading box for loading data from cms not working?

We could just not show anything until everything is loaded to avoid jumping. I think it is the `Individuelle Bereiche` that is pushed down by the professions once loaded.

SteffiStuffel:
> > > Things I noticed:
> > > 
> > > * If you go to `Bereiche verwalten`, the screen jumps a little bit because the items don't occupy space until loaded
> > 
> > 
> > It should show a correct loading box with correct height for loading information form the cms, but we will have jumping after loading id's from async storage, as we cannot know before loading ids from storage how many boxes we need. Is that what you mean or is the loading box for loading data from cms not working?
> 
> We could just not show anything until everything is loaded to avoid jumping. I think it is the `Individuelle Bereiche` that is pushed down by the professions once loaded.

But before getting the information from Async storage if we need to load zero, one or five professions, we cannot know how much space we need to reserve for the profession. At this point we can not avoid jumping unless we block rendering at all, until we have this data, which is in general not recommended.

PR #295, opened 2022-05-09
LUN-329: Show correct header for scope selection

SteffiStuffel:

Needed to change from `navigate` to `push` for ScopeSelectionScreen and ProfessionSelectionScreen, as `navigate` would use the first and second screen in the stack form initial view, which will break our stack-logic. Other solution would be to clear the stack when navigation to the HomeScreen after initial selection, but as it works fine to navigate to the home screen and navigate back to initial selection (with hardware buttons), i think this way it is the better option.

### DIFF
--- a/src/routes/scope-selection/ScopeSelectionScreen.tsx
+++ b/src/routes/scope-selection/ScopeSelectionScreen.tsx
@@ -51,7 +51,7 @@ const ScopeSelectionScreen = ({ navigation, route }: IntroScreenProps): JSX.Elem
   })
 
   const navigateToDiscipline = (item: Discipline): void => {
-    navigation.push('ProfessionSelection', {
+    navigation.navigate('ProfessionSelection', {
       discipline: item,
       initialSelection
     })
@@ -61,7 +61,10 @@ const ScopeSelectionScreen = ({ navigation, route }: IntroScreenProps): JSX.Elem
     if (selectedProfessions === null) {
       await AsyncStorage.setSelectedProfessions([])
     }
-    navigation.navigate('Home')
+    navigation.reset({
+      index: 0,
+      routes: [{ name: 'Home' }]
+    })
   }
 
   const disciplineItems = disciplines?.map(item => (
--- a/src/routes/scope-selection/ScopeSelectionScreen.tsx
+++ b/src/routes/scope-selection/ScopeSelectionScreen.tsx
@@ -1,6 +1,6 @@
 import { RouteProp, useFocusEffect } from '@react-navigation/native'
 import { StackNavigationProp } from '@react-navigation/stack'
-import React from 'react'
+import React, { useLayoutEffect } from 'react'
 import { ScrollView } from 'react-native'
 import styled from 'styled-components/native'
 
@@ -46,8 +46,12 @@ const ScopeSelectionScreen = ({ navigation, route }: IntroScreenProps): JSX.Elem
 
   useFocusEffect(refreshSelectedProfessions)
 
+  useLayoutEffect(() => {
+    navigation.setOptions({ headerShown: !initialSelection })
+  })
+
   const navigateToDiscipline = (item: Discipline): void => {
-    navigation.navigate('ProfessionSelection', {
+    navigation.push('ProfessionSelection', {
       discipline: item,
       initialSelection
     })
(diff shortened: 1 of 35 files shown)


SteffiStuffel:
@klinzo Do you prefer reseting the history or using push?

src/routes/scope-selection/ScopeSelectionScreen.tsx
```diff
@@ -46,8 +46,12 @@ const ScopeSelectionScreen = ({ navigation, route }: IntroScreenProps): JSX.Elem
 
   useFocusEffect(refreshSelectedProfessions)
 
+  useLayoutEffect(() => {
+    navigation.setOptions({ headerShown: !initialSelection })
+  })
```
klinzo, line 51: Shouldn't we use callbacks for `useLayoutEffect`? Or does this just not matter since nothing happens if the options are already set?

klinzo (approved):
Tested on android

klinzo:
> @klinzo Do you prefer reseting the history or using push?

I definitely prefer using push over resetting the history, nothing wrong with push, it's a public api of react-navigation (at least as long as there is a reason to use it over navigate). Still not sure I really understood why we need it though.

src/routes/scope-selection/ScopeSelectionScreen.tsx
f1sh1918: this only gets triggered once in any case, no? so i think its not needed to useCallback. To be honest i didn't use this hook so far ;)

f1sh1918 (approved):
Tested on ios

src/routes/scope-selection/ScopeSelectionScreen.tsx
SteffiStuffel: React Navigation Docs do it without callback, so I think it is fine to use it without.

SteffiStuffel:
> > @klinzo Do you prefer reseting the history or using push?
> 
> I definitely prefer using push over resetting the history, nothing wrong with push, it's a public api of react-navigation (at least as long as there is a reason to use it over navigate). Still not sure I really understood why we need it though.

When starting the app for the first time select a discipline, confirm and then got to "settings" and want to add a new discipline
1) ScopeSelection
2) DisciplineSelection
3) Home
4) ManageDisciplines

so now with "add new discipline", we trigger the navigation to ScopeSelection, when using `navigation.navigate('ScopeSelection')` we would jump to 1) and our history will have only one entry:
1) ScopeSelection

when using `navigation.push('ScopeSelection)` our history will look like this:
1) ScopeSelection
2) DisciplineSelection
3) Home
4) ManageDisciplines
5) ScopeSelection

And therefore back navigation works.

klinzo:
> > > @klinzo Do you prefer reseting the history or using push?
> > 
> > 
> > I definitely prefer using push over resetting the history, nothing wrong with push, it's a public api of react-navigation (at least as long as there is a reason to use it over navigate). Still not sure I really understood why we need it though.
> 
> When starting the app for the first time select a discipline, confirm and then got to "settings" and want to add a new discipline our history looks like this:
> 
>     1. ScopeSelection
> 
>     2. DisciplineSelection
> 
>     3. Home
> 
>     4. ManageDisciplines
> 
> 
> so now with "add new discipline", we trigger the navigation to ScopeSelection, when using `navigation.navigate('ScopeSelection')` we would jump to 1) and our history will have only one entry (back navigation will fail as we are on the first entry of the history):
> 
>     1. ScopeSelection
> 
> 
> when using `navigation.push('ScopeSelection)` our history will look like this:
> 
>     1. ScopeSelection
> 
>     2. DisciplineSelection
> 
>     3. Home
> 
>     4. ManageDisciplines
> 
>     5. ScopeSelection
> 
> 
> And therefore back navigation works.

Thanks for explaining, now I understand. Perhaps reset is then the better solution since we never want to go back to the first two screens :see_no_evil: This is more like an intro screen then and should be replaced completely. Keeping the routes may lead to issues later on (like displaying a back button or going back if the hardware back button even though it is unwanted). Sorry for changing my mind here.

klinzo:
@SteffiStuffel it is even the case that it is possible to go back from the home screen to the initial selection. I don't think that should be possible right? So we should use `reset` here.

SteffiStuffel:
> @SteffiStuffel it is even the case that it is possible to go back from the home screen to the initial selection. I don't think that should be possible right? So we should use `reset` here.

Yes it is possible and also works, but not really the user flow we want to encourage. I'm fine with resetting

PR #299, opened 2022-05-12
LUN-339: Introduce generic load discipline hooks

SteffiStuffel:


src/hooks/useLoadDiscipline.ts
```diff
@@ -1,13 +1,25 @@
...
+    const response = await getFromEndpoint<ServerResponseGroup[]>(url, loadData.apiKey)
+    return formatGroup(response, loadData.apiKey)
+  }
+  const url = `${ENDPOINTS.discipline}/${loadData.disciplineId}`
+  const response = await getFromEndpoint<ServerResponseDiscipline>(url)
+  return formatDiscipline(response, { parent: null })
 }
 
-export const useLoadDiscipline = (id: number): Return<Discipline> => useLoadAsync(loadDiscipline, id)
+export const useLoadDiscipline = (loadData: DisciplineRequestData): Return<Discipline> =>
```
steffenkleinle, line 24:
```suggestion
export const useLoadDiscipline = (params: DisciplineRequestParams): Return<Discipline> =>
```

src/hooks/useLoadDisciplines.ts
```diff
@@ -1,37 +1,32 @@
...
-  total_training_sets: number
-  total_discipline_children: number
-  total_documents: number
-  nested_training_sets: number[]
-}
-
-const getEndpoint = (parent: Discipline | null): string => {
-  if (parent?.needsTrainingSetEndpoint) {
-    return ENDPOINTS.trainingSet
+const getEndpoint = (loadingInfo: DisciplinesRequestData): string => {
```
steffenkleinle, line 6: Same naming thing as for `useLoadDiscipline`

src/hooks/useLoadDisciplines.ts
```diff
@@ -1,37 +1,32 @@
...
-  return ENDPOINTS.disciplines
+  return `${ENDPOINTS.disciplines}/${loadingInfo.parent?.id ?? ''}`
 }
 
-const formatServerResponse = (serverResponse: ServerResponse[], parent: Discipline | null): Discipline[] =>
-  serverResponse.map(item => formatDiscipline(item, parent))
+const formatServerResponse = (
+  serverResponse: ServerResponseDiscipline[],
+  loadingInfo: DisciplinesRequestData
+): Discipline[] => serverResponse.map(item => formatDiscipline(item, loadingInfo))
```
steffenkleinle, line 22: Optional: Just inline this, imo no need to add a separate function here

src/routes/home/components/CustomDisciplineDetails.tsx
```diff
@@ -0,0 +1,47 @@
...
+  discipline: Discipline
+}
+
+const CustomDisciplineDetails = ({ discipline }: PropsType): JSX.Element => {
+  const navigation = useNavigation<StackNavigationProp<RoutesParams, 'Home'>>()
+
+  const navigate = (): void => {
+    navigation.navigate('DisciplineSelection', {
+      discipline
+    })
```
steffenkleinle, line 31: Why no progress saving/displaying for custom disciplines? What is the difference?

src/routes/home/components/DisciplineCard.tsx
```diff
@@ -1,91 +1,49 @@
...
 
-const UnitText = styled(ContentSecondary)`
+export const UnitText = styled(ContentSecondary)`
   font-size: ${props => props.theme.fonts.headingFontSize};
 `
 
-const ButtonContainer = styled.View`
+export const ButtonContainer = styled.View`
   margin: ${props => props.theme.spacings.xxs} auto;
 `
```
steffenkleinle, line 38: Should not live here if used in other components imo.

steffenkleinle (changes requested):
Nice work :)

src/routes/home/components/ProfessionDetails.tsx
```diff
@@ -0,0 +1,100 @@
...
+  discipline: Discipline
+}
+
+const ProfessionDetails = ({ discipline }: PropsType): ReactElement => {
+  const { data: nextExercise, refresh: refreshNextExercise } = useReadNextExercise(discipline)
+  const { data: progress, refresh: refreshProgress } = useReadProgress(discipline)
+
+  const moduleAlreadyStarted = progress !== null && progress !== 0
+  const [documents, setDocuments] = useState<Document[] | null>(null)
+  const navigation = useNavigation<StackNavigationProp<RoutesParams, 'Home'>>()
```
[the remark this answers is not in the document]
SteffiStuffel: I decided to use useNavigation() hook here, as I wanted to avoid passing prop to deeply nested child, as the docs say `useNavigation is a hook which gives access to navigation object. It's useful when you cannot pass the navigation prop into the component directly, or don't want to pass it in case of a deeply nested child.`

If we decide to pass the method as you suggest, we need to add a optional prop `navigateToNextExercise()` to `DisciplineCard` and then pass it to `ProfessionDetails` and then check if the prop exists.

Which way do you prefer?

src/hooks/helpers.ts
```diff
@@ -1,14 +1,62 @@
...
+  | {
+      apiKey: string
+    }
+  | {
+      disciplineId: number
+    }
+
+export const isTypeLoadProtected = (
+  value: DisciplineRequestData | DisciplinesRequestData
+): value is { apiKey: string } => Object.prototype.hasOwnProperty.call(value, 'apiKey')
```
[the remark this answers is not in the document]
SteffiStuffel: I moved types to hooks, but kept the type guard, changed it though.

src/hooks/useLoadDiscipline.ts
```diff
@@ -1,13 +1,25 @@
...
+  ServerResponseGroup
+} from './helpers'
 import useLoadAsync, { Return } from './useLoadAsync'
-import { ServerResponse } from './useLoadDisciplines'
 
-export const loadDiscipline = async (id: number): Promise<Discipline> => {
-  const url = `${ENDPOINTS.discipline}/${id}`
-  const response = await getFromEndpoint<ServerResponse>(url)
-  return formatDiscipline(response)
+export const loadDiscipline = async (loadData: DisciplineRequestData): Promise<Discipline> => {
```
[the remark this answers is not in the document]
SteffiStuffel: Changed to params, directly destructure is not possible in this case.

src/routes/home/components/CustomDisciplineDetails.tsx
SteffiStuffel: Custom disciplines do not necessarily have a logical order, e.g. structur can be 
- School xy
  - German
    - Class 12a
    - Class 12b

So there is no possibility to suggest the next module to the user. Progress is not displayed as in the designs, because there is no use of showing a user (s)he has done 10% of the modules of the school because 100% is not the aim. 

Saving progress of custom disciplines and show it in DisciplineSelectionScreen and ExercisesScreen makes sense, I will create a ticket for it.

src/routes/home/components/DisciplineCard.tsx
SteffiStuffel: will keep as no better options

src/hooks/useLoadDiscipline.ts
```diff
@@ -10,13 +9,21 @@ import {
 } from './helpers'
 import useLoadAsync, { Return } from './useLoadAsync'
 
-export const loadDiscipline = async (loadData: DisciplineRequestData): Promise<Discipline> => {
-  if (isTypeLoadProtected(loadData)) {
+export type DisciplineRequestData =
```
steffenkleinle, line 12: Please rename type

src/routes/home/components/CustomDisciplineDetails.tsx
steffenkleinle: Thanks for explaining

steffenkleinle (approved):

src/routes/home/HomeScreen.tsx
```diff
@@ -40,57 +37,27 @@ const HomeScreen = ({ navigation }: HomeScreenProps): JSX.Element => {
   const { data: selectedProfessions, refresh: refreshSelectedProfessions } = useReadSelectedProfessions()
   const isCustomDisciplineEmpty = !customDisciplines || customDisciplines.length <= 0
 
-  useFocusEffect(
-    useCallback(() => {
-      refreshCustomDisciplines()
-      refreshSelectedProfessions()
-    }, [refreshCustomDisciplines, refreshSelectedProfessions])
-  )
+  useFocusEffect(refreshCustomDisciplines)
```
f1sh1918, line 40: why did you change these lines and removed the `useCallback`

f1sh1918 (approved):
Nice refactoring. Works fine
just some small general questions.
Tested on ios

src/routes/home/HomeScreen.tsx
steffenkleinle: because I requested it :D `refreshCustomDisciplines` and `refreshSelectedProfessions` are already wrapped in `useCallback`s and therefore no additional `useCallback` is needed if called separately. I think it also makes the code a little bit easier to read and there is no need to group both refresh together.

src/hooks/useLoadDisciplines.ts
```diff
@@ -1,37 +1,35 @@
...
+    }
 
-const getEndpoint = (parent: Discipline | null): string => {
-  if (parent?.needsTrainingSetEndpoint) {
-    return ENDPOINTS.trainingSet
+const getEndpoint = (params: DisciplinesRequestData): string => {
+  if (isTypeLoadProtected(params)) {
+    return ENDPOINTS.groupInfo
+  }
+  if (params.parent?.needsTrainingSetEndpoint) {
```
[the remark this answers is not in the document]
SteffiStuffel: not all disciplines are get by the /disciplines/{parentId} endpoint. If they are leaf disciplines they need to get by the endpoint /training_set/{parentId}. This is just because of design of the cms endpoints.

PR #302, opened 2022-05-18
LUN-265: Bottom tab navigator

steffenkleinle:


steffenkleinle:
Tests still missing, can already be reviewed/tested though.

src/navigation/BottomTabNavigator.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const theme = useTheme()
+
+  return (
+    <BottomTabNavigator.Navigator
+      screenOptions={{
+        headerShown: false,
+        tabBarActiveTintColor: theme.colors.background,
+        tabBarStyle: { backgroundColor: theme.colors.primary, height: wp('14%') },
+        tabBarItemStyle: { height: wp('14%'), padding: wp('2%') },
+        tabBarLabelStyle: { fontSize: wp('3%') }
```
f1sh1918, line 35: pls use small font size from theme

src/navigation/BottomTabNavigator.tsx
```diff
@@ -0,0 +1,93 @@
...
+import labels from '../constants/labels.json'
+import HomeStackNavigator from './HomeStackNavigator'
+import { RoutesParams } from './NavigationTypes'
+
+const BottomTabNavigator = createBottomTabNavigator<RoutesParams>()
+
+// TODO LUN-132, LUN-207, LUN-308: Remove this and use actual components
+const MockComponent = () => <></>
+
+const Navigator = (): JSX.Element | null => {
```
f1sh1918, line 25: name should match filename

src/navigation/BottomTabNavigator.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const theme = useTheme()
+
+  return (
```
f1sh1918, line 28: you should wrap the component with a  `<SafeAreaView style={{flex: 1}}>` to avoid overlapping bottom sheet on ios devices

f1sh1918 (changes requested):
Really nice work, just some small things and a fix for some ios devices because the tab bar items overlap. Tested on ios

image

assets/images/index.ts
```diff
@@ -5,6 +5,8 @@ import ArrowRightIcon from './arrow-right-icon.svg'
 import BannerGreen from './banner-green.png'
 import BannerRed from './banner-red.png'
 import BannerYellow from './banner-yellow.png'
+import BookOutlineIconGrey from './book-outline-icon-grey.svg'
```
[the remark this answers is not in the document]
steffenkleinle: Sounds good, please open an issue for this. Will keep this for now.

src/navigation/BottomTabNavigator.tsx
```diff
@@ -0,0 +1,93 @@
...
+  StarOutlineIconWhite
+} from '../../assets/images'
+import labels from '../constants/labels.json'
+import HomeStackNavigator from './HomeStackNavigator'
+import { RoutesParams } from './NavigationTypes'
+
+const BottomTabNavigator = createBottomTabNavigator<RoutesParams>()
+
+// TODO LUN-132, LUN-207, LUN-308: Remove this and use actual components
+const MockComponent = () => <></>
```
SteffiStuffel, line 23: Only valid options should be displayed. Please hide the bottom bar as long as only home is available.

src/navigation/HomeStackNavigator.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const { manageDisciplines, overview } = labels.general.header
+  const theme = useTheme()
+
+  if (loading) {
+    return null
+  }
+
+  return (
+    <Stack.Navigator
+      initialRouteName={professions !== null ? 'Home' : 'ScopeSelection'}
```
SteffiStuffel, line 38: (optional as i guess this is no change from you): Avoid using negative comparision.

src/navigation/NavigationTypes.ts
```diff
@@ -30,6 +30,11 @@ type ResultParams = ExerciseParams & {
 // https://github.com/Microsoft/Script/issues/15300
 // eslint-disable-next-line @typescript-eslint/consistent-type-definitions
 export type RoutesParams = {
+  BottomTabNavigator: undefined
+  HomeTab: undefined
+  FavoritesTab: undefined
+  DictionaryTab: undefined
+  UserVocabularyTab: undefined
   Home: undefined
```
SteffiStuffel, line 38: Why do we need `Home` and `HomeTab`?

src/navigation/Navigator.tsx
```diff
@@ -1,242 +1,55 @@
...
-          component={ImprintScreen}
-        />
-        <Stack.Screen
-          options={({ navigation }) =>
-            defaultOptions(labels.general.header.overview, ArrowLeftCircleIconWhite, navigation, false)
-          }
-          name='ManageDisciplines'
-          component={ManageSelectionsScreen}
-        />
+        <Stack.Screen name='ExerciseFinished' component={ExerciseFinishedScreen} options={{ headerShown: false }} />
```
SteffiStuffel, line 49: I think we should also add Result and ResultDetails routes here, so the app has a valid state until the ticket, where they get removed is done.

src/services/screenOptions.tsx
```diff
@@ -0,0 +1,59 @@
+import { NavigationProp } from '@react-navigation/native'
```
SteffiStuffel, line 1: Good idea to move this to a own file

Why have you decided to put this file not into navigation directory? 

just because it catched my eyes: I think we should think about good naming of services.

SteffiStuffel (approved):
Tested on android, works fine.

Back navigation does not work as in youtube app, but as it works in google podcasts, so i think this is fine. Maybe it would be nice to document somewhere (or at least for Martin in this ticket) what the behaviour is, we want to have.

Please delete bahaas old branch for this ticket if no longer needed.

Please hide bottom bar until the favorites are ready, so we can still release.

assets/images/index.ts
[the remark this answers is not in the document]
f1sh1918: [LUN-356](https://issues.tuerantuer.org/browse/LUN-356)

src/navigation/BottomTabNavigator.tsx
steffenkleinle: Not possible, small font size is a string including `px` while a number is required

steffenkleinle:
> Back navigation does not work as in youtube app, but as it works in google podcasts, so i think this is fine. Maybe it would be nice to document somewhere (or at least for Martin in this ticket) what the behaviour is, we want to have.

What are the differences? This is just the default reactnavigation behavior, right? Not sure why we should document that.

> Please delete bahaas old branch for this ticket if no longer needed.

Done

> Please hide bottom bar until the favorites are ready, so we can still release.

Done

SteffiStuffel (approved):

SteffiStuffel:
Nice work. I like it! :)

f1sh1918 (approved):

PR #307, opened 2022-05-30
LUN-132: Favorites

steffenkleinle:

Hopefully my final favorites PR :tada:


src/routes/FavoritesScreen.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  return (
+    <ServerResponseHandler error={error} loading={false} refresh={refresh}>
+      {data && (
+        <VocabularyList
+          documents={data}
+          refreshFavorites={refresh}
+          onItemPress={(index: number) =>
+            navigation.navigate('VocabularyDetail', {
+              disciplineId: 0,
```
steffenkleinle, line 30: @SteffiStuffel what did we decide here? There is obviously no discipline id, but it is used everywhere to save/read progress and so on...

src/components/FavoriteButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+  shadow-opacity: 0.5;
+`
+
+interface Props {
+  document: Document
+  refreshFavorites?: () => void
+}
+
+const FavoriteButton = ({ document, refreshFavorites }: Props): ReactElement | null => {
+  const { data: isFavorite, refresh } = useLoadAsync(AsyncStorage.isFavorite, document)
```
SteffiStuffel, line 40: Why refresh and refreshFavorites?

src/components/DocumentImageSection.tsx
```diff
@@ -0,0 +1,49 @@
...
+`
+
+interface Props {
+  document: Document
+  audioDisabled?: boolean
+  minimized?: boolean
+  submittedAlternative?: string | null
+}
+
+const DocumentImageSection = ({
```
SteffiStuffel, line 32: (opional) rename to WordImageSection

src/constants/labels.json
```diff
@@ -141,5 +141,10 @@
     "disciplines": "Module",
     "word": "Wort",
     "words": "Wörter"
+  },
+  "favorites": {
+    "favorites": "Favoriten",
+    "add": "Favorit hinzufügen",
```
SteffiStuffel, line 147: I think add and remove can be removed. See comment above

src/routes/FavoritesScreen.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const { data, refresh, error } = useLoadAsync(AsyncStorage.getFavorites, {})
+
+  useFocusEffect(refresh)
+
+  const onItemPress = (index: number) => {
+    if (!data) {
+      return
+    }
+    navigation.navigate('VocabularyDetail', {
+      disciplineId: 0,
```
SteffiStuffel, line 27: VocabularyDetail route does not need disciplineId and disciplineTitle, does it? Should be removed.

src/services/AsyncStorage.ts
```diff
@@ -76,6 +80,35 @@ export const saveExerciseProgress = async (
   await setExerciseProgress(disciplineId, exerciseKey, score)
 }
 
+const getFavorites = async (): Promise<Document[]> => {
+  const documents = await AsyncStorage.getItem(FAVORITES_KEY)
+  return documents ? JSON.parse(documents) : []
+}
+
+const setFavorites = async (favorites: Document[]): Promise<void> => {
```
SteffiStuffel, line 88: I think it would make more sense to only save id's of the favorites, and request the other data from the api, as images improve, alternative words are added and so on.

src/routes/FavoritesScreen.tsx
```diff
@@ -0,0 +1,49 @@
...
+      disciplineTitle: labels.general.favorites,
+      documents: data,
+      documentIndex: index,
+      closeExerciseAction: CommonActions.goBack()
+    })
+  }
+
+  return (
+    <ServerResponseHandler error={error} loading={false} refresh={refresh}>
```
SteffiStuffel, line 36: I think we use ServerResponseHandler only for requests from api not for loading from asyncStorage. If we switch to use it also from loading from AsyncStorage it should be renamed

src/routes/FavoritesScreen.tsx
SteffiStuffel: I think it is not used in VocabularyScreen

SteffiStuffel:
Tested on Android emulator works like a charm. :+1:

src/components/DocumentImageSection.tsx
steffenkleinle: Why? Did I miss that we decided to switch from document to word naming wise? afaik it is always called document

src/components/FavoriteButton.tsx
steffenkleinle: One refreshes the button and one the favorites list which is sadly both needed I think. Do you have another solution?

src/routes/manage-selections/ManageSelectionsScreen.tsx
```diff
@@ -42,14 +42,14 @@ const ManageSelectionsScreen = ({ navigation }: Props): ReactElement => {
 
   const professionItems = selectedProfessions?.map(id => {
     const unselectProfessionAndRefresh = () => {
-      removeSelectedProfession(id).then(refreshSelectedProfessions).catch(reportError)
+      AsyncStorage.removeSelectedProfession(id).then(refreshSelectedProfessions).catch(reportError)
```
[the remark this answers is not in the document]
steffenkleinle: It does not make sense to have both and since the usage via `AsyncStorage` was more common I removed the other occurrences and single exports.

src/components/DocumentImageSection.tsx
SteffiStuffel: Yeah, naming is messy, thats why I marked it opional. But I think we will replace document naming, as these are not really documents. :thinking:

src/services/AsyncStorage.ts
steffenkleinle: Blocked by https://issues.tuerantuer.org/browse/LUN-307 (or rather: https://github.com/digitalfabrik/lunes-cms/issues/324).

src/components/DocumentImageSection.tsx
steffenkleinle: Hmm word is not really a good alternative though imo, it should probably be called something like vocab or similar. But since right now document is the current standard I'll keep it, may be renamed when we decide to switch to something else.

src/components/FavoriteButton.tsx
SteffiStuffel: maybe renaming could make it clearer. E.g. refresh as it is, an refreshParent oder refreshList or something?

src/routes/FavoritesScreen.tsx
steffenkleinle: It is for `VocabularyList` to save the progress.

src/services/AsyncStorage.ts
SteffiStuffel: hmh, we can not deploy your version and switch later, as users may have already saved whole words as favorites. 

So either we have to disable this feature or we use one of the existing endpoints and optimize this, when the cms provides the new endpoint. I prefer the second.

src/components/DocumentImageSection.tsx
SteffiStuffel: fine for me. But vocab is - like vocabulary - only a plural word, so not really fitting.

src/routes/FavoritesScreen.tsx
SteffiStuffel: We have the VocabularyListScreen there it is used and needed. But in VocabularyDetailScreen it is not used and can be removed from route params.

Not sure if we should set it to `null` and check for non-null or if sticking to `0` is fine, seems to be less robust.

src/components/DocumentImageSection.tsx
steffenkleinle: yeah, we can discuss this when we want to change this :)

SteffiStuffel:
Please also add a release note

src/components/FavoriteButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }
+
+  if (isFavorite === null) {
+    return null
+  }
+
+  return (
+    <Container>
+      <Button accessibilityLabel={isFavorite ? labels.favorites.remove : labels.favorites.add} onPress={onPress}>
```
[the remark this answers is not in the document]
steffenkleinle: Couldn't get it to work with a dynamic test id and don't want to spend anymore time on this as it works fine with accessibility label.

src/components/FavoriteButton.tsx
steffenkleinle: Renamed to `onFavoritesChanged` as it doesn't matter if it refreshes or does something entirely else.

src/routes/FavoritesScreen.tsx
steffenkleinle: I'll change it to `null` I guess

src/routes/FavoritesScreen.tsx
SteffiStuffel: oke, fine for me. 

Not sure if it is bad practice to use "0", as this should be "null" or "undefined".

src/components/FavoriteButton.tsx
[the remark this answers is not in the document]
SteffiStuffel: Not sure, what the problem ist, but changing it to testID works just fine, see:

https://github.com/digitalfabrik/lunes-app/pull/311

src/components/VocabularyListItem.tsx
```diff
@@ -50,9 +52,10 @@ const VocabularyListItem = ({ document, onPress }: VocabularyListItemProps): Rea
       onPress={onPress}
       icon={icon}
       rightChildren={
-        <Speaker>
+        <RightChildrenContainer>
           <AudioPlayer document={document} disabled={false} />
-        </Speaker>
+          <FavoriteButton document={document} onFavoritesChanged={onFavoritesChanged} />
```
f1sh1918, line 57: Surround fav button with a view that sets the margin
```js
const ButtonContainer = styled.View`
  margin: ${props => `0 ${props.theme.spacings.sm}`};
  align-self: center;
`
```

src/components/FavoriteButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+  min-height: ${wp('9%')}px;
+`
+const IconOutline = styled(StarCircleIconGrey)`
+  min-width: ${wp('9%')}px;
+  min-height: ${wp('9%')}px;
+`
+const Button = styled.TouchableOpacity`
+  justify-content: center;
+  align-items: center;
+  shadow-radius: 5px;
```
f1sh1918, line 29: add here shadow-color as for the audio player. i think we should have the same shadow color for all the icons
```js
shadow-color: ${props => props.theme.colors.shadow};
```

src/services/AsyncStorage.ts
f1sh1918: I'm also with steffi even if it gonna be quite slow at the moment (2,5s), i would prefer to save ids and maybe fetch all the words and filter in the frontend. Lateron we just pass these wordIds to the new endpoint

f1sh1918 (commented):
Please fix styling issues like shadow and positioning. I thought it would be possible that you can just do
`styled(FavoriteButton)` and set the margin but seems like that doesn't work

src/components/FavoriteButton.tsx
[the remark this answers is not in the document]
steffenkleinle: no idea why it was not working, thx

src/routes/FavoritesScreen.tsx
steffenkleinle: removed it after all

steffenkleinle: changed to null

src/services/AsyncStorage.ts
steffenkleinle: done

src/components/FavoriteButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+import { StarCircleIconGrey, StarCircleIconGreyFilled } from '../../assets/images'
+import { Document } from '../constants/endpoints'
+import labels from '../constants/labels.json'
+import useLoadAsync from '../hooks/useLoadAsync'
+import AsyncStorage from '../services/AsyncStorage'
+import { reportError } from '../services/sentry'
+
+const Container = styled.View`
+  padding: ${props => `${props.theme.spacings.xs} 0  ${props.theme.spacings.xs} ${props.theme.spacings.sm}`};
```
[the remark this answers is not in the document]
f1sh1918: fixed

src/components/FavoriteButton.tsx
f1sh1918: added shadow

src/components/VocabularyListItem.tsx
f1sh1918: added container

SteffiStuffel (approved):
Now you can enable the feature and bottom navigation bar.

PR #308, opened 2022-05-31
LUN-131: Logic for unlocking exercises

SteffiStuffel:


src/routes/choice-exercises/components/SingleChoiceExercise.tsx
```diff
@@ -81,6 +81,7 @@ const ChoiceExerciseScreen = ({
   }, [results, currentWord])
 
   const onExerciseFinished = async (results: DocumentResult[]): Promise<void> => {
+    const progress = await getExerciseProgress()
     if (disciplineId) {
```
steffenkleinle, line 85: Why do we have to check for disciplineId here? Does it make sense to get the progess before saving new progress?

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
```diff
@@ -90,7 +91,8 @@ const ChoiceExerciseScreen = ({
       disciplineTitle,
       exercise: exerciseKey,
       results,
-      closeExerciseAction: route.params.closeExerciseAction
+      closeExerciseAction: route.params.closeExerciseAction,
+      unlockedNextExercise: progress[disciplineId] ? !progress[disciplineId]![exerciseKey] : true
```
steffenkleinle, line 95: why is this dependent on the previous progress and not the current results? I find this quite hard to understand, can you explain? What are the possible values of `progress[disciplineId][exerciseKey]`?

src/routes/exercise-finished/ExerciseFinishedScreen.tsx
```diff
@@ -54,16 +54,21 @@ interface Props {
...
 const ExerciseFinishedScreen = ({ navigation, route }: Props): ReactElement => {
-  const { exercise, results, disciplineTitle, disciplineId, documents, closeExerciseAction } = route.params
+  const {
+    exercise,
+    results,
+    disciplineTitle,
+    disciplineId,
+    documents,
+    closeExerciseAction,
+    unlockedNextExercise = false
```
steffenkleinle, line 64: Why is this optional with default to false? The only place where nothing is set is `WriteExerciseScreen`, why? Because it is the last exercise and there is no next exercise? Why not just pass `false` there explicitly?

src/routes/exercises/ExercisesScreen.tsx
```diff
@@ -56,10 +58,14 @@ const ExercisesScreen = ({ route, navigation }: ExercisesScreenProps): JSX.Eleme
     apiKey: discipline.apiKey
   })
 
-  const currentExercise: Exercise = EXERCISES[2] //  TODO: LUN-131 logic , note: currentExercise is the last level that can be accessed.
+  useFocusEffect(() => {
+    getDoneExercises(disciplineId)
+      .then(value => setNextExercise(value < EXERCISES.length ? EXERCISES[value] : null))
+      .catch(reportError)
+  })
```
steffenkleinle, line 65: Shouldn't this be wrapped in a `useCallback`?

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
SteffiStuffel: possible values are a score from 0 to 1 or undefined. With negation at the beginning we just check if it is undefined, which means the exercise was not done before, therefore is done the first time and therefore unlocks the next exercise.

src/routes/exercises/ExercisesScreen.tsx
SteffiStuffel: good catch, it should. Docs say:

`Note: To avoid the running the effect too often, it's important to wrap the callback in useCallback before passing it to useFocusEffect as shown in the example.`

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
steffenkleinle: but negation in the beginning also returns true if the score is `0`, doesn't it? I think it should be explicitely `!== undefined`

SteffiStuffel: you are right. fixed it.

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
```diff
@@ -81,16 +81,16 @@ const ChoiceExerciseScreen = ({
...
+    await saveExerciseProgress(disciplineId, exerciseKey, results)
     navigation.navigate('ExerciseFinished', {
       documents,
       disciplineId,
       disciplineTitle,
       exercise: exerciseKey,
       results,
-      closeExerciseAction: route.params.closeExerciseAction
+      closeExerciseAction: route.params.closeExerciseAction,
+      unlockedNextExercise: progress[disciplineId] ? progress[disciplineId]![exerciseKey] === undefined : true
```
steffenkleinle, line 93: If I understand this correctly this should be the same as
```suggestion
      unlockedNextExercise: progress[disciplineId]?.[exerciseKey] === undefined
```

steffenkleinle (approved):
Tested on android, works fine :tada:

f1sh1918 (approved):
really nice work :) works fine on ios

PR #309, opened 2022-05-31
LUN-351: Write Exercise incorrect answer validation 

steffenkleinle:


release-notes/unreleased/LUN-351-separated-words-wrong.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: LUN-280
+show_in_stores: true
+platforms:
+  - android
+  - ios
+de: Richtige Antworten für getrennte Wörter werden nun nicht mehr als falsch angezeigt.
```
SteffiStuffel, line 6: (opional) werden nun korrekt ausgewertet.

src/routes/write-exercise/components/__tests__/InteractionSection.spec.tsx
```diff
@@ -72,4 +72,68 @@ describe('InteractionSection', () => {
     fireEvent.press(button)
     await waitFor(() => expect(getByTestId('popover').props.isVisible).toBeTruthy())
   })
+
```
SteffiStuffel, line 75: Can you add a test case for the bug you fixed, so if the correct solution is built up of multiple words?

SteffiStuffel (approved):
Tested on Android emulator. "Der kontaktlose Spannungsprüfer" is correct now ;)

f1sh1918 (approved):
tested on ios. works fine

PR #314, opened 2022-06-14
LUN-287: Progress spinner for discipline selection

SteffiStuffel:


f1sh1918 (approved):
Looks really nice. So we use the word list also as a stage for progress. So if you just opened the wordlist you get 25% for the discipline?!

SteffiStuffel:
> Looks really nice. So we use the word list also as a stage for progress. So if you just opened the wordlist you get 25% for the discipline?!

yes, correct. We decided this makes most sense.

src/services/helpers.ts
```diff
@@ -108,9 +108,12 @@ export const getNextExercise = async (profession: Discipline): Promise<NextExerc
 }
 
 export const getProgress = async (profession: Discipline | null): Promise<number> => {
-  if (!profession || !profession.leafDisciplines) {
+  if (!profession) {
     return 0
   }
+  if (!profession.leafDisciplines) {
+    return (await getDoneExercises(profession.id)) / EXERCISES.length
```
steffenkleinle, line 115:
```suggestion
    const doneExercises = await getDoneExercises(profession.id)
    return doneExercises / EXERCISES.length
```

src/components/DisciplineListItem.tsx
```diff
@@ -10,27 +16,66 @@ interface DisciplineListItemProps {
...
+  useFocusEffect(() => {
+    if (showProgress) {
+      getProgress(item).then(setProgress).catch(reportError)
+    }
+  })
+
+  const iconWithProgress = (
+    <>
+      <Progress.Circle
+        progress={progress}
```
steffenkleinle, line 57:
```suggestion
  const { data: progress } = useReadProgress(item)

  const iconWithProgress = (
    <>
      <Progress.Circle
        progress={progress ?? 0}
```

src/components/DisciplineListItem.tsx
```diff
@@ -10,27 +16,66 @@ interface DisciplineListItemProps {
...
+        size={Math.round(wp('14%'))}
```
steffenkleinle, line 58: Why do we need `Math.round` around this? Can it only handle integers?

steffenkleinle (approved):
Tested on android :heart_eyes:

src/services/helpers.ts
SteffiStuffel: Can you explain, why you prefer this?

src/components/DisciplineListItem.tsx
SteffiStuffel: I think this does not really make sense, as we would make an slow reading request, also if it is not needed, because progress is not displayed.

src/components/DisciplineListItem.tsx
SteffiStuffel: unfortunately yes. Behavior is weird for non-integer values.

src/components/DisciplineListItem.tsx
steffenkleinle: somehow missed the if clause, nvm :)

SteffiStuffel: Well it is reading from memory, but "slow" is relative. But there are multiple requests because the screen render multiple DisicplineListItems, therefore I would prefer to stick to not load if not needed, although it wouldn't be a big problem to do so.

PR #324, opened 2022-07-20
LUN-368: Wrong button label on profession selection screen

charludo:


steffenkleinle (changes requested):
Requesting changes until https://chat.tuerantuer.org/digitalfabrik/pl/7pnk46k3rfnrpdg5unqerqe1nc is discussed

steffenkleinle:
Sorry for the bad preparation/communication regarding this issue, please adjust this such that:
* The next exercise card is always shown and
  * has a `Starten` label if no exercise was done before
  * has a `Weitermachen` label else
* The shown layout/components are the same for both cases

I.e. just keep the previous state and only adjust the labels. Sorry again!

src/routes/home/components/__tests__/ProfessionDetails.spec.tsx
```diff
@@ -36,18 +36,27 @@ describe('ProfessionDetails', () => {
       />
     )
 
-  it('should show next exercise details on the card', () => {
-    mocked(useReadProgress).mockReturnValue(getReturnOf(0))
+  it('should show next exercise details on the card if at least one module is marked done', () => {
+    mocked(useReadProgress).mockReturnValue(getReturnOf(1))
```
steffenkleinle, line 40: Please revert this, `if at least one module is marked done` is misleading now.

src/routes/home/components/ProfessionDetails.tsx
```diff
@@ -61,17 +63,16 @@ const ProfessionDetails = ({
         />
         {discipline.leafDisciplines && (
           <NumberText>
-            {moduleAlreadyStarted && `${Math.floor(progress * discipline.leafDisciplines.length)}/`}
-            {discipline.leafDisciplines.length}
+            {moduleAlreadyStarted && doneModules}/{discipline.leafDisciplines.length}
```
steffenkleinle, line 66: Check for `moduleAlreadyStarted` is redundant since it is already checked in line 40 for `doneModules`
```suggestion
            {doneModules}/{discipline.leafDisciplines.length}
```
Edit: Seems like it breaks the behavior since it now just shows `/24 Module` instead of `0/24 Module`

steffenkleinle (changes requested):
Tested on android.

src/routes/home/components/ProfessionDetails.tsx
charludo: couldn't replicate the breaking behavior, including for completely new jobs - is this still an issue?

src/routes/home/components/ProfessionDetails.tsx
```diff
@@ -61,17 +63,16 @@ const ProfessionDetails = ({
...
         )}
-        <UnitText>{moduleAlreadyStarted ? labels.home.progressDescription : childrenLabel(discipline, true)}</UnitText>
+        <UnitText>{doneModules > 0 ? labels.home.progressDescription : childrenLabel(discipline, true)}</UnitText>
       </ProgressContainer>
       <NextExerciseCard
         thumbnail={documents[0].document_image[0].image}
         onPress={() => navigateToNextExercise(nextExerciseData)}
         heading={EXERCISES[exerciseKey].title}
-        buttonLabel={labels.home.continue}
+        buttonLabel={doneModules > 0 ? labels.home.continue : labels.home.start}
```
SteffiStuffel, line 75: I think here we need moduleAlreadyStarted instead of doneModules.

src/routes/home/components/ProfessionDetails.tsx
```diff
@@ -36,6 +36,8 @@ const ProfessionDetails = ({
   const { data: nextExerciseData, refresh: refreshNextExercise } = useLoadNextExercise(discipline)
 
   const moduleAlreadyStarted = progress !== null && progress !== 0
+  const doneModules =
```
SteffiStuffel, line 39: (optional) I think it would make sense to use this `doneModules` variable for the progress spinner.

SteffiStuffel (commented):
I get "Starten" label for both - new and partly done - professions, only if at least one module was completely finished it shows "Fortsetzen", but it should show "fortsetzen", if one exercise was finished.

charludo:
> I get "Starten" label for both - new and partly done - professions, only if at least one module was completely finished it shows "Fortsetzen", but it should show "fortsetzen", if one exercise was finished.

Interesting point. I interpreted the button (starten/fortsetzen) as relating to the specific level shown above it (e.g. "Level 1 - Grundlagen Streichen") rather than the entire profession.

In the current state it's either "let's start this new level!" or "let's continue where you left off in this level!", but I can see how "let's start/continue learning about this profession!" also makes sense.

I'd rather leave that decision up to you though, let me know if it should be changed.

steffenkleinle:
> Interesting point. I interpreted the button (starten/fortsetzen) as relating to the specific level shown above it (e.g. "Level 1 - Grundlagen Streichen") rather than the entire profession.
> 
> In the current state it's either "let's start this new level!" or "let's continue where you left off in this level!", but I can see how "let's start/continue learning about this profession!" also makes sense.

Good point! I think this should be answered from a project managment/UI/UX perspective. Would you mind asking this in ~lunes-app on mattermost? Thanks!

SteffiStuffel:
> > I get "Starten" label for both - new and partly done - professions, only if at least one module was completely finished it shows "Fortsetzen", but it should show "fortsetzen", if one exercise was finished.
> 
> Interesting point. I interpreted the button (starten/fortsetzen) as relating to the specific level shown above it (e.g. "Level 1 - Grundlagen Streichen") rather than the entire profession.
> 
> In the current state it's either "let's start this new level!" or "let's continue where you left off in this level!", but I can see how "let's start/continue learning about this profession!" also makes sense.
> 
> I'd rather leave that decision up to you though, let me know if it should be changed.

I agree with you and @steffenkleinle this interpretation makes sense. Lets keep it this way (as it is already implemented already and really makes sense), but i think it is not working correctly.

So if I start an profession and finish
exercise 1-4 of module 1, then it should show "start" exercise 1 of module 2, but it shows continue

src/routes/home/components/ProfessionDetails.tsx
charludo: Disregard my previous comment, you were right. Shows 'fortsetzen' as soon as one level (not module) has been completed now.

charludo:
@SteffiStuffel Oh OK, that's a third option then, but I agree, that also makes sense. Will implement.

steffenkleinle (approved):
I like it :+1: Tested on android

release-notes/unreleased/LUN-368-wrong-button-label-on-profession-selection-screen.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: LUN-368
+show_in_stores: true
+platforms:
+  - android
+  - ios
+de: Klarerer Buttonname für unangefangene Übungen
```
f1sh1918, line 6:
```suggestion
de: Treffenderer Buttonname bei noch nicht begonnenen Übungen
```

f1sh1918 (approved):

PR #325, opened 2022-07-25
LUN-370: Change react hook exhaustive deps rule

f1sh1918:


src/hooks/useLoadAsync.ts
```diff
@@ -35,7 +35,8 @@ export const useLoadAsync = <T, P>(request: (params: P) => Promise<T>, params: P
 
   const load = useCallback(() => {
     loadAsync<T, P>(request, params, setData, setError, setLoading).catch(e => setError(e))
-    // TODO LUN-330 fix object comparison
+    /* For small objects JSON.stringify is good enough. But does not work with ESLint rule */
```
f1sh1918, line 38: since we don't send super large param objects `JSON.stringify`is fine. There are also other community options like this: https://github.com/kentcdodds/use-deep-compare-effect
But its not often updated

src/components/ExerciseHeader.tsx
```diff
@@ -43,12 +43,12 @@ interface ExerciseHeaderProps {
 }
 
 const ExerciseHeader = ({
-  navigation,
-  closeExerciseAction,
-  currentWord,
-  numberOfWords,
-  confirmClose = true
-}: ExerciseHeaderProps): JSX.Element => {
+                          navigation,
```
SteffiStuffel, line 46: I guess this will be fixed by prettier?

SteffiStuffel (approved):
Looks good, not tested.

steffenkleinle:
@f1sh1918 please fix prettier then I'll review

src/components/ExerciseHeader.tsx
f1sh1918: yes somehow my prettier config didn't run on format. Gonna fix with next commit

steffenkleinle (approved):

PR #326, opened 2022-07-26
LUN-367: Hide bottom tab bar for initial selection

steffenkleinle:


steffenkleinle:
Sadly it is not really possible to only hide the bottom tab bar if it is the initial selection, therefore I removed the bottom tab bar for every profession selection (which is fine imo). If this is a problem, the only solution would be to duplicate the route.

f1sh1918 (approved):
Tested on ios. 
SafeAreaView is missing so "Überspringen" Button is slightly overlapping bottom handler for ScopeSelection Screen
Gonna fix that in LUN-369 if we don't plan to release before

steffenkleinle:
> Tested on ios. SafeAreaView is missing so "Überspringen" Button is slightly overlapping bottom handler for ScopeSelection Screen Gonna fix that in LUN-369 if we don't plan to release before
> 
> 

Was missing before, wasn't it? So won't fix here and lets do it in your issue :)

SteffiStuffel (approved):
Looks good. 
Tested on Android emulator.

PR #330, opened 2022-07-28
LUN-383: Fav icon displaced

f1sh1918:


release-notes/unreleased/LUN-383-fav-icon-displaced.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: LUN-383
+show_in_stores: false
+platforms:
+  - android
+  - ios
+de: Das Favoriten Icon in Übungen wurde korrekt positioniert.
```
SteffiStuffel, line 6: (optional) I like when release notes sound more optimistic and less "buggy". E.g. "Die Positionierung des Icons wurde optimiert".

SteffiStuffel (approved):
Still not perfect, but fine for me.
icon

ios/Podfile.lock
```diff
@@ -638,4 +638,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: ca4eb736278d4211ffaa5cef62299ef8fc2137ca
 
-COCOAPODS: 1.11.2
+COCOAPODS: 1.11.3
```
steffenkleinle, line 641: I think this should be reverted

steffenkleinle (approved):

ios/Podfile.lock
f1sh1918: alright but it should be updated

f1sh1918:
> Still not perfect, but fine for me. icon

This relies to how the picture is positioned. Works different on android. On ios it looks way better. Maybe we could create a ticket for that picture issue

ios/Podfile.lock
steffenkleinle: Should be updated together with gems :) So yes, please do it in a separate PR

PR #331, opened 2022-07-29
LUN-386: Upgrade to yarn 2

sarahsporck:


.circleci/config.yml
```diff
@@ -187,29 +187,28 @@ commands:
...
+                command: yarn install --immutable
+                name: Install Dependencies
                 working_directory: tools
             - save_cache:
                 key: 2-yarn-{{ arch }}-{{ checksum "tools/yarn.lock" }}-{{ checksum "tools/package.json" }}
                 name: Save Yarn Tools Package Cache
                 paths:
-                    - ~/.cache/yarn-tools
-                    - tools/node_modules
+                    - tools/.yarn/cache
```
SteffiStuffel, line 211: Have you checked, if caching is working?

steffenkleinle:
@sarahsporck can you explain a little bit what was done here and why? Or share a resource you followed to do so?

SteffiStuffel (approved):

f1sh1918 (approved):
Nice work :) tested on ios

.circleci/src/jobs/check.yml
```diff
@@ -2,9 +2,9 @@ docker:
   - image: cimg/node:14.19.0
 steps:
   - checkout
-  - check_config
   - restore_yarn_cache
   - restore_yarn_tools_cache
+  - check_config
```
steffenkleinle, line 7: Any specific reason why you changed the order here? If not, I'd prefer to keep the old order since builds will fail faster/use less credits if the config is wrong.

.circleci/src/jobs/bump_version.yml
```diff
@@ -11,11 +11,11 @@ steps:
   - restore_yarn_tools_cache
   - run:
       name: Calculate next version name
-      command: echo "export NEW_VERSION_NAME=$(yarn --silent next-version calc | jq .versionName)" >> ${BASH_ENV}
+      command: echo "export NEW_VERSION_NAME=$(yarn next-version calc | jq .versionName)" >> ${BASH_ENV}
```
steffenkleinle, line 14: Is it now `silent` as default?
Edit: Looks good anyway, seems like there is no such flag anymore.

steffenkleinle:
You can disable telemetry (its still a facebook tool):
>  yarn config set --home enableTelemetry 0

.yarnrc.yml
```diff
@@ -0,0 +1,15 @@
+# compression does not really make sense with node modules
+compressionLevel: 0
+
+# react native does not yet support pnp https://yarnpkg.com/features/pnp#incompatible
```
steffenkleinle, line 4: Does it make sense to create an issue to change this once react-native does support it?

steffenkleinle (approved):
Working fine, nice work! I hope this will really reduce problems in the future :)

.yarnrc.yml
sarahsporck: I think this will will still take some time up until rn will support this, therefore, I don't see us creating an issue for this. But it would definitely make sense to use pnp, when it is possiple.

.circleci/config.yml
sarahsporck: yesss 👍🏼

sarahsporck:
> @sarahsporck can you explain a little bit what was done here and why? Or share a resource you followed to do so?

See here: https://yarnpkg.com/getting-started/migration

.circleci/src/jobs/check.yml
sarahsporck: not possible: Couldn't find the node_modules state file - running an install might help. I tried looking for a solution, but didn't really find sth but placing node <command> directly into the ci. Maybe you want to take a look?

steffenkleinle: No, I think it is fine. Shouldn't happen often anyway :)

SteffiStuffel:
@sarahsporck Any progress here?

PR #335, opened 2022-08-01
LUN-384: Use pressable

f1sh1918:

### DIFF
--- a/src/components/AddElement.tsx
+++ b/src/components/AddElement.tsx
@@ -3,10 +3,11 @@ import { widthPercentageToDP as wp } from 'react-native-responsive-screen'
 import styled from 'styled-components/native'
 
 import { AddCircleIcon } from '../../assets/images'
+import PressableOpacity from './PressableOpacity'
 import { ContentSecondary } from './text/Content'
 import { SubheadingPrimary } from './text/Subheading'
 
-const Root = styled.TouchableOpacity`
+const Root = styled(PressableOpacity)`
   margin: ${props => props.theme.spacings.sm} 0;
 `
 
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -1,4 +1,4 @@
-import React, { ComponentType, ReactElement } from 'react'
+import React, { ComponentType, ReactElement, useState } from 'react'
 import { widthPercentageToDP as wp, heightPercentageToDP as hp } from 'react-native-responsive-screen'
 import { SvgProps } from 'react-native-svg'
 import styled, { css, useTheme } from 'styled-components/native'
@@ -17,7 +17,7 @@ interface ThemedLabelProps {
   color: string
 }
 
-const ThemedButton = styled.TouchableOpacity<ThemedButtonProps>`
+const ThemedButton = styled.Pressable<ThemedButtonProps>`
   ${props =>
     props.buttonTheme === BUTTONS_THEME.outlined &&
     !props.disabled &&
@@ -27,7 +27,6 @@ const ThemedButton = styled.TouchableOpacity<ThemedButtonProps>`
     `};
   flex-direction: row;
   padding: ${props => `${wp('3%')}px ${props.theme.spacings.sm}`};
-
   width: ${wp('70%')}px;
   align-items: center;
   border-radius: ${hp('7%')}px;
@@ -54,7 +53,7 @@ interface ButtonProps {
 }
 
 const Button = (props: ButtonProps): ReactElement => {
-  const [isPressed, setIsPressed] = React.useState(false)
+  const [isPressed, setIsPressed] = useState<boolean>(false)
   const { label, onPress, disabled = false, buttonTheme = BUTTONS_THEME.outlined } = props
   const theme = useTheme()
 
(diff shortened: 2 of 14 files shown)


src/components/PressableOpacity.tsx
```diff
@@ -0,0 +1,38 @@
+import React, { ReactElement, ReactNode, useState } from 'react'
+import { StyleProp, ViewStyle } from 'react-native'
+import styled from 'styled-components/native'
+
+interface Props {
+  children: ReactNode
+  onPress: () => void
+  style?: StyleProp<ViewStyle>
```
f1sh1918, line 8: required to apply styles on custom components with styled components f.e. `styled(PressableOpacity)`

src/components/PressableOpacity.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { ReactElement, ReactNode, useState } from 'react'
```
steffenkleinle, line 1: Didn't realize there is no touch effects anymore in Pressable compared to TouchableOpacity. Do you think this PR makes still sense then? Or should we rather just keep TouchableOpacity wherever we want that effect?

src/components/PressableOpacity.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { AccessibilityRole, StyleProp, ViewStyle } from 'react-native'
+import styled from 'styled-components/native'
+
+interface Props {
+  children: ReactNode
+  onPress: () => void
+  style?: StyleProp<ViewStyle>
+  testID?: string
+  accessibilityRole?: AccessibilityRole
```
steffenkleinle, line 10: I think we don't use accessibility props in lunes.

steffenkleinle (approved):
Tested on android

src/components/PressableOpacity.tsx
f1sh1918: maybe the we could use a testID. I just kept them for not failing the tests

src/components/PressableOpacity.tsx
f1sh1918: https://blog.logrocket.com/react-native-touchable-vs-pressable-components/
Well i think we can more standardize our components and even implement fancy animations if needed. Stuff like background-color change or opacity change can be implemented easily via styled components and is better readable and customizable for us. Thats my personal opinion

steffenkleinle: Sounds good, lets keep this!

src/routes/exercise-finished/ExerciseFinishedScreen.tsx
```diff
@@ -17,6 +17,7 @@ import {
   SadSmileyIcon,
 } from '../../../assets/images'
 import Button from '../../components/Button'
+import PressableOpacity from '../../components/PressableOpacity'
```
SteffiStuffel, line 20: When on exerciseFinishedScreen, and clicking the close-icon in the header it gets a background and looks weird, it should just change color of the text/icon, e.g. like the qr-code-scanner icon.

SteffiStuffel (changes requested):

src/components/AddElement.tsx
```diff
@@ -3,10 +3,11 @@ import { widthPercentageToDP as wp } from 'react-native-responsive-screen'
 import styled from 'styled-components/native'
 
 import { AddCircleIcon } from '../../assets/images'
+import PressableOpacity from './PressableOpacity'
 import { ContentSecondary } from './text/Content'
 import { SubheadingPrimary } from './text/Subheading'
 
-const Root = styled.TouchableOpacity`
+const Root = styled(PressableOpacity)`
```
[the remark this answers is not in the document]
f1sh1918: looks like `TouchableOpacity` Container also wrapped the explanation which shouldn't be the case. Fixed by last commit

src/components/Button.tsx
```diff
@@ -17,7 +17,7 @@ interface ThemedLabelProps {
   color: string
 }
 
-const ThemedButton = styled.TouchableOpacity<ThemedButtonProps>`
+const ThemedButton = styled.Pressable<ThemedButtonProps>`
```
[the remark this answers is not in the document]
f1sh1918: @steffenkleinle 
1. I checked the current version in app store and i think is not due to my changes. these lines were not touched which are responsible for bg color changes
```js
   if (isPressed) {
      return buttonTheme === BUTTONS_THEME.contained ? theme.colors.containedButtonSelected : theme.colors.placeholder
    }
    ```
    2. If i replace it by `Button` i have to override a lot of styles like padding, margin, justify-content and also padding for the label. I'm not sure. I don't think we should replace it with a regular `TextButton`

src/routes/exercise-finished/ExerciseFinishedScreen.tsx
f1sh1918: removed bg-color for `PressableOpacity` and it should be fine now. So bg-color changes only happen with `Button`component

src/components/Button.tsx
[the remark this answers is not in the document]
SteffiStuffel: Oke I see. I will open a ticket to fix this.

SteffiStuffel (approved):
Did a quick testing again. Works fine :tada:

PR #337, opened 2022-08-02
LUN-389: Upgrade tools dependencies

steffenkleinle:


f1sh1918 (approved):

tools/git-version.ts
```diff
@@ -77,28 +81,34 @@ const commitAndTag = async (
     content: contentBase64,
     branch,
     message: commitMessage,
+    // @ts-expect-error Random typescript error: property sha is not available on type { ..., sha: string, ... }
```
SteffiStuffel, line 84: Can you please explain?

SteffiStuffel (approved):
Have you tested this?

steffenkleinle:
> Have you tested this?

Not every command but some, yes.

tools/git-version.ts
steffenkleinle: The types just don't match for some reason. Didn't investigate again but did when I upgraded the dependency for integreat a while back and came to no conclusion. Feel free to try to make this work, would be nice if you'd find the problem.

PR #344, opened 2022-08-08
LUN-209: No internet connection

charludo:

### DIFF
--- a/src/components/ErrorMessage.tsx
+++ b/src/components/ErrorMessage.tsx
@@ -1,7 +1,8 @@
 import React from 'react'
-import { widthPercentageToDP as wp } from 'react-native-responsive-screen'
+import { heightPercentageToDP as hp, widthPercentageToDP as wp } from 'react-native-responsive-screen'
 import styled from 'styled-components/native'
 
+import { NoInternetConnectionIcon } from '../../assets/images'
 import { BUTTONS_THEME } from '../constants/data'
 import { NetworkError } from '../constants/endpoints'
 import labels from '../constants/labels.json'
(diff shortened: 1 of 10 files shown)


src/components/ErrorMessage.tsx
```diff
@@ -15,25 +16,71 @@ const Container = styled.View`
...
 interface ErrorMessageProps {
   error: Error | null
   refresh: () => void
+  contained?: boolean
 }
 
-const ErrorMessage = ({ error, refresh }: ErrorMessageProps): JSX.Element | null =>
-  error && (
+const ErrorMessage = ({ error, refresh, contained }: ErrorMessageProps): JSX.Element | null =>
+  error &&
```
steffenkleinle, line 61: Optional/personal preference: Imo it is much easier to read if you use brackets and if statements.

src/components/ErrorMessage.tsx
```diff
@@ -15,25 +16,71 @@ const Container = styled.View`
...
+  width: 140%;
+  height: 67%;
+  background-color: ${props => props.theme.colors.networkErrorBackground};
+  border-bottom-left-radius: ${hp('60%')}px;
+  border-bottom-right-radius: ${hp('60%')}px;
+  margin-bottom: ${props => props.theme.spacings.xxl};
+  justify-content: center;
+  align-items: center;
+  text-align: center;
 `
```
steffenkleinle, line 52: I think we should move this to a separate component that just takes a background color as props since this is (or should be) the same as `UpperSection` in `ExerciseFinishedScreen`.

steffenkleinle (approved):
@SteffiStuffel any reason why we use this design only for network errors? Could't we just show a different icon/message here as well for other errors?

src/components/ErrorMessage.tsx
charludo: I'm not 100% sure what you mean - the conditional render was already "there", I just amended the second part. Do you want me differentiate between the cases outside the return statement?

src/components/ErrorMessage.tsx
charludo: Color and height, yeah. I'll move it to its own component.

Does this need a test...?

steffenkleinle: No, styled components don't need tests. If we were to test css we'd spend 90% of the time with testing :D

steffenkleinle: Why do we need a different height here?

src/components/ErrorMessage.tsx
steffenkleinle: If I were to write this component from scratch I'd do something like
``` tsx
 const ErrorMessage = ({ error, refresh, contained }: ErrorMessageProps): JSX.Element | null => {
  if (!error) {
    return null
  }
  if (!contained && error.message === NetworkError) {
    return xyz
  }

  return asdf
}
``
But as I said, this is just personal preference because I find it easier to see the used conditions. So your choice if you just want to keep the current or change it.

charludo: I'll hold off on these changes until it's clear whether all errors get the new layout, then implement them.

src/components/ErrorMessage.tsx
charludo: Less of a reason of "need", more of a reason of "that's what the Zeplin designs look like" 😄

steffenkleinle: Okay. I'd prefer making it of consistent height if there is not a specific reason why we need another height. The only reason I could imagine why a different height would perhaps make sense is that the status bar/header is of a different color.

charludo: Just tested it, the usual height for this is 50%, which is just barely enough on my device. Increasing the height is a space issue on exercise-finished screens.

I'm assuming that on smaller devices, it might already overflow. And even if it doesnt, using 50% height leaves a lot of empty space on the screen and just plain doesn't look as good as the current Zeplin design.

Is there a problem with making the height adjustable? I'd just let the component take another optional prop and default to 50%.

SteffiStuffel:
> 

As we do not have designs for other errors, it is completely fine to use this also for other errors yes.

src/components/ErrorMessage.tsx
SteffiStuffel: No problem with adjustable height, especially if it is needed, because of the different content of the screens using it.

src/components/ErrorMessage.tsx
SteffiStuffel: All error can get the new layout.

SteffiStuffel (approved):
Tested on Android. Looks good.

src/components/ErrorMessage.tsx
charludo: Since there are no icons for the other errors, should the icon just be removed for other errors? Granted, the highlighted area does look a bit empty then.

steffenkleinle: Will be done in a separate issue :)

charludo: was the latest commit not needed then?

steffenkleinle: Ah I see, you already did it. Yeah then this is fine, thanks! So no additional issue is needed :)

PR #346, opened 2022-08-09
LUN-371: Rename CustomModal

charludo:

### DIFF
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -57,4 +57,4 @@ const Modal = (props: Props): JSX.Element => {
     </ModalSkeleton>
   )
 }
-eport default Modal
+export default Modal
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -15,7 +15,7 @@ const Message = styled(HeadingText)`
   text-align: center;
 `
 
-export interface CustomModalProps {
+export interface ModalProps {
   visible: boolean
   onClose: () => void
   text: string
@@ -28,7 +28,7 @@ export interface CustomModalProps {
 }
 
 // TODO Further adjustments gonna be done with LUN-312
-const CustomModal = (props: CustomModalProps): JSX.Element => {
+const Modal = (props: ModalProps): JSX.Element => {
   const {
     visible,
     text,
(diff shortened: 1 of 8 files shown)


charludo:
Just to be sure, since this is just internal and not something a user would ever know about, no release note should be created, correct?

SteffiStuffel:
> Just to be sure, since this is just internal and not something a user would ever know about, no release note should be created, correct?

correct :+1:

src/components/Modal.tsx
```diff
@@ -15,7 +15,7 @@ const Message = styled(HeadingText)`
   text-align: center;
 `
 
-export interface CustomModalProps {
+export interface ModalProps {
```
SteffiStuffel, line 18: (optional) we try to name prop types just `Props`, but as this is not consistent yet over the project, just fine if you keep this here, too. But just fyi.

SteffiStuffel (approved):
Not tested, looks good.

src/components/Modal.tsx
charludo: thanks for letting me know, changed it.

src/components/Modal.tsx
```diff
@@ -28,7 +28,7 @@ export interface CustomModalProps {
 }
 
 // TODO Further adjustments gonna be done with LUN-312
```
steffenkleinle, line 30: Optional: This comment is outdated and could be removed. Would be nice if you could remove it.

steffenkleinle (approved):
Tested on android

src/components/Modal.tsx
charludo: done

steffenkleinle: @charludo I think you didn't push it, did you?

charludo: oops, sorry. It is now.

src/components/Modal.tsx
```diff
@@ -58,4 +57,4 @@ const Modal = (props: Props): JSX.Element => {
     </ModalSkeleton>
   )
 }
-export default Modal
+eport default Modal
```
steffenkleinle: Typo
```suggestion
export default Modal
```

PR #348, opened 2022-08-15
LUN-394: Create user vocabulary screen

SteffiStuffel:


f1sh1918 (changes requested):
Looks nice on ios. good job :). just some small adjustments

src/routes/UserVocabularyOverviewScreen.tsx
```diff
@@ -0,0 +1,51 @@
...
+  margin: ${theme.spacings.xl};
+`
+
+interface PropsType {
+  navigation: StackNavigationProp<RoutesParams, 'UserVocabularyOverview'>
+}
+
+const UserVocabularyOverviewScreen = ({ navigation }: PropsType): JSX.Element => (
+  <Root>
+    <StyledHeading>{getLabels().ownVocabulary.myVocabulary}</StyledHeading>
```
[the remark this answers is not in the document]
SteffiStuffel: We do not really need it, but it makes automatic import possible, with just having the labels imported from json - at least for me - i always have to import it manually as intellij cannot do it automatically. So just makes my/our live easier.

SteffiStuffel:
> Looks nice on ios. good job :). just some small adjustments

Thanks for prompt review. I would appreciate, if we only "request changes", if something bigger needs to be done / something is buggy or stuff like this. For only small adjustments that need to be done, approval or comment makes us work more efficently, i guess.

f1sh1918 (approved):

src/services/helpers.ts
```diff
@@ -129,3 +129,5 @@ export const loadTrainingsSet = async (disciplineId: number): Promise<ServerResp
   const trainingSet = await getFromEndpoint<ServerResponseDiscipline>(trainingSetUrl)
   return trainingSet
 }
+
+export const getLabels = (): any => labels
```
steffenkleinle, line 133: Doesn't this remove type annotations? Intellij currently tells me whether a label exists or does not.

steffenkleinle (approved):
Tested, works fine :)

src/services/helpers.ts
SteffiStuffel: You are right, added appropriate type.

PR #349, opened 2022-08-16
LUN-397: Service for user created vocabulary

SteffiStuffel:


src/services/AsyncStorage.ts
```diff
@@ -140,6 +141,40 @@ export const getDevMode = async (): Promise<boolean | null> => {
...
+  }
+  userVocabulary[index] = newUserDocument
+  await setUserVocabulary(userVocabulary)
+  return true
+}
+
+const deleteUserDocument = async (userDocument: UserVocabularyDocument): Promise<void> => {
+  let userVocabulary = await getUserVocabulary()
+  userVocabulary = userVocabulary.filter(item => JSON.stringify(item) !== JSON.stringify(userDocument))
+  await setUserVocabulary(userVocabulary)
```
steffenkleinle, line 175: No need to use `let` here, either inline or create a separate var.
```suggestion
  const userVocabulary = await getUserVocabulary()
  const newUserVocabulary = userVocabulary.filter(item => JSON.stringify(item) !== JSON.stringify(userDocument))
  await setUserVocabulary(newUserVocabulary)
```

steffenkleinle (approved):
Not tested

f1sh1918 (approved):

PR #351, opened 2022-08-16
LUN-207: Added dictionary

SteffiStuffel:


src/navigation/screenOptions.tsx
```diff
@@ -8,6 +8,8 @@ import NavigationHeaderLeft from '../components/NavigationHeaderLeft'
 import { COLORS } from '../constants/theme/colors'
 import { RoutesParams } from './NavigationTypes'
 
+export const headerHeightPercentage = wp('15%')
```
steffenkleinle, line 11: Optional: If `useTabletHeaderHeight` is only ever used with this value it could just be used as default there, couldn't it? Or even converted to a constant instead of a function. Or am I missing something here?

src/routes/DictionaryScreen.tsx
```diff
@@ -0,0 +1,88 @@
...
+`
+
+interface Props {
+  navigation: StackNavigationProp<RoutesParams, 'DictionaryOverview'>
+}
+
+const DictionaryScreen = ({ navigation }: Props): ReactElement => {
+  const documents = useLoadAllDocuments()
+  const [searchString, setSearchString] = useState<string>('')
+  const filteredDocuments = documents.data?.filter(item => item.word.includes(searchString))
```
steffenkleinle, line 42: I think we should do something similar to integreat and normalize the query and the word.

src/routes/DictionaryScreen.tsx
```diff
@@ -0,0 +1,88 @@
...
+                    description={`${filteredDocuments?.length ?? 0} ${
+                      documents.data.length === 1 ? labels.general.word : labels.general.words
+                    }`}
+                  />
+                  <SearchBar value={searchString} setValue={setSearchString} />
+                </Header>
+              }
+              data={filteredDocuments}
+              renderItem={renderItem}
+              keyExtractor={item => `${item.id}`}
```
steffenkleinle, line 72: Optional: I'd prefer using the `key` react prop instead of passing this key extractor functions:
https://reactnative.dev/docs/flatlist#keyextractor
> By default, the list looks for a key prop on each item and uses that for the React key. Alternatively, you can provide a custom keyExtractor prop.

src/routes/DictionaryScreen.tsx
```diff
@@ -0,0 +1,88 @@
...
+          {documents.data && (
+            <FlatList
+              keyboardShouldPersistTaps='handled'
+              ListHeaderComponent={
+                <Header>
+                  <Title
+                    title={labels.general.dictionary}
+                    description={`${filteredDocuments?.length ?? 0} ${
+                      documents.data.length === 1 ? labels.general.word : labels.general.words
+                    }`}
```
steffenkleinle, line 65: Please extract to a `description` variable.

src/routes/DictionaryScreen.tsx
```diff
@@ -0,0 +1,88 @@
...
+                  />
+                  <SearchBar value={searchString} setValue={setSearchString} />
+                </Header>
+              }
+              data={filteredDocuments}
+              renderItem={renderItem}
+              keyExtractor={item => `${item.id}`}
+              showsVerticalScrollIndicator={false}
```
steffenkleinle, line 73: Why no scroll indicator?

src/components/__tests__/Modal.spec.tsx
```diff
@@ -22,13 +22,13 @@ describe('Modal', () => {
   it('should display passed props', () => {
     const { getByText } = render(<Modal {...defaultProps} />)
     expect(getByText('Are you sure?')).toBeDefined()
-    expect(getByText('Zurück')).toBeDefined()
+    expect(getByText('zurück')).toBeDefined()
```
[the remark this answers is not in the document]
SteffiStuffel: hmh, yes I did it on purpose, but I think you are right, most other labels also start with a capital, so I keep it like this here, too.

src/constants/labels.json
```diff
@@ -155,10 +155,13 @@
...
     "disciplines": "Module",
     "word": "Wort",
-    "words": "Wörter",
-    "customModalCancel": "Zurück"
+    "words": "Wörter"
   },
   "favorites": "Favoriten",
+  "dictionary": {
+    "enterWord": "Wort eingeben",
+    "noResults": "Sorry, keine Ergebnisse gefunden"
```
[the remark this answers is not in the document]
SteffiStuffel: Yes, I also think this, but just did it as in the designs, will change as you suggested.

src/navigation/screenOptions.tsx
SteffiStuffel: Yes I thought about this too, but we cannot use a hook here and therefore I did it like this.

src/routes/DictionaryScreen.tsx
SteffiStuffel: Good point, thank you. Added it.

src/routes/DictionaryScreen.tsx
SteffiStuffel: hmh, but here it really makes sense, to tell React using the id as a key is good.

src/routes/DictionaryScreen.tsx
SteffiStuffel: VocabularyList also does not have one, so I did it the same way. But just check out how it looks like with scroll indicator, and it is a little messy, as the scroll indicator changes the size while scrolling as the list is getting "longer" while scrolling, as elements get rendered while scrolling.

f1sh1918 (approved):
Tested on ios.Works fine except of InitCap articles (check comment)

src/routes/DictionaryScreen.tsx
steffenkleinle: Thanks for the explanation, its fine the way it is :)

src/routes/DictionaryScreen.tsx
steffenkleinle: yeah, but you tell react the same by doing `<MyElement key={id} />` (which is way cleaner and less code imo, e.g. no need to convert it to a string).

src/routes/DictionaryScreen.tsx
```diff
@@ -33,13 +34,23 @@ const Header = styled.View`
...
   const documents = useLoadAllDocuments()
   const [searchString, setSearchString] = useState<string>('')
-  const filteredDocuments = documents.data?.filter(item => item.word.includes(searchString))
+  const searchStringWithoutArticle = ARTICLES.map(article => article.value).includes(
+    searchString.split(' ')[0].toLowerCase()
+  )
+    ? searchString.substring(searchString.indexOf(' ') + 1)
+    : searchString
+  const filteredDocuments = documents.data?.filter(item =>
+    item.word.toLowerCase().includes(searchStringWithoutArticle.toLowerCase().trim())
```
steffenkleinle, line 49: Did you forget the normalization or is this on purpose? I think searching for e.g `Fon` should return `der Fön`.

src/routes/DictionaryScreen.tsx
```diff
@@ -33,13 +34,23 @@ const Header = styled.View`
...
 
 const DictionaryScreen = ({ navigation }: Props): ReactElement => {
   const documents = useLoadAllDocuments()
   const [searchString, setSearchString] = useState<string>('')
-  const filteredDocuments = documents.data?.filter(item => item.word.includes(searchString))
+  const searchStringWithoutArticle = ARTICLES.map(article => article.value).includes(
+    searchString.split(' ')[0].toLowerCase()
+  )
+    ? searchString.substring(searchString.indexOf(' ') + 1)
+    : searchString
```
steffenkleinle, line 47: This probably breaks for e.g. `Fön der` or just `der`. Tbh I just keep all of the query and not try to exclude articles since there is not really a need for this (users can either type the right articles or just omit the article in the search which should be the standard).

src/routes/DictionaryScreen.tsx
```diff
@@ -58,16 +69,11 @@ const DictionaryScreen = ({ navigation }: Props): ReactElement => {
...
-                      documents.data.length === 1 ? labels.general.word : labels.general.words
-                    }`}
-                  />
-                  <SearchBar value={searchString} setValue={setSearchString} />
+                  <Title title={labels.general.dictionary} description={description} />
+                  <SearchBar query={searchString} setQuery={setSearchString} />
                 </Header>
               }
-              data={filteredDocuments}
+              data={filteredDocuments?.sort((a, b) => a.word.localeCompare(b.word))}
```
steffenkleinle, line 76: Optional: Move this out of here, should be on top of the file. E.g.:
```ts
const sortedDocuments = filteredDocuments?.sort((a, b) => a.word.localeCompare(b.word))
```

src/navigation/screenOptions.tsx
```diff
@@ -8,6 +8,8 @@ import NavigationHeaderLeft from '../components/NavigationHeaderLeft'
...
+
 const headerStyles = (headerHeight?: number) =>
```
steffenkleinle, line 13:
```suggestion
const headerStyles = (headerHeight: number = wp('15%')) =>
```

steffenkleinle (commented):
Great work :tada: 

There is the following warning, not sure if there is anything to do though:
>  VirtualizedList: You have a large list that is slow to update - make sure your renderItem function renders components that follow React performance best practices like PureComponent, shouldComponentUpdate, etc.

src/navigation/screenOptions.tsx
steffenkleinle: See suggestion below on what I meant :) You can use that or just keep the current.

src/routes/DictionaryScreen.tsx
SteffiStuffel: You are right. Thanks for clarification. Changed it.

src/routes/DictionaryScreen.tsx
SteffiStuffel: Do not really see the need, but added it.

src/routes/DictionaryScreen.tsx
SteffiStuffel: When we want to search for a word with article, we would need to concat article + word of all documents (plus alternatives of documents), which is expensive / takes long. 
Just cutting the article is faster and therefore better.

steffenkleinle: Does this work if just `die` is entered because e.g. searching for `Dietrich`?

SteffiStuffel: Yes it should work, as we only cut the "die", if the search string contains of more than one word and the first word is "die".

Tested, works as expected.

steffenkleinle (approved):
Not tested again

PR #356, opened 2022-08-23
LUN-355: Small cleanup

charludo:


SteffiStuffel (approved):

src/components/FeedbackModal.tsx
```diff
@@ -20,7 +19,7 @@ interface FeedbackModalProps {
   feedbackForId: number
 }
 
-const FeedbackModal: React.FC<FeedbackModalProps> = ({
+const FeedbackModal: FC<FeedbackModalProps> = ({
```
f1sh1918, line 22: For consistency reason i think we should remove the type here since we never used in other components
`: FC<FeedbackModalProps>` is not needed

f1sh1918 (approved):
nice refactoring :) just one small adjustment!

src/components/CheatMode.tsx
```diff
@@ -21,18 +21,18 @@ interface CheatModeProps {
...
 const CheatMode = ({ cheat }: CheatModeProps): ReactElement => {
-  const { data: cheatsEnabled } = useLoadAsync(getDevMode, null)
+  const { data: cheatsEnabled } = useLoadAsync(AsyncStorage.getDevMode, null)
   return (
     <>
       {cheatsEnabled && (
         <CheatContainer>
           <Button
-            label={labels.exercises.cheat.succeed}
+            label={getLabels().exercises.cheat.succeed}
```
steffenkleinle, line 30: @SteffiStuffel why do we need a function for this btw? Just tried this and it seems to work as well (and would be way easier syntax to use it).
``` ts
export const labels: typeof importedLabels = importedLabels
```

steffenkleinle (approved):

src/components/CheatMode.tsx
charludo: I imagine this leaves the opportunity to do some post-processing on the labels later on, should the need arise? Should I merge this for now, or revert back to `labels.`?

SteffiStuffel: Steffens solution is also fine for me. @charludo you can choose which one you prefer more.

charludo: @SteffiStuffel @steffenkleinle in that case I'll leave it as is. Swapping `getLabels().` for `labels.` in the future is easy enough.

PR #357, opened 2022-08-24
LUN-387: Add option to delete profession on error

charludo:


src/routes/home/components/DisciplineCard.tsx
```diff
@@ -63,9 +63,12 @@ const DisciplineCard = ({
   }
 
   if (!discipline) {
+    if (error?.message !== ForbiddenError && error?.message !== NetworkError) {
+      return null
+    }
     return (
       <Card>
-        {error?.message === ForbiddenError && isTypeLoadProtected(identifier) ? (
+        {error.message === ForbiddenError && isTypeLoadProtected(identifier) ? (
```
SteffiStuffel, line 71: I think it would be better if we show this part (with the button to delete the discipline) for all errors except Network error. Otherwise users may be confused, why a discipline suddenly has disappeard.

src/routes/manage-selections/components/SelectionItem.tsx
```diff
@@ -62,7 +61,7 @@ const SelectionItem = ({ identifier, deleteItem }: PropsType): JSX.Element => {
         <Explanation>{labels.manageSelection.deleteModal.explanation}</Explanation>
       </Modal>
       <ListItem
-        title={data.title}
+        title={data?.title ?? labels.general.error.unknown}
```
SteffiStuffel, line 64: I think it would really make sense to also show this ListItem with the delete-icon on `Forbidden`-error

src/routes/home/components/DisciplineCard.tsx
```diff
@@ -63,9 +63,12 @@ const DisciplineCard = ({
   }
 
   if (!discipline) {
+    if (error?.message !== NetworkError) {
```
SteffiStuffel, line 66: So still for invalid api key (e.g. revoked) there is just displayed nothing, leaving users wondering why there custom discipline just has disappeared.

You can test for this error, if you create an api key in test cms, add i to the app and revoke the api key in the test cms afterwards.

SteffiStuffel (changes requested):
For network errors the arrow is still displayed in ManageSelection screen

src/routes/home/components/DisciplineCard.tsx
charludo: Ah, I think I misunderstood your previous comment.

Instead of hiding the card on forbidden/unknown error, the card should be empty save for an error message and option to delete the card, same as in the manage selection screen?

SteffiStuffel: Oke, sorry about unclear previous comment.

Correct, on forbidden and unknown error, there should be displayed an error message and a button for deleting the custom discipline (remove api-key from storage).

Not sure how we want to handle unknown error for lunes standard discipline, I think here it would also make sense to add option/button to remove profession from selected professions

charludo: > For network errors the arrow is still displayed in ManageSelection screen

Should the arrow be removed, or replaced by the red X? For Forbidden and Unknown the option to remove makes sense, not sure about NetworkError.

SteffiStuffel: Arrow should be removed and not be replace, because of the reasons you mentioned :+1:

charludo: Done.

Wuick question, does the new DeletionModal component need a test suite? Since it's just a wrapper around Modal.

src/routes/manage-selections/components/SelectionItem.tsx
```diff
@@ -41,28 +36,20 @@ const SelectionItem = ({ identifier, deleteItem }: PropsType): JSX.Element => {
...
-  if (!data) {
-    let errorMessage = getLabels().general.error.unknown
-    if (error?.message === ForbiddenError) {
-      errorMessage = getLabels().home.errorLoadCustomDiscipline
-    } else if (error?.message === NetworkError) {
-      errorMessage = `${getLabels().general.error.noWifi} (${error.message})`
-    }
-    return <ListItem title={errorMessage} />
+  if (!data && error?.message === NetworkError) {
+    return <ListItem title={`${getLabels().general.error.noWifi} (${error.message})`} hideRightChildren />
```
SteffiStuffel, line 40: (optional) We can also just pass empty tags here `<></>` for the prop rightChildren, so we can get rid of this additional hideRightChildren prop.

src/routes/home/components/DisciplineCard.tsx
```diff
@@ -63,25 +68,43 @@ const DisciplineCard = ({
...
+    }
+
+    let deleteItem
+    let errorMessage
+    if (error?.message === ForbiddenError && isTypeLoadProtected(identifier)) {
+      deleteItem = () => AsyncStorage.removeCustomDiscipline(identifier.apiKey).then(refreshHome)
+      errorMessage = `${getLabels().home.errorLoadCustomDiscipline} ${identifier.apiKey}`
+    } else {
+      deleteItem =
+        'disciplineId' in identifier
```
SteffiStuffel, line 86: You can just use `!isTypeLoadProtected(identifier)` here, can't you?

SteffiStuffel (approved):
Tested. Works fine :tada:

src/routes/manage-selections/components/SelectionItem.tsx
steffenkleinle: Shouldn't rather just nothing be passed? I.e. the prop should be optional I think.

charludo: @SteffiStuffel that was my first thought as well, either `<></>` or just `null`. However, eslint complains (`LUN-387-add-option-to-delete-profession-on-error`). 

If you want me to, I'll disable the error for that line, I simply thought this was the cleaner solution.

src/routes/home/components/DisciplineCard.tsx
charludo: I'm not sure quite why, but reordering the conditions to use `!isTypeLoadProtected(identifier)` fails the test for unknown errors.

More importantly though, I'd argue that unknown error should be the "fallback" if no other error matches, so explicitly testing if the error is a ForbiddenError makes sense imo.

charludo: Or maybe I am misunderstanding, do you want to use `!isTypeLoadProtected(identifier)` on line 84 of your quotation, or as a replacement for 85/86?

src/components/ErrorMessage.tsx
```diff
@@ -25,7 +25,7 @@ const ErrorTitle = styled.Text`
   color: ${props => props.theme.colors.primary};
   text-align: center;
 `
-const ErrorText = styled.Text<{ centered?: boolean }>`
+export const ErrorText = styled.Text<{ centered?: boolean }>`
```
f1sh1918, line 28:
```suggestion
export const ErrorText = styled(Content)<{ centered?: boolean }>`
```
you could use existing styled text components and remove font-size and font-family in styles here to be convenient with other Text components

src/components/DeletionModal.tsx
```diff
@@ -0,0 +1,30 @@
...
+const Explanation = styled(ContentSecondary)`
+  padding: 0 ${props => props.theme.spacings.lg} ${props => props.theme.spacings.xl};
+  text-align: center;
+`
+
+const DeletionModal = ({ isVisible, onConfirm, onClose }: PropsType): JSX.Element => (
+  <Modal
+    visible={isVisible}
+    onClose={onClose}
+    text={getLabels().manageSelection.deleteModal.confirmationQuestion}
```
f1sh1918, line 23: use object destructuring like `const {confirmationQuestions, ....} = getLabels().manageSelection.deleteModal`

src/routes/home/components/DisciplineCard.tsx
SteffiStuffel: I thought as a replacement for ` 'disciplineId' in identifier` check.

src/routes/manage-selections/components/SelectionItem.tsx
f1sh1918: better to disable the error than to add unneeded code imo :)

f1sh1918 (approved):
nice work, just some minor improvements

src/routes/home/components/DisciplineCard.tsx
charludo: All right, that appears to work! Pushed the change.

src/components/ErrorMessage.tsx
charludo: Thanks for the suggestion, that definitely makes sense.

src/components/DeletionModal.tsx
charludo: Makes sense. Didn't notice when extracting the component from SelectionItem, but fixed it now. Thanks.

src/routes/manage-selections/components/SelectionItem.tsx
charludo: Getting conflicting messages here 😄 

Is it really unnecessary code if the alternative is actively suppressing an error? Plus, `ListItem` is currently being used in a couple of different contexts, I can imagine there are other situations where we would not want to have an icon.

Though having none by default might make more sense.

SteffiStuffel: You are free to choose, which solution you prefer. Both is fine.

PR #358, opened 2022-08-24
LUN-413: Save score on exercise finish

SteffiStuffel:


src/services/AsyncStorage.ts
```diff
@@ -84,12 +84,38 @@ const setExerciseProgress = async (disciplineId: number, exerciseKey: ExerciseKe
   await AsyncStorage.setItem(PROGRESS_KEY, JSON.stringify(savedProgress))
 }
 
+const calculateScore = (documentsWithResults: DocumentResult[]): number => {
```
steffenkleinle, line 87: Not sure if it is the right place here to do something like this. Imo `AsyncStorage` should be just about CRUD operations to the async storage. Not about calculating scores and similar.

src/services/AsyncStorage.ts
```diff
@@ -84,12 +84,38 @@ const setExerciseProgress = async (disciplineId: number, exerciseKey: ExerciseKe
...
+            break
+          case 2:
+            acc += SCORE_SECOND_TRY
+            break
+          case 3:
+            acc += SCORE_THIRD_TRY
+            break
+        }
+        /* eslint-enable no-param-reassign */
+        return acc
```
steffenkleinle, line 108:
```suggestion
        let score = acc
        switch (document.numberOfTries) {
          case 1:
            score += SCORE_FIRST_TRY
            break
          case 2:
            score += SCORE_SECOND_TRY
            break
          case 3:
            score += SCORE_THIRD_TRY
            break
        }
        return score
```

steffenkleinle (approved):

src/services/AsyncStorage.ts
SteffiStuffel: moved to helpers

f1sh1918 (approved):
tested on ios. calculation works fine! nice work :)

PR #362, opened 2022-08-30
LUN-404: fix buttons on favorite detail screen

charludo:


SteffiStuffel (approved):
(minor) The icon of the back button is still the X and not the back-arrow, but I think this is fine as in the long term, we may change the behaviour of the favorites-detail view to be similar to the Dictionary-detail view.

charludo:
> (minor) The icon of the back button is still the X and not the back-arrow, but I think this is fine as in the long term, we may change the behaviour of the favorites-detail view to be similar to the Dictionary-detail view.

Hm, right. Looks like `NavigationHeaderLeft` has a prop to toggle that. Will do.

f1sh1918 (approved):
Tested on ios. nice work 👍

PR #364, opened 2022-08-31
LUN-415: Improve keyboard handling feedback modal

f1sh1918:


steffenkleinle (changes requested):
Not working for me:

f1sh1918:
> Not working for me: 

looks like behavior on android differs from ios :/ gonna check that

f1sh1918:
Updated: Now i implemented a different handling for android and ios. please test on android again

src/hooks/useKeyboard.ts
```diff
@@ -1,14 +1,14 @@
 import { useEffect, useState } from 'react'
 import { Keyboard, KeyboardEvent } from 'react-native'
 
-// This hook listens to the keyboard and returns boolean whether keyboard is opened or closed
+// This hook listens to the keyboard and returns boolean whether keyboard is opened or closed & keyBoardHeight
```
steffenkleinle, line 4: Optional: I think it is pretty self explanatory so I'd remove the comment
```suggestion
```

SteffiStuffel (commented):
On Android it is jumping twice, looking weird (see attached video). Also the "send" button disappears. I don't think it is intuitive, that you need to scroll.

https://user-images.githubusercontent.com/18463686/190106511-88b3bafb-8625-4044-8efb-84de00a53311.mp4

f1sh1918:
> On Android it is jumping twice, looking weird (see attached video). Also the "send" button disappears. I don't think it is intuitive, that you need to scroll.
> 
>  LUN-415.mp4

Well i dunno what to do against the jumping to be honest. Since this modal is very flexible you will reach the point where you have to scroll. We can not just resize it. So feel free for other suggestions than scrolling

SteffiStuffel (approved):
Not really happy with the current solution because of the jumping, but it is better than the status quo, so fine for me if it gets merged. Maybe you can open a ticket for fixing the jumping or we wait for UI/UX or testing to complain about ^^

f1sh1918:
> Not really happy with the current solution because of the jumping, but it is better than the status quo, so fine for me if it gets merged. Maybe you can open a ticket for fixing the jumping or we wait for UI/UX or testing to complain about ^^

@SteffiStuffel 
Well the jumping issue is related to the resizing of the height that happens after the keyboard was opened. But we cannot resize the height before because we do not know the height of the keyboard before it did show.
We can only expect a particular keyboard height and calculate the Modal container height using it.
Then use this height even if keyboard is opened or closed

SteffiStuffel:
> 

I would appreciate most, if we just block rerendering until everything is calculated and then render it (maybe useLayoutEffect will help here), as the jumping is cause by first updating the padding and in a second step updating the height.

f1sh1918:
> > 
> 
> I would appreciate most, if we just block rerendering until everything is calculated and then render it (maybe useLayoutEffect will help here), as the jumping is cause by first updating the padding and in a second step updating the height.

fixed the bottom positioning jumping but can not get the recalc height work before updating/rendering component. you can give it a try if you want. its not just by adding `useLayoutEffect`

charludo (approved):
(Tested on Android Emulator)
Very nice, works as intended. Never lost track of the cursor, and all fields + send button accessible without closing the keyboard.

Minor nitpick: at least on my rather low resolution emulator screen, the margins around the modal could be smaller to show a bit more of its content, which could be useful in some cases. For example, the bottom of the email field is just barely cut off after opening the keyboard

PR #370, opened 2022-09-21
LUN-407: Add evaluation of levels

charludo:

Small note about current behaviour: using devmode to "fail all" doesn't mark an exercise as done, so not negative feedback badge is shown.


src/components/ListItem.tsx
```diff
@@ -160,17 +188,21 @@ const ListItem = ({
...
-      <FlexContainer>
-        {titleToRender}
-        <DescriptionContainer>
-          {badgeLabel && <BadgeLabel pressed={pressed}>{badgeLabel}</BadgeLabel>}
-          {description && description.length > 0 && <Description pressed={pressed}>{description}</Description>}
-        </DescriptionContainer>
-        {children}
-      </FlexContainer>
-      {rightChildrenToRender}
+      <FeedbackBadge levelIdentifier={levelIdentifier} setFeedback={setFeedback} />
```
SteffiStuffel, line 193: I think this breaks other ListItems, e.g. the icons in the Vocabulary list, maybe it would make sense, to have a prop if feedback should be displayed, because currently it is always displayed ig a levelIdentifier is passed, or?

lun-407

src/routes/exercises/ExercisesScreen.tsx
```diff
@@ -90,9 +89,9 @@ const ExercisesScreen = ({ route, navigation }: ExercisesScreenProps): JSX.Eleme
           title={item.title}
           description={item.description}
           onPress={() => handleNavigation(item)}
-          arrowDisabled={nextExercise === null || item.level > nextExercise.level}>
-          <Trophy level={item.level} />
-        </ListItem>
+          arrowDisabled={nextExercise === null || item.level > nextExercise.level}
```
SteffiStuffel, line 92: Maybe it would be better to read Progress here and pass the results to the listItems?

src/routes/exercises/ExercisesScreen.tsx
```diff
@@ -10,7 +10,6 @@ import Modal from '../../components/Modal'
 import RouteWrapper from '../../components/RouteWrapper'
 import ServerResponseHandler from '../../components/ServerResponseHandler'
 import Title from '../../components/Title'
-import Trophy from '../../components/Trophy'
```
SteffiStuffel, line 13: Trophy icon can be removed from assets.

src/components/ListItem.tsx
SteffiStuffel: Maybe it would be better to have Something like a ListItemWithFeedback or something, where the List Item is used inside?

src/components/FeedbackBadge.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <ThumbsUpIcon width={wp('6%')} height={wp('6%')} />
+        <BadgeText>{getLabels().exercises.feedback.positive}</BadgeText>
+      </BadgeContainer>
+    )
+  }
+
+  if (feedback === FEEDBACK.NEGATIVE) {
+    return (
+      <BadgeContainer testID='negative-badge'>
+        <ThumbsDownIcon width={wp('6%')} height={wp('6%')} />
```
SteffiStuffel, line 72: We changed to use `hp` instead of `wp` for better tablet support.

SteffiStuffel (commented):
The alignment of the locking lane is not really good (centered) now, but if there is no good solution to solve this, fine for me:
lun-407-height

src/routes/dictionary/__tests__/DictionaryScreen.spec.tsx
```diff
@@ -10,6 +10,7 @@ import { getReturnOf } from '../../../testing/helper'
 import render from '../../../testing/render'
 import DictionaryScreen from '../DictionaryScreen'
 
+jest.mock('@react-navigation/native')
```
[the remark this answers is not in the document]
charludo: Omitting them fails all tests in the respective testsuite because FeedbackBadge relies on `useIsFocused` which requires the component to be wrapped in a `NavigationContainer`. That's always the case in the actual app, but not inside these unit tests.

The same thing has been done in a number of other files already, and it appears to be the preferred solution to the problem.

src/components/ListItem.tsx
charludo: > Maybe it would be better to have Something like a ListItemWithFeedback or something, where the List Item is used inside?

Good idea. Will do.

src/routes/dictionary/__tests__/DictionaryScreen.spec.tsx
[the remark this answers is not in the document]
SteffiStuffel: I see, thanks for explanation.

src/components/ListItem.tsx
charludo: @SteffiStuffel I just had another look at this, the misalignment wasn't from the badge itself, but rather from a `align-items: center` that got lost somewhere along the way. Should be fixed now.

The `levelIdentifier` also only gets passed from `ExerciseScreen` and not passing it means `FeedbackBadge` returns `null`, so your first suggestion is already implemented.

Please have another look at if the current state is satisfactory.

(A separate component *would* be nice, I agree, but wouldn't really reduce the complexity in the `ListItem` component, since it still needs to react to state updates from the badge (mainly setting the border styles))

src/components/__tests__/FeedbackBadge.spec.tsx
```diff
@@ -0,0 +1,94 @@
...
+describe('FeedbackBadge', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  const renderFeedbackBadge = (levelIdentifier: { disciplineId: number; level: number } | null) =>
+    render(<FeedbackBadge levelIdentifier={levelIdentifier} setFeedback={mockSetFeedback} />)
+
+  it('should not show when no level info is provided', async () => {
+    mocked(useLoadAsync).mockImplementation(() => ({
```
SteffiStuffel, line 27: you can use "mockData" function from `mockUseLoadFromEndpoint.ts`
Same for other tests in this file.

SteffiStuffel (approved):

src/routes/exercises/ExercisesScreen.tsx
charludo: @SteffiStuffel just moved the feedback logic to `ExerciseScreen`. Good call of yours, this simplified the ListItem and FeedbackBadge. 

Sorry to request a re-review, but I'd be great if you could have a look at it again.

f1sh1918 (changes requested):
Works nice! Good work.
There are some design adjustments which especially for bigger screens are mandatory

The List Item element height (for Exercises) shouldn't increase that much. Looks quite weird if you have elements with and without Feedback Badge. According to the design they should be equal. Probably you have to return an empty view placeholder instead of null or sth similar to keep them equal

image

charludo:
> There are some design adjustments which especially for bigger screens are mandatory
> 
> The List Item element height (for Exercises) shouldn't increase that much. Looks quite weird if you have elements with and without Feedback Badge. 

Have a look again, the badge should stay smaller on larger screens now, and the ListItem has reduced padding when the badge is shown. Not quite equal overall height with/without the badge (see below).

> According to the design they should be equal. Probably you have to return an empty view placeholder instead of null or sth similar to keep them equal

Not sure if increasing the ListItem size everywhere is worth  this though?

f1sh1918:
> > There are some design adjustments which especially for bigger screens are mandatory
> > The List Item element height (for Exercises) shouldn't increase that much. Looks quite weird if you have elements with and without Feedback Badge.
> 
> Have a look again, the badge should stay smaller on larger screens now, and the ListItem has reduced padding when the badge is shown. Not quite equal overall height with/without the badge (see below).
> 
> > According to the design they should be equal. Probably you have to return an empty view placeholder instead of null or sth similar to keep them equal
> 
> Not sure if increasing the ListItem size everywhere is worth this though?

The main problem is, that the `FeedbackBadge` is not inside of the ContentContainer (where `min-height` is set, so the container increases not equally because it does not include your new component. Find my suggestions in comment to fix this. But check for side effects :)

src/components/ListItem.tsx
```diff
@@ -4,28 +4,60 @@ import { heightPercentageToDP as hp, widthPercentageToDP as wp } from 'react-nat
...
-  border: 1px solid ${prop => (prop.pressed ? prop.theme.colors.primary : prop.theme.colors.disabled)};
+`
+
+const ContentContainer = styled.View<{ pressed: boolean; disabled: boolean; feedback: EXERCISE_FEEDBACK }>`
+  min-height: ${hp('12%')}px;
+  display: flex;
+  flex-direction: row;
   padding: ${props =>
-    `${props.theme.spacings.sm} ${props.theme.spacings.xs} ${props.theme.spacings.sm} ${props.theme.spacings.sm}`};
+    props.feedback !== EXERCISE_FEEDBACK.NONE
```
f1sh1918, line 57: remove conditional padding with
`padding: ${props =>
          `${props.theme.spacings.sm} ${props.theme.spacings.xs} ${props.theme.spacings.sm} ${props.theme.spacings.sm}`};
  `

src/components/ListItem.tsx
```diff
@@ -4,28 +4,60 @@ import { heightPercentageToDP as hp, widthPercentageToDP as wp } from 'react-nat
...
   background-color: ${prop => {
     if (prop.disabled) {
       return prop.theme.colors.disabled
     }
     return prop.pressed ? prop.theme.colors.primary : prop.theme.colors.backgroundAccent
   }};
-  border: 1px solid ${prop => (prop.pressed ? prop.theme.colors.primary : prop.theme.colors.disabled)};
+`
+
+const ContentContainer = styled.View<{ pressed: boolean; disabled: boolean; feedback: EXERCISE_FEEDBACK }>`
```
f1sh1918, line 52: remove unneeded `feedback` prop

src/components/ListItem.tsx
```diff
@@ -160,17 +194,21 @@ const ListItem = ({
...
-        {titleToRender}
-        <DescriptionContainer>
-          {badgeLabel && <BadgeLabel pressed={pressed}>{badgeLabel}</BadgeLabel>}
-          {description && description.length > 0 && <Description pressed={pressed}>{description}</Description>}
-        </DescriptionContainer>
-        {children}
-      </FlexContainer>
-      {rightChildrenToRender}
+      <FeedbackBadge feedback={feedback} />
+      <ContentContainer pressed={pressed} disabled={disabled} feedback={feedback}>
```
f1sh1918, line 200: remove `feedback`prop

charludo:
@f1sh1918 makes sense. I pushed your suggestions. Doesn't appear to break anything, though I also can't really see the difference on my small screen (ListItems apparently are always greater then 12% in height at my resolution) - but I agree, looks much better in the screenshot you commented.

PR #374, opened 2022-09-27
LUN-432: Change console log linting rule to error

charludo:


src/components/AudioPlayer.tsx
```diff
@@ -54,8 +54,10 @@ const AudioPlayer = ({ document, disabled, submittedAlternative }: AudioPlayerPr
         }
       })
       .catch(async (error: TtsError) => {
+        /* eslint-disable-next-line no-console */
```
f1sh1918, line 57: is there a reason why not to use `reportError` fkt here?

src/components/AudioPlayer.tsx
```diff
@@ -54,8 +54,10 @@ const AudioPlayer = ({ document, disabled, submittedAlternative }: AudioPlayerPr
...
         console.error(`Tts-Error: ${error.code}`)
         if (error.code === 'no_engine') {
+          /* eslint-disable-next-line no-console */
           await Tts.requestInstallEngine().catch(e => console.error('Failed to install tts engine: ', e))
```
f1sh1918, line 61: same as above

src/services/AsyncStorage.ts
```diff
@@ -172,7 +172,6 @@ const editUserDocument = async (oldUserDocument: Document, newUserDocument: Docu
 }
 
 const deleteUserDocument = async (userDocument: Document): Promise<void> => {
-  console.log('hi')
```
steffenkleinle, line 175: :smile:

steffenkleinle (approved):

src/components/AudioPlayer.tsx
charludo: Wasn't sure this needed reporting. At least on my devices, this happens constantly. Apparently the original implementer also didn't see the value in logging the actual error, so I'm assuming this is only being logged to satisfy `typescript-eslint/no-floating-promises`

SteffiStuffel (approved):
Looks good :tada:

PR #381, opened 2022-10-11
LUN-417: Remove sentence when solution is shown

lunars97:


SteffiStuffel (changes requested):
You did this a little bit more complicated that it needs to be.

Just use the "submission"-Prop of `Feedback` to check if the "Das war leider nicht richtig"-sentence should be displayed.

You can check in the `else`-case, if this prop is not null, and just adjust the message. 

Currently in the else case message is set like this: 
```   
message = needsToBeRepeated
      ? `${getLabels().exercises.write.feedback.wrong}\n${wrongWithCorrectSolution}`
      : wrongWithCorrectSolution
``` 
can be replaced with
```   
message = submission
      ? wrongWithCorrectSolution
      : `${getLabels().exercises.write.feedback.wrong}\n${wrongWithCorrectSolution}`
``` 

Then you can undo the changes in all the other files.

Please adjust tests in Feedback.spec.tsx for this. 
The test `'should render finally incorrect feedback'` and `'should render incorrect feedback with retries not exceeded'` need to be adjusted. 
We use react testing library. Maybe you want to read a bit about it: https://testing-library.com/docs/react-testing-library/intro/

src/routes/write-exercise/components/Feedback.tsx
```diff
@@ -66,9 +70,7 @@ const Feedback = ({ documentWithResult, submission, needsToBeRepeated }: Feedbac
   return (
     <Background source={background} testID='background-image'>
       <Icon width={28} height={28} />
-      <StyledText numberOfLines={3} ellipsizeMode='tail'>
-        {message}
-      </StyledText>
+      <StyledText>{message}</StyledText>
```
SteffiStuffel, line 73: These changes are from your other branch LUN-418 and should not appear here. This happens if you do `git checkout -b LUN-417...` when being on LUN-418 branch. You always need to do `git checkout main` before checking out a new branch.
To fix this you normally need to change to base branch see here: https://stackoverflow.com/questions/10853935/change-branch-base
But in this case, you can just merge your LUN-418 branch and it is just fine.

src/routes/write-exercise/components/InteractionSection.tsx
```diff
@@ -48,15 +47,14 @@ const InteractionSection = (props: InteractionSectionProps): ReactElement => {
...
-
   const textInputRef = useRef<View>(null)
 
   useEffect(() => {
     if (!isAnswerSubmitted) {
       setInput('')
+      setSubmittedInput('')
     }
   }, [isAnswerSubmitted])
-
```
SteffiStuffel, line 59: This newline should stick, as it is between to new functions and makes the file better readable.

src/routes/write-exercise/components/InteractionSection.tsx
```diff
@@ -69,6 +67,7 @@ const InteractionSection = (props: InteractionSectionProps): ReactElement => {
     if (validAnswers.some(answer => stringSimilarity.compareTwoStrings(answer.word, word) > almostCorrectThreshold)) {
       return 'similar'
     }
+
```
SteffiStuffel, line 70: Please remove this newline

SteffiStuffel (approved):
Tested on Android. Works as expected :tada:

src/routes/write-exercise/components/InteractionSection.tsx
```diff
@@ -39,7 +39,6 @@ const InteractionSection = (props: InteractionSectionProps): ReactElement => {
   const [isArticleMissing, setIsArticleMissing] = useState<boolean>(false)
   const [input, setInput] = useState<string>('')
   const [submittedInput, setSubmittedInput] = useState<string | null>(null)
-
```
steffenkleinle, line 42: In general its best to avoid changes to formatting/code style if there is not a specific reason for it. Otherwise we will just change from one to another style because everybody likes something different.

src/routes/write-exercise/components/Feedback.tsx
```diff
@@ -56,6 +56,10 @@ const Feedback = ({ documentWithResult, submission, needsToBeRepeated }: Feedbac
     message = `${getLabels().exercises.write.feedback.almostCorrect1} „${submission}“ ${
       getLabels().exercises.write.feedback.almostCorrect2
     }`
+  } else if (!submission) {
```
steffenkleinle, line 59: I think you could merge this additional `else if` with the else block since its exactly identical except the message. There you already have a condition where the else case is what you want. So you'd just have to update the condition there.

lunars97: ok, thank you

src/routes/write-exercise/components/Feedback.tsx
```diff
@@ -59,11 +58,8 @@ const Feedback = ({ documentWithResult, submission, needsToBeRepeated }: Feedbac
   } else {
     Icon = CloseCircleIconBold
     background = BannerRed
-    message = needsToBeRepeated
-      ? `${getLabels().exercises.write.feedback.wrong}\n${wrongWithCorrectSolution}`
-      : wrongWithCorrectSolution
+    message = !submission ? wrongWithCorrectSolution : `${getLabels().exercises.write.feedback.wrong}`
```
steffenkleinle, line 61: Why did you remove the check for `needsToBeRepeated`?

lunars97: I thought that it was irrelevant in this exact component, because without including !submission, it was doing nothing.

lunars97: I have included both of them in else block now

src/routes/write-exercise/components/Feedback.tsx
```diff
@@ -56,18 +56,14 @@ const Feedback = ({ documentWithResult, submission, needsToBeRepeated }: Feedbac
...
-    background = BannerRed
-    message = wrongWithCorrectSolution
   } else {
     Icon = CloseCircleIconBold
     background = BannerRed
-    message = needsToBeRepeated
-      ? `${getLabels().exercises.write.feedback.wrong}\n${wrongWithCorrectSolution}`
-      : wrongWithCorrectSolution
+    message =
+      needsToBeRepeated && !submission
```
steffenkleinle, line 63: Doesn't this have to be an or?
```suggestion
      needsToBeRepeated || !submission
```

lunars97: only the first one function as intended

steffenkleinle: sorry, can you elaborate a little what you mean here? I am relatively sure it has to be a `||` here. We want to show the solution either if the user clicked on show solution (`!submission`) or the user entered it too many times (`!needsToBeRepeated`).

lunars97: yes, you are right. Logically it fits more when we say needsToBeRepeated || !submission. It works.

lunars97: It appears that the problem is in InteractionSection, needsToBeSubmitted was first declared there. Input and button becomes disabled once the word is entered and there is no chance to enter and to check a given word

sarahsporck (commented):
Don't forget to fix the pending comments :)

sarahsporck:
And don't forget to add release notes for thins that visibly change the app.

src/routes/write-exercise/components/Feedback.tsx
lunars97: Hi Steffen! needsToBeRepeated || !submission returns only true, but when I  use needsToBeRepeated && !submission, it both expressions are met successfully

steffenkleinle: Hey, sorry but I don't understand what you mean, could you elaborate a little bit more please?

lunars97: **needsToBeRepeated || !submission** are validating only the first expression in ternary operator, but we need to say that if the user's input is incorrect, show an error message with correct answer. Otherwise show **correct answer**. 
**needsToBeRepeated && !submission** are validating both expressions in ternary operator. It is validating correctly.

Hope I was able to deliver my thoughts precisely:)

steffenkleinle: Ahh now I get it, you changed the if with the else. Before the if was `${getLabels().exercises.write.feedback.wrong}\n${wrongWithCorrectSolution}` and the else was `wrongWithCorrectSolution`, now it's the other way round. Yes, then of course it is correct as it is.

src/routes/write-exercise/components/Feedback.tsx
```diff
@@ -59,11 +59,11 @@ const Feedback = ({ documentWithResult, submission, needsToBeRepeated }: Feedbac
...
+        ? wrongWithCorrectSolution
```
steffenkleinle, line 64: Could you please rename `wrongWtihCorrectSolution` to `correctSolutionHint` in L43 and the label `exercises.write.feedback.wrongWithSolution` in labels.json L78 to `solution`? Thanks :)

src/routes/write-exercise/components/Feedback.tsx
steffenkleinle: I think we don't have to check for `needsToBeRepeated` at all. We should show that the answer is wrong in any case when an anser was entered. So it should be:
``` ts
    message = submission
      ? `${getLabels().exercises.write.feedback.wrong}\n${wrongWithCorrectSolution}`
      : wrongWithCorrectSolution
```
Please mind: I changed the order back because positive conditions are easier to read then negative ones. Then you can also remove `needsToBeRepeated` from the props.

f1sh1918 (commented):
Please resolve conflicts and code formatting issues

f1sh1918 (approved):

PR #383, opened 2022-10-17
LUN-399: Add audio own vocabulary

f1sh1918:

Tested on Android (real device) & iOS(Emulator & real device)

Check current designs in Figma


android/build.gradle
```diff
@@ -3,19 +3,21 @@
 buildscript {
     ext {
         buildToolsVersion = "30.0.2"
-        minSdkVersion = 21
+        minSdkVersion = 24
```
f1sh1918, line 6: needed for `react-native-audio-recorder-player` library

src/routes/process-user-vocabulary/UserVocabularyProcessScreen.tsx
```diff
@@ -118,10 +182,76 @@ const UserVocabularyProcessScreen = ({ navigation }: UserVocabularyProcessScreen
   const pushImage = (uri: string): void => setImages(old => [...old, uri])
   const deleteImage = (uri: string): void => setImages(images => images.filter(image => image !== uri))
 
+  const onStartRecording = async (): Promise<void> => {
+    setMeteringResults([])
+    const id = await getNextUserVocabularyId()
+    const filename = `audio-${id}`
+    const path = Platform.select({
+      ios: `${filename}.m4a`,
```
f1sh1918, line 190: here we use the platform standards

src/routes/process-user-vocabulary/UserVocabularyProcessScreen.tsx
```diff
@@ -118,10 +182,76 @@ const UserVocabularyProcessScreen = ({ navigation }: UserVocabularyProcessScreen
   const pushImage = (uri: string): void => setImages(old => [...old, uri])
   const deleteImage = (uri: string): void => setImages(images => images.filter(image => image !== uri))
 
+  const onStartRecording = async (): Promise<void> => {
+    setMeteringResults([])
+    const id = await getNextUserVocabularyId()
+    const filename = `audio-${id}`
+    const path = Platform.select({
```
f1sh1918, line 189: Here we use the platform standards

jest.config.ts
```diff
@@ -24,4 +24,11 @@ export default {
   ],
   modulePaths: ['<rootDir>'],
   moduleDirectories: ['node_modules'],
+  globals: {
```
f1sh1918, line 27: Seems like the library got TS-Errors in their `index.ts` not sure how to ignore these in jest (tests fail) in a better way. You can comment out this and run `UserVocabularyProcessScreen.spec` to check error

ios/Lunes/Info.plist
```diff
@@ -44,6 +44,8 @@
 	</dict>
 	<key>NSCameraUsageDescription</key>
 	<string>Zum Scannen von QR-Codes benötigt LUNES Zugriff auf deine Kamera.</string>
+	<key>NSMicrophoneUsageDescription</key>
+    <string>Zur Aufnahme von Audio Files benötigt LUNES Zugriff auf dein Mikrofon</string>
```
SteffiStuffel, line 48: Maybe "Zur Aufnahme von Audios benötigt LUNES Zugriff auf dein Mikrofon". Finde das "Audio Files" irgendwie komisch.

Maybe you can also adjust the string for camera usage, as this is not only for QR-Codes any longer and I forgot to adjust this.

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
```diff
@@ -0,0 +1,153 @@
...
+  onStopRecording,
+  recordingTime,
+  meteringResults,
+}: AudioRecordOverlayProps): ReactElement => {
+  const [isPressed, setIsPressed] = useState<boolean>(false)
+  const permissionGranted = useGrantPermissions(
+    Platform.OS === 'android'
+      ? [
+          PERMISSIONS.ANDROID.RECORD_AUDIO,
+          PERMISSIONS.ANDROID.READ_EXTERNAL_STORAGE,
```
SteffiStuffel, line 101: I think this permissions should be removed. Images are also not saved to external storage and it would really make sense to store images and audios in the same location. Also saving to internal storage ensures audios cannot be deleted e.g. from media library. Please change to app-specific storage

src/hooks/useGrantPermissions.ts
```diff
@@ -0,0 +1,31 @@
+import { useEffect, useState } from 'react'
```
SteffiStuffel, line 1: Good idea to create a hook for this, but i think it really makes sense to use it everywhere, so please adjust CameraOverlay.tsx, so it uses this hook.

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
```diff
@@ -0,0 +1,153 @@
...
+          <CloseIcon width={theme.spacingsPlain.lg} height={theme.spacingsPlain.lg} />
+        </Icon>
+        <Content>
+          <HeadingContainer>
+            <Heading>{isPressed ? talk : hold}</Heading>
+          </HeadingContainer>
+          <InfoContainer>
+            <MeteringInfo>
+              {cleanedMetering.map((element, index) => (
+                // eslint-disable-next-line react/no-array-index-key -- no better key available
```
SteffiStuffel, line 124: I am not really happy with this, as I think linting rules should not be disabled and this rule does really prevent bug, but in this case there should not occure a bug, but i think this can be improved.
Please add e.g. a timestamp or id for each bar.

SteffiStuffel (changes requested):
Can you please add some tests?

When missing a permission, everything but the record button is displayed. Please show a proper error message, with option to navigate to settings to adjust permissions (See QR code scanner)

Audio button should change color, when audio is played, as it is done in wordlist and exercises.

Delete-icon is cut off. On my emulator (Pixel 4a) it is cut off on the left, on my real device it is cut off on the top.


I think the structure here should be changed: The "onStartRecording, onStopRecording, metringResults, recordingTime..." should be moved to AudioRecordOverlay.tsx, as the UserVocabularyScreen is not responsible for the audioRecording-stuff.

When clicking on saveVocabulary the file in the temporary recordingPath is saved with "react-native-fs"-writeFile method saved to internal storage. Compare to handling of images.

jest.config.ts
sarahsporck: Disabling typechecking is currently not possible for importet files in typescript.
See https://github.com/Microsoft/TypeScript/wiki/FAQ#why-is-a-file-in-the-exclude-list-still-picked-up-by-the-compiler
Best would probably be to open an pr in the repo and wait until they resolve this...

src/routes/process-user-vocabulary/UserVocabularyProcessScreen.tsx
```diff
@@ -157,14 +284,27 @@ const UserVocabularyProcessScreen = ({ navigation }: UserVocabularyProcessScreen
...
-          iconSize={theme.spacingsPlain.xl}
-        />
+        {recordingPath ? (
+          <AudioContainer style={{ opacity: fadeAnim }}>
+            <>
+              <AudioText>{recordingPath.substring(recordingPath.lastIndexOf('/') + 1).toUpperCase()}</AudioText>
+              <DeleteContainer onPress={() => deleteRecording()} testID='delete-audio-recording'>
+                <CloseCircleIconBlue width={theme.spacingsPlain.lg} height={theme.spacingsPlain.lg} />
+              </DeleteContainer>
+              <PlayContainer onPress={() => onStartPlay()}>
```
sarahsporck, line 294: Could you keep the button pressed, while the audio is played? Also there is no indication that the button is pressed. Wouldn't the color change to dark if you press the icon?

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
```diff
@@ -0,0 +1,153 @@
...
+            <RecordingInfo>{recordingTime.slice(1, recordingTime.length)}</RecordingInfo>
+          </InfoContainer>
+          {permissionGranted && (
+            <RecordIcon
+              onPressIn={() =>
+                onStartRecording()
+                  .catch(reportError)
+                  .finally(() => setIsPressed(true))
+              }
+              onPressOut={() =>
```
sarahsporck, line 137: This pressout is so incredibly slow. It takes about 2 or three seconds until the recording has stopped. Maybe memoing the MeteringInfo and bar like [this](https://stackoverflow.com/questions/71844887/reactjs-memo-dynamic-for-loop-container-with-dynamic-memoized-children-how-to) could help?
Id' also prefer if the whole metering set up would happen inside the Overlay, meaning we'd need to pass the audio recorder and stuff. But feel free to argue against that. :)

sarahsporck (commented):
It seems the `x` is cur off on android - Tested on real device.

Screenshot_audio_x_cut_off

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
f1sh1918: unfortunately timestamp also generate key duplicate so i decided to use uuid generator

src/hooks/useGrantPermissions.ts
f1sh1918: i adjusted the hook to make it work for `CameraOverlay` and added in there

src/routes/process-user-vocabulary/UserVocabularyProcessScreen.tsx
f1sh1918: well the outline svg icons don't have a styleable background. I switched to the normal AudioPlayer Icon and used the same color indicators

f1sh1918: is that fine for the moment? If not i have to get new svg files from the design team

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
f1sh1918: removed

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
f1sh1918: hm i tested on real devices and can not reproduce any slowness. If you pressOut the Modal is closed immediately. Tested on real device! How did you reproduce that?

Well then we would have to split the recording and play functions because playing is executed in the `UserVocabularyProcessScreen`
We also have to delete the metering results from this screen by clicking delete button.
I think its getting more complicated, even i understand that from a logical perspective it makes sense to put it in the `AudioRecordOverlay`

f1sh1918:
> Can you please add some tests?
> 
> When missing a permission, everything but the record button is displayed. Please show a proper error message, with option to navigate to settings to adjust permissions (See QR code scanner)
> 
> Audio button should change color, when audio is played, as it is done in wordlist and exercises.
> 
> Delete-icon is cut off. On my emulator (Pixel 4a) it is cut off on the left, on my real device it is cut off on the top.
> 
> I think the structure here should be changed: The "onStartRecording, onStopRecording, metringResults, recordingTime..." should be moved to AudioRecordOverlay.tsx, as the UserVocabularyScreen is not responsible for the audioRecording-stuff.
> 
> When clicking on saveVocabulary the file in the temporary recordingPath is saved with "react-native-fs"-writeFile method saved to internal storage. Compare to handling of images.

1. I added one test unfortunately the audioRecorder 
2. fixed that by adding notAuth view
3. Talked to Nikola and replaced Icon by our standard red icon and added indicator for `isPlaying`
4. As mentioned can not reproduce that cut off, no idea why (not pixel 4a, nor real device) pls help.
5. changed the structure, pls check
6. I dunno what you mean. I have to pass a path to the `startRecording` method. Can not find a hint to first get a temporary path, so the file gets overwritten if the item was not saved.

jest.config.ts
f1sh1918: and so i should keep this pr open until fixed? This can take months... I can open an issue and mention that in the `jest.config.js`

src/routes/process-user-vocabulary/UserVocabularyProcessScreen.tsx
f1sh1918: we can use the audio player standard icon

f1sh1918:
I opened an issue for the internal & external write permissions on android
https://github.com/hyochan/react-native-audio-recorder-player/issues/492

Unfortunately there is no appropriate audio recording lib

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
f1sh1918: i moved stuff to overlay

jest.config.ts
f1sh1918: fixed this with a path. gonna create a issue in the library for that

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
```diff
@@ -0,0 +1,201 @@
...
+`
+
+const MeteringInfo = styled.View`
+  flex-direction: row;
+  align-self: center;
+  min-height: ${hp('12%')}px;
+`
+
+const MeteringBar = styled.View<{ height: number }>`
+  min-height: ${props => props.theme.spacings.xxs}
```
sarahsporck, line 65:
```suggestion
  min-height: ${props => props.theme.spacings.xxs};
```

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
```diff
@@ -0,0 +1,201 @@
...
+}
+
+const recordingTimeInit = '00:00'
+const factor = 1000
+// The recording library does align the android power recording level to ios for metering as it's done in different libraries https://github.com/ziscloud/sound_recorder/blob/46544fc23b71e6f929b372fad0313c70b0301371/android/src/main/java/com/neuronbit/sound_recorder/SoundRecorderPlugin.java#L375
+const androidFactor = 0.25
+
+// Zero alignment of values with the minimum metering value
+const cleanUpMeteringResults = (meteringResults: number[]): number[] => {
+  const filteredResults = meteringResults.filter(el => el !== Math.min(...meteringResults))
```
sarahsporck, line 94: Why not do this? 
```suggestion
  const minimumMeteringValue = Math.min(...meteringResults)
  return meteringResults.map(el => el + Math.abs(minimumMeteringValue))
```

and why do you need to filter the minimum value? Shouldn't it be enough to normalize (the second step)?

sarahsporck (commented):
So for the icon i just had to replace the svg. The viewBox was wy off, so my old phone didn't know what to do. Let's keep an eye on this to prevent this from happening again.

The recording is still pretty slow on my phone, but I think that's because it's old. If you don't have any issues we can ignore that, I guess.

And finally I'd like to propose, that recordings should have a maximum length of 10 seconds. This will help avoid any memory  or performance issues in the first place. Just tell me what's your opinion on this :)

Otherwise, the feature seems to work fine. I roughly rereviewd ou're changes, but cannot really give any input on the permission patch. As far as I've looked it up you **have to** specify the permissions possibly used in the Manifest, even when you will not use a specific Permission in the app under certain circumstances.

f1sh1918:
> So for the icon i just had to replace the svg. The viewBox was wy off, so my old phone didn't know what to do. Let's keep an eye on this to prevent this from happening again.
> 
> The recording is still pretty slow on my phone, but I think that's because it's old. If you don't have any issues we can ignore that, I guess.
> 
> And finally I'd like to propose, that recordings should have a maximum length of 10 seconds. This will help avoid any memory or performance issues in the first place. Just tell me what's your opinion on this :)
> 
> Otherwise, the feature seems to work fine. I roughly rereviewd ou're changes, but cannot really give any input on the permission patch. As far as I've looked it up you **have to** specify the permissions possibly used in the Manifest, even when you will not use a specific Permission in the app under certain circumstances.

I think you can create a ticket to stop recording after 10 seconds. There are enough changes in this pr and was no requirement yet.
Well according to the permissions for external storage, the library does not handle them good, they should check if the passed path is internal or external and then request the permission if needed. They request them in any case.But yes you will have it in the manifest then merged from 3rd party library, even they are not prompted. 
Maybe @steffenkleinle has any idea about that :)

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
f1sh1918: I'm gonna check your solution. the filtering is an android problem. When the records start there are dozens of minimum metering values without any sense and they make the metering bar very ugly. you can just log the metering results on android and gonna see why.

f1sh1918:
> So for the icon i just had to replace the svg. The viewBox was wy off, so my old phone didn't know what to do. Let's keep an eye on this to prevent this from happening again.
> 
> The recording is still pretty slow on my phone, but I think that's because it's old. If you don't have any issues we can ignore that, I guess.
> 
> And finally I'd like to propose, that recordings should have a maximum length of 10 seconds. This will help avoid any memory or performance issues in the first place. Just tell me what's your opinion on this :)
> 
> Otherwise, the feature seems to work fine. I roughly rereviewd ou're changes, but cannot really give any input on the permission patch. As far as I've looked it up you **have to** specify the permissions possibly used in the Manifest, even when you will not use a specific Permission in the app under certain circumstances.

Ah well and it limited the recording to 5seconds. No need for separate ticket

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
f1sh1918: checked and would keep my solution with the filtering. looks way better

android/app/src/main/AndroidManifest.xml
```diff
@@ -1,8 +1,9 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-  package="app.lunes">
+          package="app.lunes">
```
SteffiStuffel, line 2: please undo.

ios/Lunes.xcodeproj/project.pbxproj
```diff
@@ -634,7 +634,7 @@
 				COPY_PHASE_STRIP = NO;
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
 				ENABLE_TESTABILITY = YES;
-				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = "arm64 ";
+				"EXCLUDED_ARCHS[sdk=iphonesimulator*]" = "";
```
SteffiStuffel, line 637: Why have you changes this?

src/constants/labels.json
```diff
@@ -161,9 +161,15 @@
...
     "camera": {
       "noAuthorization": {
-        "description": "Kamerazugriff erforderlich. Öffnen Sie die Einstellungen, tippen Sie auf Berechtigungen und tippen Sie auf Erlauben.",
-        "settings": "Einstellungen"
+        "description": "Kamerazugriff erforderlich. Öffnen Sie die Einstellungen, tippen Sie auf Berechtigungen und tippen Sie auf Erlauben."
       }
+    },
+    "audio": {
+      "noAuthorization": {
+        "description": "Mikrofonzugriff erforderlich. Öffnen Sie die Einstellungen, tippen Sie auf Berechtigungen und tippen Sie auf Erlauben."
```
SteffiStuffel, line 169: I see this was wrong also for images, but lunes uses "du" not "Sie". Please adjust.

patches/react-native-audio-recorder-player+3.5.1.patch
```diff
@@ -0,0 +1,4025 @@
+diff --git a/node_modules/react-native-audio-recorder-player/android/build/.transforms/1074abf937aa2e6e3bf6d87bd7a32c8e/results.bin b/node_modules/react-native-audio-recorder-player/android/build/.transforms/1074abf937aa2e6e3bf6d87bd7a32c8e/results.bin
```
SteffiStuffel, line 1: I think it would be nice, if you can add the actual patch you wrote, not only the result the tool outputs, as it is impossible to understand what happens here, and also hard to check if the patch is needed after an update?

package.json
```diff
@@ -98,6 +102,8 @@
     "jest-extended": "^2.0.0",
     "jest-transform-stub": "^2.0.0",
     "metro-react-native-babel-preset": "^0.66.2",
+    "patch-package": "^6.5.0",
+    "postinstall-postinstall": "^2.1.0",
```
SteffiStuffel, line 106: Is this postinstall-postinstall package really necessary. This is only needed if we need to execute a postinstall after removing a package. This is not the case, is it?

src/routes/process-user-vocabulary/__tests__/UserVocabularyProcessScreen.spec.tsx
```diff
@@ -31,4 +31,12 @@ describe('UserVocabularyProcessScreen', () => {
     expect(a).toHaveLength(3)
     expect(getByText(getLabels().userVocabulary.creation.addImage)).toBeDisabled()
   })
+  it('should be able to add a recording', async () => {
+    const { findByTestId, getByText } = render(<UserVocabularyProcessScreen navigation={navigation} />)
+    const addAudioButton = getByText(getLabels().userVocabulary.creation.addAudio)
+    expect(addAudioButton).not.toBeNull()
+    fireEvent.press(addAudioButton)
+    const recordButton = await findByTestId('record-audio-button')
+    expect(recordButton).not.toBeNull()
```
SteffiStuffel, line 40: It does not really make sense to check for notNull if query with "findBy", as the findBy throws an error, if there is no match. You need to use queryBy to get null for no matches.

src/routes/process-user-vocabulary/__tests__/UserVocabularyProcessScreen.spec.tsx
```diff
@@ -31,4 +31,12 @@ describe('UserVocabularyProcessScreen', () => {
     expect(a).toHaveLength(3)
     expect(getByText(getLabels().userVocabulary.creation.addImage)).toBeDisabled()
   })
+  it('should be able to add a recording', async () => {
```
SteffiStuffel, line 34: Please add new line before the test.

src/routes/process-user-vocabulary/UserVocabularyProcessScreen.tsx
```diff
@@ -162,14 +270,27 @@ const UserVocabularyProcessScreen = ({ navigation }: UserVocabularyProcessScreen
...
-          onPress={() => null}
-          label={addAudio}
-          buttonTheme={BUTTONS_THEME.text}
-          iconLeft={MicrophoneCircleIcon}
-          iconSize={theme.spacingsPlain.xl}
-        />
+        {recordingPath ? (
+          <AudioContainer style={{ opacity: fadeAnim }}>
+            <>
+              <AudioText>{recordingPath.substring(recordingPath.lastIndexOf('/') + 1).toUpperCase()}</AudioText>
```
SteffiStuffel, line 276: So here we are using the `react-native-audio-recorder-player` for playing the sound file and in all other cases we are using `react-native-sound-player` for playing audio. I think it would be good to reuse the existing AudioPlayer.tsx component for playing an audio to have this consistent.

SteffiStuffel (changes requested):
Looks already really good. Tested and works fine. 

Just some small adjustments needed. Please especially unify the audio playing with the already existing AudioPlayer.

ios/Lunes.xcodeproj/project.pbxproj
f1sh1918: just by mistake. this is only for simulators needed on m1. undone

patches/react-native-audio-recorder-player+3.5.1.patch
f1sh1918: i could add the the kotlin file i adjusted under patchfiles. and i removed the permission in the android manifest of the library. Gonna create a ticket to check if this has to be applied after next lib update

f1sh1918: https://issues.tuerantuer.org/browse/LUN-458

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
```diff
@@ -0,0 +1,205 @@
...
+        <>
+          {permissionGranted && (
+            <Content>
+              <HeadingContainer>
+                <Heading>{isPressed ? talk : hold}</Heading>
+              </HeadingContainer>
+              <InfoContainer>
+                <MeteringInfo>
+                  {cleanedMetering.map(element => (
+                    <MeteringBar key={uuid.v4().toString()} height={element} />
```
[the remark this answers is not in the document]
f1sh1918: what is your suggestion in this case. index shouldn't be used and timestamps don't work here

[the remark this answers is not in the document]
steffenkleinle: From the react docs (https://reactjs.org/docs/lists-and-keys.html#keys):
> When you don’t have stable IDs for rendered items, you may use the item index as a key as a last resort

and
> We don’t recommend using indexes for keys if the order of items may change. This can negatively impact performance and may cause issues with component state.

I think since the order of items should not change we should be good with just indexes. Especially since with your solution `uuid.v4()` is run on every render for every element anew, therefore changes each time and therefore is even worse performance wise (didn't test this so not that sure if that is actually the case).

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
```diff
@@ -0,0 +1,205 @@
...
+  justify-content: flex-end;
+`
+
+interface AudioRecordOverlayProps {
+  onClose: () => void
+  setShowAudioRecordOverlay: Dispatch<SetStateAction<boolean>>
+  audioRecorderPlayer: AudioRecorderPlayer
+  setRecordingPath: Dispatch<SetStateAction<string | null>>
+  recordingPath: string | null
+  setRecordSeconds: Dispatch<SetStateAction<number>>
```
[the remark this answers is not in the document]
f1sh1918: @steffenkleinle  i don't understand this suggestion. These interface are correct

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
```diff
@@ -0,0 +1,205 @@
...
+const factor = 1000
+// The recording library does not align the android power recording level to ios for metering as it's done in different libraries https://github.com/ziscloud/sound_recorder/blob/46544fc23b71e6f929b372fad0313c70b0301371/android/src/main/java/com/neuronbit/sound_recorder/SoundRecorderPlugin.java#L375
+const androidFactor = 0.25
+const maxRecordingTime = 5000
+
+// Zero alignment of values with the minimum metering value
+const cleanUpMeteringResults = (meteringResults: number[]): number[] => {
+  const filteredResults = meteringResults.filter(el => el !== Math.min(...meteringResults))
+  return filteredResults.map(el => el + Math.abs(Math.min(...filteredResults)))
+}
```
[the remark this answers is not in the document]
f1sh1918: updated in last commit!

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  const onStartRecording = async (): Promise<void> => {
+    setMeteringResults([])
+    const uri = await audioRecorderPlayer.startRecorder(undefined, undefined, true)
+    audioRecorderPlayer.addRecordBackListener(async e => {
+      setMeteringResults(oldMeteringResults => [
+        ...oldMeteringResults,
+        Math.floor(getCurrentMetering(e.currentMetering)),
+      ])
```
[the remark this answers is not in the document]
f1sh1918: no in the setState method its `previousState` its just set empty once on start recording but not in the `addRecordListener`

Example clog
```js
 LOG  []
 LOG  [-40]
 LOG  [-40, -23]
 LOG  [-40, -23, -10]
 LOG  [-40, -23, -10, -5]
```

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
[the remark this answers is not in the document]
f1sh1918: ok perfect. i just revert that to index

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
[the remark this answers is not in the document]
steffenkleinle: okay, missed that, thanks for explaining :)

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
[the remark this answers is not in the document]
steffenkleinle: Yeah, but I think they are too specific imo. AudioRecordOverylay should not care if it is a function updating a react state, a function updating redux state or a function doing something completely else. The additional specificity doesn't add any benefit and only may have to be changed in the future. For example if we at some point have to do something else when the recording seconds change instead of just updating the state we have to change the types here (may also already be the case depending on the other comments on not using useEffects on state changes).

src/hooks/useGrantPermissions.ts
```diff
@@ -0,0 +1,48 @@
...
+    const subscription = AppState.addEventListener('change', nextAppState => {
+      if ((appState.current === 'inactive' || appState.current === 'background') && nextAppState === 'active') {
+        setPermissionRequested(false)
+      }
+      appState.current = nextAppState
+    })
+    return subscription.remove
+  }, [setPermissionRequested])
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
f1sh1918: hm can you give me an example how to do it in this case

src/hooks/useGrantPermissions.ts
```diff
@@ -0,0 +1,48 @@
...
+import { Permission, request, requestMultiple, RESULTS } from 'react-native-permissions'
+
+import { reportError } from '../services/sentry'
+
+interface UseGrantPermissionsReturnType {
+  permissionRequested: boolean
+  permissionGranted: boolean
+}
+
+const useGrantPermissions = (permissions: Permission | Permission[]): UseGrantPermissionsReturnType => {
```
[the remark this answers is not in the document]
f1sh1918: adjusted to single permission

src/routes/process-user-vocabulary/UserVocabularyProcessScreen.tsx
f1sh1918: alright then i think we should refactor the `vocabularyItem` to just `audio` passend

src/routes/process-user-vocabulary/UserVocabularyProcessScreen.tsx
```diff
@@ -52,10 +56,53 @@ const ThumbnailContainer = styled.View`
...
   navigation: StackNavigationProp<RoutesParams, 'UserVocabularyProcess'>
 }
 
+const accuracy = 0.1
+const playTolerance = 200
+
+const audioRecorderPlayer = new AudioRecorderPlayer()
+audioRecorderPlayer.setSubscriptionDuration(accuracy).catch(e => e)
+
 const UserVocabularyProcessScreen = ({ navigation }: UserVocabularyProcessScreenProps): ReactElement => {
```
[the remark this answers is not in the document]
f1sh1918: i used our `AudioRecorder` component. since we use the `fadeIn` after setRecording path its a little bit complicated to extract the Animation. If you want you can create a separate component for the animation and refactor it

src/routes/process-user-vocabulary/components/AudioRecordOverlay.tsx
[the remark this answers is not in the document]
f1sh1918: alright i got you point thx for explanation

src/routes/process-user-vocabulary/UserVocabularyProcessScreen.tsx
```diff
@@ -162,14 +270,27 @@ const UserVocabularyProcessScreen = ({ navigation }: UserVocabularyProcessScreen
...
+          </AudioContainer>
+        ) : (
+          <AddAudioButton
+            onPress={() => setShowAudioRecordOverlay(true)}
+            label={addAudio}
+            buttonTheme={BUTTONS_THEME.text}
+            iconLeft={MicrophoneCircleIcon}
+            iconSize={theme.spacingsPlain.xl}
+          />
+        )}
```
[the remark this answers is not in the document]
f1sh1918: But this not the AudioRecorder its just a component to play or delete an audiofile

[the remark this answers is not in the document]
f1sh1918: moved to separate component. thx for all the hints. I removed the fadeAnimation, since its breaking it separate component. If this is mandatory lets create a separate ticket for a reusable animation container or sth

src/routes/process-user-vocabulary/__tests__/UserVocabularyProcessScreen.spec.tsx
f1sh1918: i need the await and findByTestId since its only shown after `setState` adjusted `expect(recordButton).toBeDefined()`

src/routes/process-user-vocabulary/UserVocabularyProcessScreen.tsx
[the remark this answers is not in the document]
f1sh1918: put everything in a new component

src/hooks/useGrantPermissions.ts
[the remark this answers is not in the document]
steffenkleinle: I think you could just use the code above :) Or what more do you need?

steffenkleinle:
@f1sh1918 I am really sorry but the changes I proposed lead to flickering of the screen if you disallow the permissions (see 

https://user-images.githubusercontent.com/29095487/203601900-3ac6046a-9b28-4e7c-8396-c87977914273.mp4

). 

I don't get why but couldn't figure it out therefore I reverted the changes in #394 and fixed the flickering record audio button. Sorry!

I also created an proposal to refactor AudioPlayer in #395 to make use of tts more explicit. Feel free to close it if you don't like it.


I still get an error though if you just press the recording button and don't hold it:

https://user-images.githubusercontent.com/29095487/203602109-9a10da0d-f08a-4cb7-8575-95263176e49f.mp4

steffenkleinle:
@f1sh1918 not 100% sure but only seems like this only happens if a vocabulary was saved before already and does not happen again once the recorder is closed and opened again. Can you reproduce? If that is the case it seems like the audio player is somehow not correctly stopped or something.

Seems to be the same issue: https://github.com/hyochan/react-native-audio-recorder-player/issues/490

f1sh1918:
> @f1sh1918 ~Not 100% sure but only seems like this only happens if a vocabulary was saved before already and does not happen again once the recorder is closed and opened again. Can you reproduce? If that is the case it seems like the audio player is somehow not correctly stopped or something.~
> 
> Edit: No idea if the above is true but seems to be a common problem ([hyochan/react-native-audio-recorder-player#490](https://github.com/hyochan/react-native-audio-recorder-player/issues/490), https://stackoverflow.com/questions/16221866/mediarecorder-failed-when-i-stop-the-recording). Since it seems to be on purpose that an error is thrown in that case, I figured just ignoring is totally fine. Made a proposal in #397, let me know what you think.

I'm fine with that and merged it.
@sarahsporck @SteffiStuffel @LeandraH  @steffenkleinle  please leave a review :)

docs/patches.md
```diff
@@ -0,0 +1,13 @@
...
+Therefore here is a short description how to patch a 3rd party package.
+
+## Manual
+
+- `yarn install <package name>` Install the package you want to fix. Ensure that you didn't change anything yet
+- apply the changes in the `node_modules/<package name>` files.
+- `yarn android` or retrigger build in Xcode to test your changes
+- `npx patch-package <package name>` (patch-package is required as dev depencency)
+- commit the created patch file
+- lock the package in `package.json` because the patch may not apply correctly on different package versions
```
LeandraH, line 13: Very cool, thank you!

LeandraH (approved):
Looks good, tested on a real iPhone. Where, by the way, the recording was at a very low volume when comparing to a recording on Voice Memos. Maybe you can increase that a little still?

f1sh1918:
> > @f1sh1918 ~Not 100% sure but only seems like this only happens if a vocabulary was saved before already and does not happen again once the recorder is closed and opened again. Can you reproduce? If that is the case it seems like the audio player is somehow not correctly stopped or something.~
> > Edit: No idea if the above is true but seems to be a common problem ([hyochan/react-native-audio-recorder-player#490](https://github.com/hyochan/react-native-audio-recorder-player/issues/490), https://stackoverflow.com/questions/16221866/mediarecorder-failed-when-i-stop-the-recording). Since it seems to be on purpose that an error is thrown in that case, I figured just ignoring is totally fine. Made a proposal in #397, let me know what you think.
> 
> I'm fine with that and merged it. @sarahsporck @SteffiStuffel @LeandraH @steffenkleinle please leave a review :)

Hm i would have to patch the used library. There are no option to set the recording volume only the playing volume which we don't use

src/hooks/useGrantPermissions.ts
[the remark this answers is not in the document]
steffenkleinle: My proposal did not work so we just stuck with @f1sh1918 version. We both did not have the time to look more into this.

steffenkleinle (approved):

PR #384, opened 2022-10-18
LUN-449: Add jsx-expressions/strict-logical-expressions linting rule

steffenkleinle:

Went for https://github.com/hpersson/eslint-plugin-jsx-expressions instead of https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-leaked-render.md because it supports ts types and therefore prevents false positives (e.g. for boolean props). Sadly, the rule only supports inline jsx, created an issue for that: https://github.com/hpersson/eslint-plugin-jsx-expressions/issues/11
Therefore still take care with jsx oustide of the return function.


SteffiStuffel:
Hmh, if the new plugin supports ts types  `someBoolean && <Text>...` should be safe. As this is only not working, if the value is non-boolean or undefined, which is already prevented by ts-checks or am I missing something here?

steffenkleinle:
> Hmh, if the new plugin supports ts types `someBoolean && <Text>...` should be safe. As this is only not working, if the value is non-boolean or undefined, which is already prevented by ts-checks or am I missing something here?

@SteffiStuffel not sure if I understand your question correctly but no, it is not prevented yet. It is still possible to do stuff like
``` tsx
const asdf = ''
return asdf && <div />
```
What am I missing?

SteffiStuffel:
If the plugin supports types. This should be fine:
```
const b: Boolean = true
...
render (...
   b && <Text>....
```

So why is it needed to write this?
```
const b: Boolean = true
...
render (...
   !!b && <Text>....
```

steffenkleinle:
Sorry but I still don't really get what you mean. The following is totally fine and not errored by the linter:
> ``` tsx
> const b: boolean = props.b
> return (
>   <View>
>      {b && <Text>some text</Text>}
>   </View>
>  )
> ```

However, the following is not fine and errored because it may leak an empty string which would crash react native:
> ``` tsx
> const s: string = props.s
> return (
>   <View>
>      {s && <Text>some text</Text>}
>   </View>
>  )
> ```

steffenkleinle:
@sarahsporck what do you think?

f1sh1918 (approved):

SteffiStuffel:
Ah oke, I see. Thanks for the explanation.

SteffiStuffel (approved):

PR #391, opened 2022-11-07
LUN-452: upgraded react-native version in order to fix the build failure

lunars97:


sarahsporck (approved):
We should probably generally think about updating our react version to a [supported version](https://github.com/reactwg/react-native-releases#which-versions-are-currently-supported). However, as a hotfix this update is totally fine.

steffenkleinle (approved):

PR #393, opened 2022-11-16
LUN-454: User vocabulary image thumbnail

f1sh1918:

- Removed unneeded getImages fkt
- instead moved image files from temp folder to destination folder
- prefixed uri for local image files
- update podfile.lock, since was missing after react native update


src/services/helpers.ts
```diff
@@ -206,16 +204,3 @@ export const getSortedAndFilteredVocabularyItems = (
 
 export const willNextExerciseUnlock = (previousScore: number | undefined, score: number): boolean =>
   score > SCORE_THRESHOLD_UNLOCK && (previousScore ?? 0) <= SCORE_THRESHOLD_UNLOCK
-
```
f1sh1918, line 209: i'm not sure why we reading here a text file with the reference to the cached image (which may be deleted meanwhile)

SteffiStuffel: This function is needed, as we need to load the images, for e.g. showing the VocabularyListItem  or any other component correctly. Own vocabulary should be possible to be passed and handled like lunes vocab or anything else.
This causes an error, e.g. if you open the list of own vocabulary:  `ERROR  TypeError: undefined is not an object (evaluating 'images.length')`

SteffiStuffel (changes requested):

src/routes/process-user-vocabulary/UserVocabularyProcessScreen.tsx
```diff
@@ -94,8 +94,8 @@ const UserVocabularyProcessScreen = ({ navigation }: UserVocabularyProcessScreen
 
       const imagePaths = await Promise.all(
         images.map(async (image, index) => {
-          const path = `${DocumentDirectoryPath}/image-${id}-${index}.txt`
-          await writeFile(path, image, 'utf8')
+          const path = `file:///${DocumentDirectoryPath}/image-${id}-${index}.jpg`
+          await moveFile(image, path)
```
SteffiStuffel, line 98: File should not be moved, as the file should only be cached, when the the camera overlay "saves" = caches it. So if a user cancels the creation of a vocabulary no image files that do not belong to a saved images are there. Only if a user vocabulary is saved, the images should be saved.

src/services/helpers.ts
f1sh1918: can not reproduce this error. tested on real devices

src/routes/process-user-vocabulary/UserVocabularyProcessScreen.tsx
f1sh1918: the file will be cached and only moved if you save the vocabulary.

steffenkleinle (approved):
Tested on android, looks good to me :+1:

src/services/helpers.ts
steffenkleinle: Didn't see it on android as well

SteffiStuffel: My bad, needed to clear async storage.

SteffiStuffel (approved):
Good work, thanks for fixing this :+1:

PR #398, opened 2022-11-29
LUN-442: Add a back button to the word list

LeandraH:
This pull request belongs to the issue https://issues.tuerantuer.org/browse/LUN-442 on our bug tracker.

It adds a back button to the word list.
image


release-notes/unreleased/LUN-442-back-button-in-word-list.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: LUN-442
+show_in_stores: true
+platforms:
+  - android
+  - ios
+de: Button zum vorherigen Wort hinzugefügt
```
f1sh1918, line 6:
```suggestion
de: In der Wortliste kann jetzt zum vorherigen Wort navigiert werden
```
Is just a suggestion. It think the release notes should be understandable for the user what was improved. Maybe also simplify "navigiert" somehow

f1sh1918 (approved):
Tested on ios. Works fine! Nice work :)

release-notes/unreleased/LUN-442-back-button-in-word-list.yml
LeandraH: Great idea, thank you!

src/routes/vocabulary-detail-exercise/VocabularyDetailExerciseScreen.tsx
```diff
@@ -54,31 +57,42 @@ const VocabularyDetailExerciseScreen = ({ route, navigation }: VocabularyDetailE
...
+            {hasPreviousVocabularyItem && (
+              <Button
+                label={getLabels().exercises.previous}
+                iconLeft={ArrowLeftIcon}
+                onPress={goToPreviousWord}
+                buttonTheme={BUTTONS_THEME.text}
+              />
+            )}
+            {!hasNextVocabularyItem && (
+              <Button label={closeExerciseLabel} onPress={navigation.goBack} buttonTheme={BUTTONS_THEME.contained} />
```
steffenkleinle, line 85: Why did you change this? Is it in the designs like that? I personally think that the placement of the previous word button should not change/switch with this one only on the last word. Instead, I would place the next word/close exercise always on top and the previous word button always on bottom.

steffenkleinle (approved):
Tested on android

PR #404, opened 2023-01-10
LUN-377: Upgrade react-native to 69

LeandraH:
This pull request belongs to the issue https://issues.tuerantuer.org/browse/LUN-377.

I had to patch two packages because of deprecated prop types in 0.69. There are two tickets open ([one](https://issues.tuerantuer.org/browse/LUN-439) and [two](https://issues.tuerantuer.org/browse/LUN-465)) to replace those two but I decided it was out of scope of this PR.


sarahsporck (commented):
just fyi: lots of the new architecture stuff will be removed again in react native 0.71.
Also, the types which have been removed in rn 0.69 will (temporarily) be readded in 0.71. That means we can remove the patches then.

sarahsporck (approved):
Tested on android. (Not including individual Area). 
Tests need to be fixed. Ping me if you need help with this.

f1sh1918 (approved):
Works fine on ios :)
Nice work!

If not already happend pls create small tasks in jira for each patch you created with some documentation and a reminder to remove it when possible

LeandraH:
I'd deprioritized this ticket to be honest, haven't looked at it in two weeks. Will give it another whirl tomorrow and get back to you if I can't figure it out, thanks for checking :)

sarahsporck:
I had a similar error while upgrading Integreat. Updating [react-native-sound-player](https://github.com/johnsonsu/react-native-sound-player/pull/148) should fix the build :)

PR #405, opened 2023-01-10
LUN-450: write tests for NotAuthorizedView

lunars97:

### DIFF
--- a/src/components/__tests__/NotAuthorisedView.spec.tsx
+++ b/src/components/__tests__/NotAuthorisedView.spec.tsx
@@ -1,0 +1,43 @@
+import { fireEvent } from '@testing-library/react-native'
+import { mocked } from 'jest-mock'
+import React from 'react'
+import { Linking } from 'react-native'
+
+import { getLabels } from '../../services/helpers'
+import render from '../../testing/render'
+import NotAuthorisedView from '../NotAuthorisedView'
+
+jest.mock('react-native/Libraries/Linking/Linking', () => ({
+  openSettings: jest.fn(),
+}))
+
+describe('NotAuthorisedView', () => {
+  const setVisible = jest.fn()
+
+  it('should render description', () => {
+    const { getByText } = render(
+      <NotAuthorisedView setVisible={setVisible} description={getLabels().general.camera.noAuthorization.description} />
+    )
+    const isNoAuthDescription = getByText(getLabels().general.camera.noAuthorization.description)
+    expect(isNoAuthDescription).toBeTruthy()
+  })
+
+  it('should successfully go back', () => {
+    const { getByText } = render(
+      <NotAuthorisedView setVisible={setVisible} description={getLabels().general.camera.noAuthorization.description} />
+    )
+    const buttonVisible = getByText(getLabels().general.back)
+    fireEvent.press(buttonVisible)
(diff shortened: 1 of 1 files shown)


src/components/__tests__/NotAuthorisedView.spec.tsx
```diff
@@ -0,0 +1,38 @@
...
+  openSettings: jest.fn(),
+}))
+
+describe('NotAuthorisedView', () => {
+  const setVisible = jest.fn()
+  it('should render description', () => {
+    const { getByTestId } = render(
+      <NotAuthorisedView setVisible={setVisible} description={getLabels().general.camera.noAuthorization.description} />
+    )
+    expect(getByTestId('no-auth')).toBeTruthy()
```
LeandraH, line 20: This isn't actually checking for the description though, it's checking for the container. You should be able to get the description with `getByText(getLabels().general.camera.noAuthorization.description)` (for which you might want to set a well-named variable).

src/components/__tests__/NotAuthorisedView.spec.tsx
```diff
@@ -0,0 +1,38 @@
...
+  })
+  it('should successfully go back', () => {
```
LeandraH, line 22: I'd add a new line before every `it`, just to improve readability.

src/components/__tests__/NotAuthorisedView.spec.tsx
```diff
@@ -0,0 +1,38 @@
...
+    const { getByText } = render(
+      <NotAuthorisedView setVisible={setVisible} description={getLabels().general.camera.noAuthorization.description} />
+    )
+    const buttonVisible = getByText(getLabels().general.back)
+    fireEvent.press(buttonVisible)
```
LeandraH, line 27: This test doesn't have an `expect` statement. Does it run? Can it fail?

LeandraH (commented):
Very nice! Just a few small things :)

src/components/__tests__/NotAuthorisedView.spec.tsx
lunars97: I was not sure about it as well. I will fix it, thanks

src/components/__tests__/NotAuthorisedView.spec.tsx
lunars97: it ran without error, I will add **expect**

src/components/__tests__/NotAuthorisedView.spec.tsx
```diff
@@ -0,0 +1,43 @@
...
+}))
+
+describe('NotAuthorisedView', () => {
+  const setVisible = jest.fn()
+
+  it('should render description', () => {
+    const { getByText } = render(
+      <NotAuthorisedView setVisible={setVisible} description={getLabels().general.camera.noAuthorization.description} />
+    )
+    const isNoAuthDescription = getByText(getLabels().general.camera.noAuthorization.description)
```
LeandraH, line 21: Sorry, I meant that you can set the variable before the render statement because you use `getLabels().general.camera.noAuthorization.description` twice. But it's also okay not to, it's only used twice here :)

LeandraH (approved):
Nicely done!

f1sh1918 (approved):
nice work :)

PR #410, opened 2023-01-23
LUN-467: Playstore issues

SteffiStuffel:


steffenkleinle (approved):
Can you explain why you changed it first and now reverted it?

SteffiStuffel:
> Can you explain why you changed it first and now reverted it?

Docs say ```
This element sets whether the activity can be launched by components of other applications:
    If "true", the activity is accessible to any app, and is launchable by its exact class name.
    If "false", the activity can be launched only by components of the same application, applications with the same user ID, or privileged system components. This is the default value when there are no intent filters.
```

So the app cannot be started from the play store now.

LeandraH (approved):
Cool, thanks for the explanation, and figuring that out :)

PR #412, opened 2023-01-27
LUN-464: Fixed ExerciseScreen to lock level 2 and level 3 if level 1 …

lunars97:
…has no scores as well as renamed some functions to make it more understandable


src/services/__tests__/helpers.spec.ts
```diff
@@ -113,8 +113,8 @@ describe('helpers', () => {
       expect(willNextExerciseUnlock(undefined, SCORE_THRESHOLD_UNLOCK + 1)).toBeTruthy()
     })
 
-    it('should unlock if exercise is done good enough for the scond time', () => {
-      expect(willNextExerciseUnlock(SCORE_THRESHOLD_UNLOCK, SCORE_THRESHOLD_UNLOCK + 1)).toBeTruthy()
+    it('should unlock if exercise is done good enough for the second time', () => {
+      expect(willNextExerciseUnlock(SCORE_THRESHOLD_UNLOCK, SCORE_THRESHOLD_UNLOCK + 2)).toBeTruthy()
```
SteffiStuffel, line 117: Thanks for fixing these typos. Why have you changed this from ` + 1 ` to ` +  2`. It is not needed to change this, but it is fine.

src/services/helpers.ts
```diff
@@ -74,15 +75,22 @@ export const shuffleArray = <T>(array: T[]): T[] => {
   return shuffled
 }
 
-const getDoneExercisesByProgress = (disciplineId: number, progress: Progress): number => {
+const getDoneExercisesNumberInProgressByDiscipline = (disciplineId: number, progress: Progress): number => {
```
SteffiStuffel, line 78: I think a good naming would be: `getWellEnoughDoneExercises`

src/services/helpers.ts
```diff
@@ -74,15 +75,22 @@ export const shuffleArray = <T>(array: T[]): T[] => {
...
   const progressOfDiscipline = progress[disciplineId]
+  const exerciseKeyForWordChoice = ExerciseKeys.wordChoiceExercise
   return progressOfDiscipline
-    ? Object.keys(progressOfDiscipline).filter(item => progressOfDiscipline[item] !== undefined).length
+    ? Object.keys(progressOfDiscipline).filter(item => {
```
SteffiStuffel, line 82: Your function here does not really make sense:
1) You are iterating over the Object keys of the progress of on discipline. So you are iterating over the exercise keys: 0, 1, 2, so `item.length` is always undefined.
2) You are comparing the score of a exercise with `ExerciseKeys.wordChoiceExercise`. This does not make sense, as a score is "how good a exericse is done" and the ExerciseKey is "which exercise is done". These two cannot be compared.

What you want to do is: Filter and use only those entries, where either the ExerciseKey is vocabularyList OR the score is bigger than the threshold.

SteffiStuffel (changes requested):

src/services/helpers.ts
LeandraH: @lunars97 Do you want help with implementing the suggested function?

src/services/helpers.ts
```diff
@@ -98,7 +104,7 @@ export const getNextExercise = async (profession: Discipline): Promise<NextExerc
   }
   const progress = await getExerciseProgress()
   const firstUnfinishedDisciplineId = leafDisciplineIds.find(
-    id => getDoneExercisesByProgress(id, progress) < EXERCISES.length
+    id => getWellEnoughDoneExercises(id, progress) < EXERCISES.length
   )
 
   if (!firstUnfinishedDisciplineId) {
```
SteffiStuffel, line 110: Blow in line 123 you also need to adjust the line 
`const nextExerciseKey = EXERCISES.find(exercise => disciplineProgress[exercise.key] === undefined)`

to use your new funtion: `const nextExerciseKey = getWellEnoughDoneExercises(firstUnfinishedDisciplineId, progress)`. Otherwise here it would be ignored if the exercise is done well enough.

src/services/__tests__/helpers.spec.ts
```diff
@@ -51,9 +51,8 @@ describe('helpers', () => {
     it('should open first exercise of second discipline, if first discipline was finished yet', async () => {
       mocked(loadDiscipline).mockReturnValueOnce(Promise.resolve(mockDisciplines(true)[0]))
```
SteffiStuffel, line 52: Some of these tests here, that you adjusted, do not what they should do. Read the title carefully, what the test should check, and the check if the testdata matches the expected output.

SteffiStuffel (changes requested):

src/services/__tests__/helpers.spec.ts
```diff
@@ -29,7 +29,7 @@ describe('helpers', () => {
       mocked(getExerciseProgress).mockReturnValueOnce(Promise.resolve({ '10': { '0': 1, '1': 1 } }))
       const { disciplineId, exerciseKey } = await getNextExerciseWithCheck()
       expect(disciplineId).toBe(mockDisciplines()[0].leafDisciplines![0])
-      expect(exerciseKey).toBe(2)
+      expect(exerciseKey).toBe(1)
```
SteffiStuffel, line 32: The title of this test is "should open third exercise of first discipline, if two exercise were finished yet", but it returned 1, not as before 2. This is because the two exercieses, that are finished, are not finished good enough, there score is only 1, as you can see in Line 29. So if you want to fix this test correctly, you need to change back the expected result to 2, and the score of the done exercises above the threshold.

src/services/__tests__/helpers.spec.ts
```diff
@@ -40,21 +40,20 @@ describe('helpers', () => {
...
     })
 
-    it('should open third exercise of first discipline, if three exercises were finished yet', async () => {
+    it('should open third exercise of first discipline, if two exercises were finished yet', async () => {
       mocked(loadDiscipline).mockReturnValueOnce(Promise.resolve(mockDisciplines(true)[0]))
       mocked(getExerciseProgress).mockReturnValueOnce(Promise.resolve({ '10': { '0': 1, '1': 1, '2': 1 } }))
       const { disciplineId, exerciseKey } = await getNextExerciseWithCheck()
       expect(disciplineId).toBe(mockDisciplines()[0].leafDisciplines![0])
-      expect(exerciseKey).toBe(3)
+      expect(exerciseKey).toBe(1)
```
SteffiStuffel, line 48: Similar problem as with the test above: In the test data (line 45) 3 exercises are done, but not well enough.
Adjust the test data and change this and the title back to what is was before.

src/services/__tests__/helpers.spec.ts
```diff
@@ -40,21 +40,20 @@ describe('helpers', () => {
...
     })
 
     it('should open first exercise of second discipline, if first discipline was finished yet', async () => {
       mocked(loadDiscipline).mockReturnValueOnce(Promise.resolve(mockDisciplines(true)[0]))
       mocked(getExerciseProgress).mockReturnValueOnce(Promise.resolve({ '10': { '0': 1, '1': 1, '2': 1, '3': 1 } }))
-
       const { disciplineId, exerciseKey } = await getNextExerciseWithCheck()
-      expect(disciplineId).toBe(mockDisciplines()[0].leafDisciplines![1])
-      expect(exerciseKey).toBe(0)
+      expect(disciplineId).toBe(mockDisciplines()[0].leafDisciplines![0])
```
SteffiStuffel, line 55: Same here, adjust test data and undo the changes to the expected results.

SteffiStuffel (changes requested):
Nice work. Code looks good, also tested on Android emulator. :rocket: 

Just the test file is not ready yet, see my comments. Please also add a test case "Should open second exercise, if first was done well enough, but second was not done well enough."

SteffiStuffel (approved):
Looks good, nice work. :tada: 

Please delete the also the second branch you created for this ticket. You can do this when you open the lunes-app project in github -> click on branches and then select delete.

Tested on Android emulator.

LeandraH (approved):
Nice work, tested on iOS emulator

PR #415, opened 2023-01-31
LUN-xxx: Use Lunes emoji for Mattermost notifications

timoludwig:
I strongly believe that it is extremely important to always use the appropriate emojis on all occasions.


SteffiStuffel (approved):
Nice, thank you :)

sarahsporck (approved):
Nice 😅

PR #417, opened 2023-02-06
LUN-435: Add user created vocabulary to dictionary

SteffiStuffel:


src/hooks/useGetAllWords.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const getAllWords = async (): Promise<VocabularyItem[]> => {
+  const lunesStandardVocabulary = loadAllVocabularyItems()
+  const userVocabulary = getUserVocabularyItems()
+  let response: VocabularyItem[] = []
+  return Promise.all([lunesStandardVocabulary, userVocabulary]).then(values => {
+    response = response.concat(values[0])
+    response = response.concat(values[1])
+    return response
+  })
```
f1sh1918, line 14:
```suggestion
 return Promise.all([lunesStandardVocabulary, userVocabulary]).then(values => [...values[0],...values[1]])
```

f1sh1918: you can use es6 spread operator

f1sh1918 (approved):
just small refactoring found. nice work. Tested on ios

src/hooks/useGetAllWords.ts
SteffiStuffel: Good point, thanks.

LeandraH (approved):
Nice work, tested on iOS emulator

PR #421, opened 2023-02-21
LUN-465: Remove react native snap carousel

SteffiStuffel:


LeandraH (approved):
Nice, tested on iOS emulator

src/components/ImageCarousel.tsx
```diff
@@ -60,13 +76,10 @@ const ImageCarousel = ({ images, minimized = false }: ImageCarouselProps): React
...
-        <Pagination
-          activeDotIndex={currentIndex - 1}
-          dotsLength={allSize}
-          dotStyle={{ backgroundColor: theme.colors.primary }}
-        />
-      )}
+      {!!allSize &&
+        !!currentIndex &&
+        allSize > 1 &&
+        imagesUrls.map((item, index) => (index + 1 === currentIndex ? <ActiveDot /> : <Dot />))}
```
f1sh1918, line 82:
```suggestion
        imagesUrls.map((item, index) => (index + 1 === currentIndex ? <ActiveDot key={item.url}/> : <Dot key={item.url}/>))}
```

f1sh1918: missing keys

f1sh1918 (approved):
nice refactoring. Just keys are missing.
I wonder if the design has some ideas to display the indicators a bit smarter. I think they are getting lost

SteffiStuffel:
> I wonder if the design has some ideas to display the indicators a bit smarter. I think they are getting lost

Sure there are better options, but I think this is out of scope. Currently we have way more important topics, where we can improve lunes, so i would just keep it, open no tickets or so for now. But if you think it is important now, feel free to open a design ticket.

PR #422, opened 2023-02-22
LUN-455: Changed sort sequence to distinguish words based on substantives not …

lunars97:
…adjectives as well as it does not distinguish between Umlaute and basic letters anymore


src/services/helpers.ts
```diff
@@ -194,12 +194,19 @@ export const getSortedAndFilteredVocabularyItems = (
   vocabularyItems: VocabularyItem[] | null,
   searchString: string
 ): VocabularyItem[] => {
+  const collator = new Intl.Collator('de-De', { sensitivity: 'base', usage: 'sort' })
+
   const normalizedSearchString = normalizeSearchString(searchString)
 
+  const getNouns = (item: { word: string }) => {
```
SteffiStuffel, line 201: You can simplify the function declaration to `const getNouns = (word: string): string => ...`

src/services/helpers.ts
```diff
@@ -194,12 +194,19 @@ export const getSortedAndFilteredVocabularyItems = (
...
+    const words = item.word.split(' ')
+    return words.find((word: string) => word[0] === word[0].toUpperCase()) as string
```
SteffiStuffel, line 203: This `as string` cast is needed, because if no word matches your find function the function retruns nothing. It would be better to do:
`return words.find((word: string) => word[0] === word[0].toUpperCase()) ?? word`
as if there is some error or word without a noun in the words, e.g. "der helm", it would just return "helm" for sorting instead of nothing. This should not happen, but it is possible, so we should also think of this case.

src/services/helpers.ts
```diff
@@ -194,12 +194,19 @@ export const getSortedAndFilteredVocabularyItems = (
...
+  }
+
   const filteredVocabularyItems = vocabularyItems?.filter(
     item => item.word.toLowerCase().includes(normalizedSearchString) || matchAlternative(item, normalizedSearchString)
   )
-  return filteredVocabularyItems?.sort((a, b) => a.word.localeCompare(b.word)) ?? []
+  return filteredVocabularyItems?.sort((a, b) => collator.compare(getNouns(a), getNouns(b))) ?? []
```
SteffiStuffel, line 209: If you changed the function declaration above, you can just call the function here with `getNouns(a.word)`

android/app/build.gradle
```diff
@@ -111,7 +111,7 @@ def enableProguardInReleaseBuilds = false
  * give correct results when using with locales other than en-US.  Note that
  * this variant is about 6MiB larger per architecture than default.
  */
-def jscFlavor = 'org.webkit:android-jsc:+'
+def jscFlavor = 'org.webkit:android-jsc-intl:+'
```
SteffiStuffel, line 114: Can you please explain what this is / does?

SteffiStuffel (commented):
Tested on browserstack android and ios, works as expected.

android/app/build.gradle
lunars97: It is needed to use **Intl.collator** properly so that we could sort based on character-specific letters (as I understood it makes it more internalized) Here is a link:  https://stackoverflow.com/questions/56943813/using-intl-properly-in-android-react-native-app

src/services/helpers.ts
lunars97: I have tried this way, but it was giving off error

lunars97: I have tried to write this way, but it gives off an error

f1sh1918: you can just pass a `word` to this function when you call it  instead of an `item` , since you only need the word(string) and not the entire `item`
Then steffis hint work

PR #424, opened 2023-02-23
LUN-455: Changed sort sequence based on substantives and it does not …

lunars97:
…distinguish between umlaut and basic letters anymore


LeandraH (approved):
Very nice, good find on the Intl.Collator.compare! Tested on real iOS

SteffiStuffel (approved):
Looks good. Really nice work, also a nice clean git history with only one commit. Well done :clap:

PR #425, opened 2023-02-27
LUN-464: Fixed exercise screen to lock levels if score threshold unlo…

lunars97:
…ck is not above as well as renamed some functions to make them more clear


SteffiStuffel (commented):
Why have you closed the old PR and opend a new one? Now it is necessary to review everything again. Can you please reactivate the old branch and PR, so we can continue work and discuss on this or tell why it was necessary to open a new one?

PR #426, opened 2023-03-01
LUN-423: Upgrade react native to 0.70.7

sarahsporck:


android/app/build.gradle
```diff
@@ -202,13 +191,13 @@ android {
             // Due to a bug inside AGP, we have to explicitly set a dependency
             // between configureNdkBuild* tasks and the preBuild tasks.
             // This can be removed once this is solved: https://issuetracker.google.com/issues/207403732
```
LeandraH, line 193: Can this comment be removed now?

android/app/src/main/jni/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+cmake_minimum_required(VERSION 3.13)
+
+# Define the library name here.
+project(rndiffapp_appmodules)
```
LeandraH, line 4:
```suggestion
project(lunes_appmodules)
```

LeandraH (approved):
Very nice, tested on a real iOS device

android/app/build.gradle
sarahsporck: Can probably be removed when upgrading the android gradle plugin to 7.3+
but has to stay for now

SteffiStuffel (approved):
Nice work, thank you :pray:

PR #432, opened 2023-03-06
LUN-472: Dependabot vulnerability alerts

LeandraH:
This pull request belongs to https://issues.tuerantuer.org/browse/LUN-472. It fixes the vulnerability alerts that Dependabot can't fix by itself.


SteffiStuffel (approved):
Nice. Tested on Andoird and ios browserstack.

sarahsporck (approved):
Tested on Android.
Tested on iOS simulator.

PR #436, opened 2023-03-28
LUN-xx: Fix release notes CICD

SteffiStuffel:


LeandraH (approved):
Looks reasonable but I don't understand what the problem is. Is this connected to the deployments failing yesterday?

SteffiStuffel:
> Looks reasonable but I don't understand what the problem is. Is this connected to the deployments failing yesterday?

No not related to failing deployments. I relates to the yarn updates and fixes the correct display of the release notes.

PR #438, opened 2023-04-12
LUN-480: Fix filtering with special characters

lunars97:

### DIFF
--- a/src/services/__tests__/helpers.spec.ts
+++ b/src/services/__tests__/helpers.spec.ts
@@ -1,10 +1,22 @@
-import { NextExercise, SCORE_THRESHOLD_UNLOCK, SimpleResult } from '../../constants/data'
+import {
+  ARTICLES,
+  NextExercise,
+  SCORE_THRESHOLD_UNLOCK,
+  SimpleResult,
+  VOCABULARY_ITEM_TYPES,
+} from '../../constants/data'
 import { loadDiscipline } from '../../hooks/useLoadDiscipline'
 import { VocabularyItemResult } from '../../navigation/NavigationTypes'
 import VocabularyItemBuilder from '../../testing/VocabularyItemBuilder'
 import { mockDisciplines } from '../../testing/mockDiscipline'
 import { getExerciseProgress } from '../AsyncStorage'
-import { calculateScore, getNextExercise, getProgress, willNextExerciseUnlock } from '../helpers'
+import {
+  calculateScore,
+  getNextExercise,
+  getProgress,
+  getSortedAndFilteredVocabularyItems,
+  willNextExerciseUnlock,
+} from '../helpers'
 
 import mocked = jest.mocked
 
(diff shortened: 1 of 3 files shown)


src/services/helpers.ts
```diff
@@ -182,27 +180,26 @@ export const calculateScore = (vocabularyItemsWithResults: VocabularyItemResult[
...
 
 export const getSortedAndFilteredVocabularyItems = (
   vocabularyItems: VocabularyItem[] | null,
   searchString: string
 ): VocabularyItem[] => {
-  const collator = new Intl.Collator('de-De', { sensitivity: 'base', usage: 'sort' })
+  const collator = new Intl.Collator('de-De', { sensitivity: 'base', usage: 'sort', localeMatcher: 'lookup' })
 
-  const normalizedSearchString = normalizeSearchString(searchString)
+  const normalizedSearchString = searchString
```
SteffiStuffel, line 199: By changing this like that, we will have problems: e.g. when entering 
"Die Ölk" the result should be "Die Ölkanne", 
when entering "große Ölkanne", the result should be empty.

In the previous `normalizeSearchString` function this was working properly, as we there only removed the first word from the searchString if it is an Article, see:
```
  const searchStringWithoutArticle = ARTICLES.map(article => article.value).includes(
    searchString.split(' ')[0].toLowerCase()
  )
    ? searchString.substring(searchString.indexOf(' ') + 1)
    : searchString
```
If the first word is an article we remove it from the search string, if the first word is something else (not an article) we keep it in the search string.

This behaviour should not be changes by this PR.

SteffiStuffel (commented):
Nice work, looks good :rocket: 

Just one small hint: Please squash the commits together into one. If you need help with this, we can have a look at it together today.

src/services/helpers.ts
```diff
@@ -1,5 +1,4 @@
 import { AxiosResponse } from 'axios'
-import normalizeStrings from 'normalize-strings'
```
LeandraH, line 2: It looks to me like we can then remove the package 'normalize-strings' from our code. Can you check if that is indeed the case?

lunars97: yes, we don't use it anymore. I will remove it.

lunars97: seems like the changes related to  package removal disappeared after committing it. There is no package on my local repository

.circleci/config.yml
```diff
@@ -583,48 +583,7 @@ parameters:
         type: boolean
 version: 2.1
 workflows:
-    browserstack_delivery:
```
SteffiStuffel, line 586: Please undo the changes in this file.

SteffiStuffel (commented):
It is not working now. If entering `ölk...` i get no result.

Edit: Also not working on main, think there is something weird with my emulator.

steffenkleinle (changes requested):
I think the current solution is not perfect. It prevents users from finding words including special characters using the "normalized" character.
Example: Finding `der Abhänger` with the filter query `Abha` is not working, only the use of `Abhä` works. In my opinion both should work though.

src/services/helpers.ts
```diff
@@ -210,7 +209,12 @@ export const getSortedAndFilteredVocabularyItems = (
   }
 
   const filteredVocabularyItems = vocabularyItems?.filter(
-    item => item.word.toLowerCase().includes(normalizedSearchString) || matchAlternative(item, normalizedSearchString)
+    item =>
+      item.word
+        .toLowerCase()
+        .normalize('NFD')
+        .replace(/[\u0300-\u036f]/g, '')
```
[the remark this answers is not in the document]
lunars97: I just looked for solutions and stumbled upon normalize() method [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize](url) It basically decomposes combined characters into their base character and accent marks ( for ex., ä,ö, ü). The regex `/[\u0300-\u036f]/g` is used to match and remove accent marks, tildes, and other symbols that modify characters.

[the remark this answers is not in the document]
lunars97: I thought that it might only be need on search query if someone types basic characters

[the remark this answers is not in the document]
SteffiStuffel: Please put the regex into a variable with a speaking name, so everyone reading this code, knows what is happening here.

SteffiStuffel (commented):
Please add tests for this function, it should test:
1) Ordering of the words in the list is correct. Words starting with special chars should not be placed at the end.
2) Words with ö,ä,ü are found correctly when searching for both "ö,ü,a" and "o,u,a"
3) Words with "o,a,u" are not found, when searching for "ö,ä,ü"

src/services/helpers.ts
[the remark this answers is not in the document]
steffenkleinle: > Please add tests for this function, it should test:
> 
>     1. Ordering of the words in the list is correct. Words starting with special chars should not be placed at the end.
> 
>     2. Words with ö,ä,ü are found correctly when searching for both "ö,ü,a" and "o,u,a"
> 
>     3. Words with "o,a,u" are not found, when searching for "ö,ä,ü"

Tbh I disagree with 3., if you type the normal chars (a, o, u) also the special chars (ä, ö, ü) should be found. That is also the reason why I personally would rather just keep the library...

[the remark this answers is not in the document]
steffenkleinle: There are also other edge cases like `'`, ß...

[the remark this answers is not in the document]
lunars97: should I revert everything back as it was with the library?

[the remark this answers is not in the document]
steffenkleinle: I would use normalize strings on both the search string/query AND every word we are comparing. I think that is the easiest, most thorough and covers all the edge cases. If you or @SteffiStuffel would rather use a custom regex or some other way, that is completely fine for me as well :)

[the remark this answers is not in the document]
SteffiStuffel: I also think using a library is better than custom regex :+1:

src/services/__tests__/helpers.spec.ts
```diff
@@ -167,4 +179,148 @@ describe('helpers', () => {
...
+        {
+          id: 10,
+          type: VOCABULARY_ITEM_TYPES.lunesStandard,
+          word: 'Untergrund',
+          article: ARTICLES[1],
+          audio: '',
+          images: [{ id: 2, image: 'image' }],
+          alternatives: [],
+        },
+      ]
```
steffenkleinle, line 321: Can't you just get those from the VocabularyItemBuilder?

steffenkleinle (approved):
Well done :rocket:

src/services/__tests__/helpers.spec.ts
lunars97: @steffenkleinle I wanted to compare this data with the one which was created by the **VocabularyItemBuilder**

steffenkleinle: But basically you are just checking that `getSortedAndFilteredVocabularyItems` doesn't change the input, right? Or am I missing something?

lunars97: @steffenkleinle I am checking the correct order of the list whether the umlaut characters are not at the very end of the list

steffenkleinle: But you are doing
``` 
      const sortedTestData = getSortedAndFilteredVocabularyItems(sortedData, '')
      expect(sortedData).toEqual(sortedTestData)
```
which checks for equality of `sortedData` and `sortedTestData`. So actually you are just making sure that `sortedData` is still the same after calling `getSortedAndFilteredVocabularyItems` on it. Am I missing something?

steffenkleinle (approved):

src/services/__tests__/helpers.spec.ts
lunars97: @steffenkleinle you are right. I am comparing unsorted list with sorted

steffenkleinle: However, the naming does not represent this, both is called "sorted", not "unsorted". And if one is sorted and one is unsorted, it should not be equal after sorting, should it?

steffenkleinle: You should pass an unsorted list as params though then, not the already sorted list

lunars97: @steffenkleinle you are right. Now I understand what I did here is a illogical. Thank you:)

sarahsporck (approved):

sarahsporck:
Tested on simulated ios :)

src/services/__tests__/helpers.spec.ts
steffenkleinle: Good to merge :rocket:

PR #439, opened 2023-04-17
LUN-488: Renamed correctly the headline of the user vocabulary editin…

lunars97:
…g screen


LeandraH (approved):
Very nice, tested in simulated iOS

f1sh1918 (approved):
works as expected! Nice :)

PR #440, opened 2023-04-18
LUN-483: Fixed header button back to word list on word list screen

lunars97:

### DIFF
--- a/src/components/ExerciseHeader.tsx
+++ b/src/components/ExerciseHeader.tsx
@@ -65,17 +65,11 @@ const ExerciseHeader = ({
   const theme = useTheme()
   const showProgress = numberOfWords !== undefined && numberOfWords > 0 && currentWord !== undefined
   const progressText = showProgress ? `${currentWord + 1} / ${numberOfWords}` : ''
-  const isWordList = exerciseKey === 0 || exerciseKey === undefined
-  const isLastWordInList = isWordList && numberOfWords === (currentWord ?? 0) + 1
-  const title =
-    labelOverride ?? isLastWordInList
-      ? getLabels().results.action.backToWordlist
-      : getLabels().general.header.cancelExercise
 
   useEffect(() => {
     const renderHeaderLeft = () => (
       <NavigationHeaderLeft
-        title={title}
+        title={getLabels().general.header.cancelExercise}
         onPress={confirmClose ? () => setIsModalVisible(true) : () => navigation.dispatch(closeExerciseAction)}
         isCloseButton={isCloseButton}
       />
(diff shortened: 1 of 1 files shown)


src/components/ExerciseHeader.tsx
```diff
@@ -73,12 +73,16 @@ const ExerciseHeader = ({
       : getLabels().general.header.cancelExercise
 
   useEffect(() => {
+    const handleOnPress = () => {
+      if (!confirmClose && isLastWordInList) {
```
LeandraH, line 77: I'm not sure we still need the `!confirmClose`. Can you explain its role here to me?

LeandraH (approved):
Nice!

src/components/ExerciseHeader.tsx
lunars97: Thank you @LeandraH for your review and question. I also think it does nothing here. I think because of this previous line of code was not working properly. I think I can convert my changes into one line of code like it was before only replacing `!confirmClose` with `isLastWordInList`

LeandraH: Hmmm, no, when you look at the props of the ExerciseHeader, we assign the default `confirmClose = true` in line 58.

Before this PR, there were two possible cases here of what happens after a user taps on the close button: Either we ask if the user is sure that they want to close the exercise, or it closes immediately. In both cases, the user is sent to the overview of the levels, either straight away or after confirming. What we want to add in this PR is a third scenario where the user is not sent to the overview of the levels but to the word list. So we need to add a possible outcome, not replace one.

SteffiStuffel (changes requested):
Thanks for your PR, this is the correct place to start working on this bug :+1: 

Though there are some problems with your suggested solution:
1) When we open the word list and click on a word (= open the detail view) and then in the header click the close icon, we expect that there should be no confirmation modal, but with your changes now the confirmation modal opens. 
2) The confirmation modal should still be visible for all other exercises, when closing them. Now the confirmation modal is not shown for all other exercises. 
3) You can not replace the `dispatch(closeExerciseAction)` with `navigation.goBack()`. In a lot of places this will lead to the correct solutions, but for example do the following:
Go to an exercise -> complete it sucessfully -> at the result screen click on "Next exercise" and the in this second exercise click on the header-close-icon. Then we will see the last screen, as this what "goBack" does, but this last screen is the result screen and this is wrong, when this second exercise is cancelt it should go back to the screen where the first exercise was started from, e.g. ExercisesScreen or HomeScreen.

I think there are two solutions for this task:
1) The simpler (but not that nice one): Just rename the header for the last word, so it always says "Übung beenden" for every word, also the last one.
2) The nicer solution would be to adjust the behavior of the close button in the header, so when in an Detail-View of a word, a click on "Übung beenden" always goes back to the word-list.

You can choose which one you prefer to do.

SteffiStuffel (approved):
Looks good. Tested on Android emulator.

PR #445, opened 2023-05-15
LUN-XXX: Update Hermes to Match RN

LeandraH:
This should fix the build that keeps failing because of incompatible Hermes versions

https://app.circleci.com/pipelines/github/digitalfabrik/lunes-app/3255/workflows/31778c16-eb63-44f5-8b8e-f064134f0940/jobs/11257


LeandraH:
https://github.com/digitalfabrik/lunes-app/pull/446 is more thorough :)

PR #446, opened 2023-05-15
LUN-XXX: Update cocoapods hermes fix

f1sh1918:
- updated cocoapods
- fixed hermes
- updated ruby gems


LeandraH (approved):
Nice, thanks!

steffenkleinle (approved):
Not tested.

I think we should be careful with creating PRs without issues. Bigger stuff should always have a reference and therefore an issue associated. This PR is in maybe still okay though.

PR #447, opened 2023-05-15
LUN-xxx: Update copyright year

steffenkleinle:

### DIFF
--- a/src/routes/home/components/HomeFooter.tsx
+++ b/src/routes/home/components/HomeFooter.tsx
@@ -20,7 +20,7 @@ type HomeFooterProps = {
 
 const HomeFooter = ({ navigateToImprint }: HomeFooterProps): ReactElement => (
   <Container>
-    <ContentSecondary>{'\u00A9'}LUNES2022</ContentSecondary>
+    <ContentSecondary>{'\u00A9'}LUNES2023</ContentSecondary>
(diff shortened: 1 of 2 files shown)


src/routes/home/components/HomeFooter.tsx
```diff
@@ -20,7 +20,7 @@ type HomeFooterProps = {
 
 const HomeFooter = ({ navigateToImprint }: HomeFooterProps): ReactElement => (
   <Container>
-    <ContentSecondary>{'\u00A9'}LUNES2022</ContentSecondary>
+    <ContentSecondary>{'\u00A9'}LUNES2023</ContentSecondary>
```
LeandraH, line 23: Maybe we can use `new Date().getFullYear()` instead so we don't have to update it every year.

LeandraH (approved):

src/routes/home/components/HomeFooter.tsx
steffenkleinle: Thought about that as well, but didn't do it bc then in theory users can set their phone to some other year and maybe claim smth? But I am not an expert here if that would be a problem.

LeandraH: Oooooh, interesting point. No idea 🤷

sarahsporck (approved):

PR #450, opened 2023-06-13
LUN-497: Rename Lexikon to Suche

LeandraH:
This pull request belongs to the issue https://issues.tuerantuer.org/browse/LUN-497 . I decided to also change the code naming from dictionary to search, mostly to not confuse the next developers about why it's called dictionary.


SteffiStuffel (changes requested):
I think a renaming of a button should not lead to a renaming in our code. I think it is just fine, if we still call this Dictionary. Please undo. 

(Renaming of icons makes sense.)

SteffiStuffel (approved):
Looks good, tested on Android Emulator

SteffiStuffel:
Please squash the commit when merging.

PR #453, opened 2023-06-21
LUN-476: Add long term exercise overview screen

lunars97:

**The text that should be displayed in the info icon modal. I included it on another cleanup ticket while it has not been yet completed by the UX/UI team. Here is** [the link](https://issues.tuerantuer.org/browse/LUN-478)


src/navigation/BottomTabNavigator.tsx
```diff
@@ -39,6 +42,13 @@ const BottomTabNavigator = (): JSX.Element | null => {
   const renderDictionaryTabIcon = ({ focused }: { focused: boolean }) =>
     focused ? <BookIconWhite width={iconSize} height={iconSize} /> : <BookIconGrey width={iconSize} height={iconSize} />
 
+  const renderRepetitionTabIcon = ({ focused }: { focused: boolean }) =>
```
sarahsporck, line 45: Another tab Icon? I do not blame you as you did not do the design, but I think long term we should reconstruct this. The tab text is always cut off. :/ What do you think?

sarahsporck (approved):
Tested in ios simulator

src/constants/labels.json
```diff
@@ -183,6 +184,12 @@
   "dictionary": {
     "noResults": "Es wurden leider keine Ergebnisse gefunden."
   },
+  "repetition": {
+    "repeatWords": "Wörter wiederholen",
+    "repeatNow": "Jetzt wiederholen",
+    "wordsToRepeat": "Wörter zum wiederholen bereit",
```
steffenkleinle, line 190:
```suggestion
    "wordsToRepeat": "Wörter zum Wiederholen bereit",
```

src/navigation/RepetitionStackNavigator.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const options = screenOptions(useTabletHeaderHeight())
+  const { back } = getLabels().general
+
+  return (
+    <Stack.Navigator>
+      <Stack.Screen
+        name='Repetition'
+        component={RepetitionScreen}
+        options={({ navigation }) => options(back, navigation)}
+      />
```
steffenkleinle, line 21: If we just have one screen inside the navigator, it should not be necessary to wrap it in another StackNavigator. Or is it planned to add more screens here later on?

steffenkleinle (changes requested):
Tested, works as expected I guess, but we should not merge this in the current state. If we want to merge this already, we should hide the repetition tab in the bottom navigation bar (e.g. by commenting them out).

Also agreeing with @sarahsporck that it is too many tab icons, the text is cut for me as well (both Wiederholen and Meine Wörter).

src/navigation/BottomTabNavigator.tsx
sarahsporck: Yes I think it's because my emulators screen is a little bit narrow. I also just left the comment to keep track of this. :)

src/navigation/RepetitionStackNavigator.tsx
sarahsporck: I don't think so. Check out the design on Figma: https://www.figma.com/file/4bW33nU6zZ6ySeTqrM37Jq/Lunes_Master?type=design&node-id=301-2481&mode=design&t=D20ByDd0SMUTdbAB-0

lunars97: @sarahsporck  should I better comment out the tab on navigation?

src/navigation/BottomTabNavigator.tsx
steffenkleinle: Happens on my real device as well though 🙈

src/navigation/RepetitionStackNavigator.tsx
sarahsporck: No that's not neccessary. If even, we'll do this in a separate branch. I think @steffenkleinle wanted to point out, that as this route does not have any more screens. Therefore, you can directly put the `RepetitionScreen` into the `RepetitionTab` in the `BottomTabNavigator`, then you could get rid of the `RepetitonStackNavigator`.

steffenkleinle: I think we should comment out the tab, otherwise we can't release anymore bc this screen is not yet finished.

steffenkleinle (approved):

PR #454, opened 2023-06-29
LUN-520: Fix editing bug

LeandraH:
This pull request belongs to the issue https://issues.tuerantuer.org/browse/LUN-520.
The phone automatically saves the photo to its library already, we don't need to do that a second time with a new name. When editing, we were trying to save the picture a third time, with the same new name, which Android seems to ignore and iOS seems to not.

To test: I managed to reproduce this on my real iPhone, not on an emulated one.


src/routes/process-user-vocabulary/UserVocabularyProcessScreen.tsx
```diff
@@ -118,13 +118,7 @@ const UserVocabularyProcessScreen = ({ navigation, route }: UserVocabularyProces
...
       }
 
-      const imagePaths = await Promise.all(
-        images.map(async (image, index) => {
-          const path = `file:///${DocumentDirectoryPath}/image-${id}-${index}.jpg`
-          await moveFile(image, path)
-          return { id: index, image: path }
-        })
-      )
+      const imagePaths = await Promise.all(images.map(async (image, index) => ({ id: index, image })))
```
steffenkleinle, line 121: I think there is no need for a promise/async anymore

steffenkleinle (approved):
Works as expected, tested on android. Could you please check the linked issue in jira and reject if it is referring to the same issue?

src/routes/process-user-vocabulary/UserVocabularyProcessScreen.tsx
```diff
@@ -118,13 +118,7 @@ const UserVocabularyProcessScreen = ({ navigation, route }: UserVocabularyProces
...
       }
 
-      const imagePaths = await Promise.all(
-        images.map(async (image, index) => {
-          const path = `file:///${DocumentDirectoryPath}/image-${id}-${index}.jpg`
-          await moveFile(image, path)
-          return { id: index, image: path }
-        })
-      )
+      const imagePaths = images.map((image, index) => ({ id: index, image }))
```
SteffiStuffel, line 121: You cannot remove this part, as it is really important.
We move these images as when they are taken in the app, they are only saved to a cache, to persist them and see them later, we need to save them to a permanent path. You can see this error, when you create an own vocabulary, then close the app, clear the app cache and reopen the app, then the image of this word is not available. 
As this seems to be related to the sentry bug, you should investigate, why the id counter does not work properly and this bug occurs.

SteffiStuffel (changes requested):

SteffiStuffel (changes requested):
Now it is not possible to change the image of a word. Steps to reproduce:
create a word with an image
save
edit the image by removing the existing image and then adding a new one.
Result: Still the old image is displayed.

So we need to delete old images if they are removed.

SteffiStuffel (approved):
Almost there, we now need to delete images from file storage, if they are deleted by the app. Feel free to create a own ticket for this, if you do not want to do it in this PR.

LeandraH:
Sounds good, I'll do that separately since we aren't doing it so far.

PR #956, opened 2023-08-21
902: Repetition Exercise footer badge

SteffiStuffel:
### Short description

With this solution the footer only updates, when an icon is pressed. I think this is enough for now, as we otherwise would need to implement some sort of listener, which I think is out of scope.

### Resolved issues

Fixes: #902

---

### DIFF
--- a/src/navigation/BottomTabNavigator.tsx
+++ b/src/navigation/BottomTabNavigator.tsx
@@ -1,114 +1,117 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import {createBottomTabNavigator} from '@react-navigation/bottom-tabs'
-import React, {useState} from 'react'
-import {isTablet} from 'react-native-device-info'
-import {heightPercentageToDP as hp} from 'react-native-responsive-screen'
-import {useSafeAreaInsets} from 'react-native-safe-area-context'
-import {useTheme} from 'styled-components/native'
+import { createBottomTabNavigator } from '@react-navigation/bottom-tabs'
+import React, { useState } from 'react'
+import { isTablet } from 'react-native-device-info'
+import { heightPercentageToDP as hp } from 'react-native-responsive-screen'
+import { useSafeAreaInsets } from 'react-native-safe-area-context'
+import { useTheme } from 'styled-components/native'
 
 import {
-    HeartIconGrey,
-    HeartIconWhite,
-    HomeIconGrey,
-    HomeIconWhite,
-    MagnifierIconGrey,
-    MagnifierIconWhite,
-    StarIconGrey,
-    StarIconWhite,
-    RepeatIconGrey,
-    RepeatIconWhite,
+  HeartIconGrey,
+  HeartIconWhite,
+  HomeIconGrey,
+  HomeIconWhite,
+  MagnifierIconGrey,
+  MagnifierIconWhite,
+  StarIconGrey,
+  StarIconWhite,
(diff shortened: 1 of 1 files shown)


src/navigation/BottomTabNavigator.tsx
```diff
@@ -1,5 +1,6 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
SteffiStuffel, line 1: Added this, so not all import and stuff related to repetition exercise needs to be comment out, as long as we do not publish this feature, but only the relevant lines need to be commented out. Added a hint to the clean up ticket, that this comment needs to be removed.

src/navigation/BottomTabNavigator.tsx
```diff
@@ -92,11 +96,13 @@ const BottomTabNavigator = (): JSX.Element | null => {
         component={DictionaryStackNavigator}
         options={{ tabBarIcon: renderDictionaryTabIcon, title: getLabels().general.dictionary }}
       />
-      {/*      <Navigator.Screen
-        name='RepetitionTab'
-        component={RepetitionStackNavigator}
-        options={{ tabBarIcon: renderRepetitionTabIcon, title: getLabels().general.repetition }}
-      /> */}
+      {/* <Navigator.Screen */}
```
LeandraH, line 99: Gah, this makes it so much harder to uncomment again!

LeandraH (approved):
Great idea!

src/navigation/BottomTabNavigator.tsx
steffenkleinle: In IntelliJ you can just use a shortcut and it uncomments this automatically :) For me it is `CTRL + Shift + 7`.

steffenkleinle: This also means that atm this never updates, even if this changes, e.g. by doing the repetition exercise?

src/navigation/BottomTabNavigator.tsx
```diff
@@ -35,6 +35,12 @@ const BottomTabNavigator = (): JSX.Element | null => {
 
   const iconSize = hp('3.5%')
 
+  const [repetitionBadgeNumber, setRepetitionBadgeNumber] = useState<undefined | number>(undefined)
+
+  getNumberOfWordNodeCardsNeedingRepetitionWithUpperBound()
+    .then(result => setRepetitionBadgeNumber(result > 0 ? result : undefined))
+    .catch(() => setRepetitionBadgeNumber(undefined))
```
steffenkleinle, line 42: I think we should use our `useLoadAsync` hook for this for lifecycle stuff and correct handling of errors and such :) 

```suggestion
 const { data: repetitionBadgeNumber } = useLoadAsync(
    getNumberOfWordNodeCardsNeedingRepetitionWithUpperBound,
    undefined
  )
```
This would also make updating the number (e.g. after doing repetition exercise) later as we can just call `refresh`.

steffenkleinle (approved):
Tested on android

src/navigation/BottomTabNavigator.tsx
LeandraH: Hmm, my (un)commenting shortcuts only remove the outermost brackets, but I also use VS Code 🤷  It's also fine, just mildly annoying

steffenkleinle: IntelliJ :heart:

PR #989, opened 2023-08-28
907: Bar chart for repetition exercise

SteffiStuffel:
###Short description
- Added chart to long term memory exercise

### Proposed changes
- added victory chart library
- added function to fill long term memory exercise with test data
### Resolved issues

Fixes: #907


LeandraH (approved):
Nice! Tested on emulated iOS

src/constants/labels.json
```diff
@@ -188,7 +188,12 @@
     "repeatWords": "Wörter wiederholen",
     "repeatNow": "Jetzt wiederholen",
     "wordsToRepeat": "Wörter zum Wiederholen bereit",
-    "yourLearningProgress": "Dein Lernfortschritt"
+    "yourLearningProgress": "Dein Lernfortschritt",
+    "chart": {
+      "new_": "Neue",
```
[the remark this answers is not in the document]
SteffiStuffel: yes "new" is a keyword and cannot be used, but open for better ideas to name this.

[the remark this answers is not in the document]
LeandraH: image

Can't say I like any of these. Maybe recent? But new_ might be clearer...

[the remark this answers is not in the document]
steffenkleinle: But like even though it is a keyword, using it in strings should not be a problem right?

src/constants/labels.json
```diff
@@ -188,7 +188,12 @@
...
+      "learned": "Gelernte",
```
steffenkleinle, line 194: I think it would be better to also include the word `Wörter` here since it adds more context and the order might be different for other languages (if we should ever decide to translate the app).
```suggestion
      "learnedWords": "Gelernte Wörter",
```

src/constants/theme/colors.ts
```diff
@@ -24,6 +24,10 @@ export const COLORS = {
   link: 'blue',
   lightGreyBackground: '#e0e4ed',
   audioRecordingActive: '#97aef1',
+  chartColor1: '#E0E4ED',
```
steffenkleinle, line 27: We are starting to count at one? :smile:

src/routes/repetition/components/Chart.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { LineSegment, VictoryAxis, VictoryBar, VictoryChart, VictoryLabel } from 'victory-native'
+
+import theme from '../../../constants/theme'
+import useLoadAsync from '../../../hooks/useLoadAsync'
+import { getNumberOfWordsInEachSection, sections } from '../../../services/RepetitionService'
+import { getLabels } from '../../../services/helpers'
+
+const { chartColor1, chartColor2, chartColor3, chartColor4 } = theme.colors
+const barColors = [chartColor1, chartColor2, chartColor2, chartColor2, chartColor2, chartColor2, chartColor3]
+const Chart: () => ReactElement = () => {
```
steffenkleinle, line 13: Optional: If this is repetition specific this should also have a repetition specific name imo. While using this, one does not get what kind of chart this :see_no_evil:

src/routes/repetition/components/Chart.tsx
```diff
@@ -0,0 +1,68 @@
...
+    })
+  )
+
+  const { data, refresh } = useLoadAsync(getNumberOfWordsInEachSection, undefined)
+
+  useEffect(() => {
+    if (data) {
+      setChartData(data.map((item, index) => ({ y: item, x: index, fill: barColors[index] })))
+    }
+  }, [data])
```
steffenkleinle, line 30: I think this could be improved and simplified without the using of a `useState`, `useLoadAsync` and a `useEffect`:

```suggestion
  const getChartData = useCallback(async () => {
    const data = await getNumberOfWordsInEachSection()
    return data.map((item, index) => ({ y: item, x: index, fill: barColors[index] }))
  }, [])
  const { data, refresh } = useLoadAsync(getChartData, undefined)
  const chartData =
    data ??
    Array(sections.length).fill({
      y: 0,
      fill: chartColor2,
    })

```

src/routes/repetition/components/Chart.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  useFocusEffect(useCallback(() => refresh(), [refresh]))
```
steffenkleinle, line 32: Why not just
```suggestion
  useFocusEffect(refresh)
```

steffenkleinle (commented):
Tested, looks good. I just feel that the horizontal bar between the bars for Lernstufen 1 - 5 looks a little like a scroll bar but thats just from the design :+1:

src/constants/labels.json
SteffiStuffel: We need "gelernte" and "wörter" as two separte strings, as this is how the library works. I think is is easier to just define them here as it is and avoid using split functions.

src/routes/repetition/components/Chart.tsx
SteffiStuffel: Good idea :+1:

src/routes/repetition/components/Chart.tsx
SteffiStuffel: to avoid unneccessary rerenders. [Docs}(https://reactnavigation.org/docs/use-focus-effect/) say "Note: To avoid the running the effect too often, it's important to wrap the callback in useCallback before passing it to useFocusEffect as shown in the example."

steffenkleinle: Pretty sure refresh is already wrapped in useCallback, not sure if it is then still necessary

SteffiStuffel: yes it is.

src/routes/repetition/components/RepetitionProgressChart.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { LineSegment, VictoryAxis, VictoryBar, VictoryChart, VictoryLabel } from 'victory-native'
+
+import theme from '../../../constants/theme'
+import useLoadAsync from '../../../hooks/useLoadAsync'
+import { getNumberOfWordsInEachSection, sections } from '../../../services/RepetitionService'
+import { getLabels } from '../../../services/helpers'
+
+const { chartColor1, chartColor2, chartColor3, chartColor4 } = theme.colors
+const barColors = [chartColor1, chartColor2, chartColor2, chartColor2, chartColor2, chartColor2, chartColor3]
+const RepetitionProgressChart: () => ReactElement = () => {
```
f1sh1918, line 13: is there a reason not to use the standard react fc annotation?
```suggestion
const RepetitionProgressChart = (): ReactElement => {
```

f1sh1918 (approved):
Really nice implementation of the bar chart :)
Just some minor issues.

One thing i recognized at least on ios.
When i enabled devMode the ReptitionTab is not appearing.
I have to make a refresh in the terminal. Same for deactivating
Since its devMode this is a minor issue but maybe we can create a issue for that

And here is a margin missing, but that hasn't to be fixed in this pr
image

src/routes/repetition/components/RepetitionProgressChart.tsx
```diff
@@ -0,0 +1,71 @@
...
+              transform: 'scale(0.9 1)',
+            }}
+          />
+        }
+      />
+      <VictoryBar
+        animate={{
+          animationWhitelist: ['data', 'style'],
+        }}
+        labels={({ datum }) => String(Math.round(datum.y))} // Round is needed to avoid floats during animation
```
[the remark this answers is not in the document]
SteffiStuffel: its a naming coming from the library: https://formidable.com/open-source/victory/docs/victory-label#datum

src/routes/repetition/components/RepetitionProgressChart.tsx
```diff
@@ -0,0 +1,71 @@
+import { useFocusEffect } from '@react-navigation/native'
+import React, { ReactElement, useCallback } from 'react'
+import { heightPercentageToDP as hp } from 'react-native-responsive-screen'
+import { LineSegment, VictoryAxis, VictoryBar, VictoryChart, VictoryLabel } from 'victory-native'
+
+import theme from '../../../constants/theme'
```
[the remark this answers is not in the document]
SteffiStuffel: Yes i moved the color stuff outside the component, but I can move it inside and use `useTheme`

SteffiStuffel:
> 
> One thing i recognized at least on ios. When i enabled devMode the ReptitionTab is not appearing. I have to make a refresh in the terminal. Same for deactivating Since its devMode this is a minor issue but maybe we can create a issue for that
> 

Closing and reopening the app should help. There is no update/rerender mechanism in the bottom navigator, when something changes, but this will be added, as we need to update the badgeNumber (how may words need to repeated)

PR #998, opened 2023-10-16
905: Adjust write exercise for repetition exercise

ztefanie:
### Proposed changes

- [x] Split Write exercise into rendering and logic stuff
- [x] Created abstract class for generic write exercise data and functions
- [x] Moved existing standard and userVocabulary logic into own service
- [ ] ~~Rename StandardWriteExerciseService~~
- [x] Create RepetitionWriteExerciseService

### Side effects
Could break any existing logic of the writing exercise. Should be tested.

### Resolved issues

Fixes: #905

### DIFF
--- a/src/routes/write-exercise/services/__tests__/RepetitionWriteExerciseService.spec.ts
+++ b/src/routes/write-exercise/services/__tests__/RepetitionWriteExerciseService.spec.ts
@@ -1,0 +1,130 @@
+import { CommonActions, RouteProp } from '@react-navigation/native'
+import { waitFor } from '@testing-library/react-native'
+import React from 'react'
+
+import { SIMPLE_RESULTS } from '../../../../constants/data'
+import { VocabularyItem } from '../../../../constants/endpoints'
+import { RoutesParams, VocabularyItemResult } from '../../../../navigation/NavigationTypes'
+import { RepetitionService } from '../../../../services/RepetitionService'
+import VocabularyItemBuilder from '../../../../testing/VocabularyItemBuilder'
+import createNavigationMock from '../../../../testing/createNavigationPropMock'
+import RepetitionWriteExerciseService from '../RepetitionWriteExerciseService'
+
+jest.mock('../../../../services/AsyncStorage', () => ({
+  saveExerciseProgress: jest.fn(),
+}))
+
+describe('RepetitionWriteExerciseService', () => {
+  let vocabularyItems: VocabularyItem[]
+  let vocabularyItemsWithResults: VocabularyItemResult[]
+  let service: RepetitionWriteExerciseService
+  let route: RouteProp<RoutesParams, 'WriteExercise'>
+
+  const navigation = createNavigationMock<'WriteExercise'>()
+  const setCurrentIndex: React.Dispatch<React.SetStateAction<number>> = jest.fn()
+  const setIsAnswerSubmitted: React.Dispatch<React.SetStateAction<boolean>> = jest.fn()
+  const setVocabularyItemWithResults: React.Dispatch<React.SetStateAction<VocabularyItemResult[]>> = jest.fn()
+
+  const initTestData = () => {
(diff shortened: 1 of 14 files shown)


src/routes/write-exercise/services/__tests__/RepetitionWriteExerciseService.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+    )
+  }
+
+  RepetitionService.moveWordToPreviousSection = jest.fn()
+  RepetitionService.updateWordNodeCard = jest.fn()
+
+  beforeEach(initTestData)
+
+  describe('tryLater', () => {
+    it('should move word to then end and change nothing else in the results nor the nodeCards', () => {
```
LeandraH, line 60:
```suggestion
    it('should move word to the end and change nothing else in the results nor the nodeCards', () => {
```

src/routes/write-exercise/services/__tests__/StandardWriteExerciseService.spec.tsx
```diff
@@ -0,0 +1,131 @@
...
+      setCurrentIndex,
+      setIsAnswerSubmitted,
+      setVocabularyItemWithResults
+    )
+  }
+
+  beforeEach(initTestData)
+
+  describe('tryLater', () => {
+    it('should move word to then end and change nothing else in the results', () => {
```
LeandraH, line 56:
```suggestion
    it('should move word to the end and change nothing else in the results', () => {
```

LeandraH (commented):
Tested in emulated iOS, looks great, thank you! A few small issues:

I seem to have a small issue that doesn't happen with all words where
1. the word is ready to be trained again and in Lernstufe 1
2. I tap on "Jetzt wiederholen"
3. without doing the exercise, I tap to end the exercise and confirm that
4. and the word is downgraded to the not-learned bucket.
I would expect the word in that case to not move?

Also, the Solve All cheat in the repetition exercise seems to only solve the current VocabularyItem and then close the exercise. When I have three words to repeat and tap on solve all at the first word, the exercise closes and I still have two words to repeat.

I have also had at least one moment where I had one word to repeat, I did that (manually), and the RepetitionScreen showed that I had no words left to repeat but the footer showed one word to repeat.

ztefanie:
> Tested in emulated iOS, looks great, thank you! A few small issues:
> 
> I seem to have a small issue that doesn't happen with all words where
> 
> 1. the word is ready to be trained again and in Lernstufe 1
> 2. I tap on "Jetzt wiederholen"
> 3. without doing the exercise, I tap to end the exercise and confirm that
> 4. and the word is downgraded to the not-learned bucket.
>    I would expect the word in that case to not move?

I checked this and cannot reproduce. If you do :
1) Start the exercise
2) close the exercise
The progress stays the same, nothing is changed, every word is in the same bucket

If you do)
1) Start the exercise
2) Click "Show solution"
3) Close the exercise
The first word you did "wrong" (=you looked up the solution) is moved to a lower bucket. This is expected behaviour.

Can you please share some more information?

> 
> Also, the Solve All cheat in the repetition exercise seems to only solve the current VocabularyItem and then close the exercise. When I have three words to repeat and tap on solve all at the first word, the exercise closes and I still have two words to repeat.
> 
Good catch, fixed it.

> I have also had at least one moment where I had one word to repeat, I did that (manually), and the RepetitionScreen showed that I had no words left to repeat but the footer showed one word to repeat.

Yes updating of the footer does not yet work correctly, it is part of #906

LeandraH (approved):
Can't reproduce it anymore 🤷‍♀️

f1sh1918 (commented):
Just added this pr to fix building the app with xcode 15.
https://github.com/digitalfabrik/lunes-app/pull/1000

Without i can not test your issue

f1sh1918 (approved):
Nice :) couldn't find any issues for the writing exercises
Tested on ios
As i can see in #906 the vocabulary of the finished module won't be added to repetition in this issue

PR #1002, opened 2023-12-11
Unify apple account used for app signing

ztefanie:
### Short description
Changed the app store mail to be the same as for all other apps, as this makes the regular updates of the certificates easier and we are more likely not to look us out from our shared accounts, as they can be quite a pain, because of how the 2FA works.


LeandraH (approved):
Great idea!

sarahsporck (approved):

PR #1003, opened 2023-12-13
921: Add searching for professions

LeandraH:
### Short description

Adding a search function to the screen for selecting a profession.

### Proposed changes

- Added new reference to the endpoint to get all professions
- Added a search bar and a function to filter through all those professions (I decided against a fancier search function because we currently have 33 professions and so it feels like a premature optimization)
- Added the results, including a function to highlight the search string in the title

### Side effects

- I added a placeholder parameter to the SearchBar.

### Resolved issues

Fixes: #921 

---

### DIFF
--- a/release-notes/unreleased/921-search-for-profession.yml
+++ b/release-notes/unreleased/921-search-for-profession.yml
@@ -1,0 +1,6 @@
+issue_key: 921
+show_in_stores: true
+platforms:
+  - android
+  - ios
+de: Man kann nun nach Berufen suchen
--- a/src/constants/labels.json
+++ b/src/constants/labels.json
@@ -3,7 +3,8 @@
     "welcome": "Willkommen bei Lunes!",
     "selectProfession": "Wähle deinen Beruf!",
     "skipSelection": "Überspringen",
-    "confirmSelection": "Starten"
+    "confirmSelection": "Starten",
+    "searchProfession": "Suche Beruf"
   },
   "home": {
     "welcome": "Willkommen bei Lunes!",
--- a/src/hooks/useLoadDisciplines.ts
+++ b/src/hooks/useLoadDisciplines.ts
@@ -9,11 +9,15 @@ export type RequestParams =
     }
   | {
       parent: Discipline | null
+      allLevels?: boolean
     }
 
 const getEndpoint = (params: RequestParams): string => {
   if (isTypeLoadProtected(params)) {
     return ENDPOINTS.groupInfo
+  }
+  if (params.allLevels) {
+    return `${ENDPOINTS.discipline}/`
   }
   if (params.parent?.needsTrainingSetEndpoint) {
     return `${ENDPOINTS.trainingSet}/${params.parent.id}`
--- a/src/routes/scope-selection/ScopeSelectionScreen.tsx
+++ b/src/routes/scope-selection/ScopeSelectionScreen.tsx
@@ -1,12 +1,16 @@
 import { RouteProp, useFocusEffect } from '@react-navigation/native'
 import { StackNavigationProp } from '@react-navigation/stack'
-import React, { useLayoutEffect } from 'react'
+import normalize from 'normalize-strings'
+import React, { useLayoutEffect, useMemo, useState } from 'react'
+import { Pressable, Text } from 'react-native'
 import styled, { useTheme } from 'styled-components/native'
 
 import Button from '../../components/Button'
 import DisciplineListItem from '../../components/DisciplineListItem'
 import Header from '../../components/Header'
+import HorizontalLine from '../../components/HorizontalLine'
 import RouteWrapper from '../../components/RouteWrapper'
+import SearchBar from '../../components/SearchBar'
 import ServerResponseHandler from '../../components/ServerResponseHandler'
 import { ContentSecondary } from '../../components/text/Content'
 import { Heading } from '../../components/text/Heading'
@@ -15,19 +19,19 @@ import { Discipline } from '../../constants/endpoints'
 import { useLoadDisciplines } from '../../hooks/useLoadDisciplines'
 import useReadSelectedProfessions from '../../hooks/useReadSelectedProfessions'
 import { RoutesParams } from '../../navigation/NavigationTypes'
-import { setSelectedProfessions } from '../../services/AsyncStorage'
-import { getLabels } from '../../services/helpers'
+import { pushSelectedProfession, setSelectedProfessions } from '../../services/AsyncStorage'
+import { getLabels, splitTextBySearchString } from '../../services/helpers'
 
-const DisciplineContainer = styled.View`
-  margin: 0 ${props => props.theme.spacings.sm};
+const HighlightContainer = styled.View`
+  flex-direction: row;
 `
 
-const ButtonContainer = styled.View`
-  margin: ${props => props.theme.spacings.md} auto;
+const BoldText = styled.Text`
+  font-weight: bold;
 `
 
-const StyledText = styled(ContentSecondary)`
-  text-align: center;
+const StyledScrollView = styled.ScrollView`
+  background-color: ${props => props.theme.colors.background};
 `
 
 const TextContainer = styled.View`
@@ -35,9 +39,26 @@ const TextContainer = styled.View`
   margin-bottom: ${props => props.theme.spacings.lg};
 `
 
-const StyledScrollView = styled.ScrollView`
-  background-color: ${props => props.theme.colors.background};
+const StyledText = styled(ContentSecondary)`
+  text-align: center;
 `
+
+const SearchContainer = styled.View`
+  margin: ${props => props.theme.spacings.sm};
+`
+
+const ScopeContainer = styled.View`
+  margin: 0 ${props => props.theme.spacings.sm};
+`
+
+const ButtonContainer = styled.View`
+  margin: ${props => props.theme.spacings.md} auto;
+`
+
+export const searchProfessions = (disciplines: Discipline[] | null, searchKey: string): Discipline[] | undefined =>
+  disciplines?.filter(discipline =>
+    normalize(discipline.title).toLowerCase().includes(normalize(searchKey.toLowerCase()))
+  )
 
 type IntroScreenProps = {
   route: RouteProp<RoutesParams, 'ScopeSelection'>
@@ -48,6 +69,13 @@ const ScopeSelectionScreen = ({ navigation, route }: IntroScreenProps): JSX.Elem
   const { initialSelection } = route.params
   const { data: disciplines, error, loading, refresh } = useLoadDisciplines({ parent: null })
   const { data: selectedProfessions, refresh: refreshSelectedProfessions } = useReadSelectedProfessions()
+  const { data: allProfessions } = useLoadDisciplines({ parent: null, allLevels: true })
+  const [searchString, setSearchString] = useState<string>('')
+  const [showSearchResults, setShowSearchResults] = useState<boolean>(false)
+  const filteredProfessions = useMemo(
+    () => searchProfessions(allProfessions, searchString),
+    [allProfessions, searchString]
+  )
   const theme = useTheme()
 
   useFocusEffect(refreshSelectedProfessions)
@@ -73,6 +101,14 @@ const ScopeSelectionScreen = ({ navigation, route }: IntroScreenProps): JSX.Elem
     })
   }
 
+  const highlightText = (textArray: [string] | [string, string, string]): JSX.Element => (
+    <HighlightContainer>
+      <Text>{textArray[0]}</Text>
+      <BoldText>{textArray[1]}</BoldText>
+      <Text>{textArray[2]}</Text>
+    </HighlightContainer>
+  )
+
   const disciplineItems = disciplines?.map(item => (
     <DisciplineListItem key={item.id} item={item} onPress={() => navigateToDiscipline(item)} hasBadge={false} />
   ))
@@ -92,9 +128,37 @@ const ScopeSelectionScreen = ({ navigation, route }: IntroScreenProps): JSX.Elem
           )}
           <StyledText>{getLabels().scopeSelection.selectProfession}</StyledText>
         </TextContainer>
-        <ServerResponseHandler error={error} loading={loading} refresh={refresh}>
-          <DisciplineContainer>{disciplineItems}</DisciplineContainer>
-        </ServerResponseHandler>
+        <SearchContainer>
+          <SearchBar
+            query={searchString}
+            setQuery={input => {
+              setSearchString(input)
+              setShowSearchResults(input.length > 0)
+            }}
+            placeholder={getLabels().scopeSelection.searchProfession}
+          />
+        </SearchContainer>
+        {showSearchResults ? (
+          <ScopeContainer>
+            {filteredProfessions?.map(profession => (
+              <Pressable
+                key={profession.id}
+                onPress={async () => {
+                  await pushSelectedProfession(profession.id).then(() => {
+                    navigation.navigate('ManageSelection')
+                  })
+                }}>
+                {highlightText(splitTextBySearchString(profession.title, searchString))}
+                <Text>{profession.parentTitle}</Text>
+                <HorizontalLine />
+              </Pressable>
+            ))}
+          </ScopeContainer>
+        ) : (
+          <ServerResponseHandler error={error} loading={loading} refresh={refresh}>
+            <ScopeContainer>{disciplineItems}</ScopeContainer>
+          </ServerResponseHandler>
+        )}
         {initialSelection && (
           <ButtonContainer>
             <Button
--- a/src/services/helpers.ts
+++ b/src/services/helpers.ts
@@ -169,7 +169,7 @@ export const calculateScore = (vocabularyItemsWithResults: VocabularyItemResult[
     vocabularyItemsWithResults
       .filter(doc => doc.result === 'correct')
       .reduce((acc, vocabularyItemResult) => {
-        let score = acc
+        let score: number = acc
         switch (vocabularyItemResult.numberOfTries) {
           case 1:
             score += SCORE_FIRST_TRY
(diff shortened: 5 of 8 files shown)


release-notes/unreleased/921-search-for-profession.yml
```diff
@@ -0,0 +1,6 @@
+issue_key: 921
+show_in_stores: true
+platforms:
+  - android
+  - ios
+de: Man kann nun nach Berufen suchen
```
sarahsporck, line 6: I think we prefer "Nominalisierte" sentences
```suggestion
de: Eine Suche für Berufe wurde hinzugefügt.
```

src/constants/labels.json
```diff
@@ -3,7 +3,8 @@
     "welcome": "Willkommen bei Lunes!",
     "selectProfession": "Wähle deinen Beruf!",
     "skipSelection": "Überspringen",
-    "confirmSelection": "Starten"
+    "confirmSelection": "Starten",
+    "searchProfession": "Suche Beruf"
```
sarahsporck, line 7: In my opinion the term "Suche Beruf" sounds a little bit "holperig". How about prompting the user with "Beruf eingeben" or "nach Beruf suchen".

src/routes/scope-selection/ScopeSelectionScreen.tsx
```diff
@@ -92,9 +128,37 @@ const ScopeSelectionScreen = ({ navigation, route }: IntroScreenProps): JSX.Elem
...
+          />
+        </SearchContainer>
+        {showSearchResults ? (
+          <ScopeContainer>
+            {filteredProfessions?.map(profession => (
+              <Pressable
+                key={profession.id}
+                onPress={async () => {
+                  await pushSelectedProfession(profession.id).then(() => {
+                    navigation.navigate('ManageSelection')
```
sarahsporck, line 148: This screen is not navigatable, when you are in the initial profession selection (on first startup). I think you need to add the screen to `src/navigation/Navigator.tsx`

src/routes/scope-selection/ScopeSelectionScreen.tsx
```diff
@@ -92,9 +128,37 @@ const ScopeSelectionScreen = ({ navigation, route }: IntroScreenProps): JSX.Elem
...
+                  })
+                }}>
+                {highlightText(splitTextBySearchString(profession.title, searchString))}
+                <Text>{profession.parentTitle}</Text>
```
sarahsporck, line 152: Is the parentTitle always null? I never see it in the search.

src/routes/scope-selection/ScopeSelectionScreen.tsx
```diff
@@ -48,6 +69,13 @@ const ScopeSelectionScreen = ({ navigation, route }: IntroScreenProps): JSX.Elem
   const { initialSelection } = route.params
   const { data: disciplines, error, loading, refresh } = useLoadDisciplines({ parent: null })
   const { data: selectedProfessions, refresh: refreshSelectedProfessions } = useReadSelectedProfessions()
+  const { data: allProfessions } = useLoadDisciplines({ parent: null, allLevels: true })
+  const [searchString, setSearchString] = useState<string>('')
+  const [showSearchResults, setShowSearchResults] = useState<boolean>(false)
+  const filteredProfessions = useMemo(
+    () => searchProfessions(allProfessions, searchString),
+    [allProfessions, searchString]
+  )
```
sarahsporck, line 78: you should also filter professions the user hes already selected. As it should not be possible to add the same profession twice.

src/routes/scope-selection/ScopeSelectionScreen.tsx
```diff
@@ -73,6 +101,14 @@ const ScopeSelectionScreen = ({ navigation, route }: IntroScreenProps): JSX.Elem
     })
   }
 
+  const highlightText = (textArray: [string] | [string, string, string]): JSX.Element => (
+    <HighlightContainer>
+      <Text>{textArray[0]}</Text>
+      <BoldText>{textArray[1]}</BoldText>
+      <Text>{textArray[2]}</Text>
+    </HighlightContainer>
+  )
```
sarahsporck, line 110: Just as a note. This only highlights the first occurence of the match. In this case this should be fine though :).

src/routes/scope-selection/ScopeSelectionScreen.tsx
```diff
@@ -92,9 +128,37 @@ const ScopeSelectionScreen = ({ navigation, route }: IntroScreenProps): JSX.Elem
...
+            placeholder={getLabels().scopeSelection.searchProfession}
+          />
+        </SearchContainer>
+        {showSearchResults ? (
+          <ScopeContainer>
+            {filteredProfessions?.map(profession => (
+              <Pressable
+                key={profession.id}
+                onPress={async () => {
+                  await pushSelectedProfession(profession.id).then(() => {
```
sarahsporck, line 147: Actually, I think the buttons should not be pressable, once you've selected a profession. Maybe we can show a loading screen or just disable the buttons.

sarahsporck (changes requested):
Tested on Android emulator. 
Nice, finally a new Lunes feature. I left you a few comments :)

src/routes/scope-selection/ScopeSelectionScreen.tsx
LeandraH: Yes, I decided that's fine here :)

src/routes/scope-selection/ScopeSelectionScreen.tsx
LeandraH: Yes, it is. I decided to put it in, in case someone fills that field in the future.

src/routes/scope-selection/ScopeSelectionScreen.tsx
LeandraH: Hmm, but if someone wants to select a second profession, I don't see a problem with that. You already can't select the same profession twice anymore since it doesn't show up in the search anymore.

sarahsporck: But when I search for the profession it is still visible. At least if I add the module via the settings
Bildschirmfoto 2024-01-17 um 15 21 34

sarahsporck (changes requested):

LeandraH:
You're right, good catch! I now changed it around a little so that an already selected profession does show up in the search result but is disabled.

simulator_screenshot_468D32FF-6821-40AF-9879-BAE36FE0FFEA

ztefanie (changes requested):
Currently it is possible to select root disiciplines, e.g. "Bau, Farbe und Holz". They should be filtered and not be in the list, as they are not a valid choice. You can filter them, by their parentTitle, it should not be null.

src/hooks/useLoadDisciplines.ts
```diff
@@ -28,7 +32,12 @@ export const loadDisciplines = async (loadingInfo: RequestParams): Promise<Disci
   const url = getEndpoint(loadingInfo)
   const apiKey = isTypeLoadProtected(loadingInfo) ? loadingInfo.apiKey : loadingInfo.parent?.apiKey
   const response = await getFromEndpoint<ServerResponseDiscipline[]>(url, apiKey)
-  return response.map(item => formatDiscipline(item, loadingInfo))
+  if (!isTypeLoadProtected(loadingInfo) && loadingInfo.onlyChildren) {
```
ztefanie, line 35: I strongly prefer to move this logic to the component. So just load all disciplines here, no need to pass "onlyChildren" or anything.

I recommend:
Create a new hook useLoadAllDisciplines
Call this hook from the component and filter the root disciplines in the component.

src/navigation/Navigator.tsx
```diff
@@ -43,6 +44,7 @@ const HomeStackNavigator = (): JSX.Element | null => {
           initialParams={{ initialSelection: true }}
           options={({ navigation }) => options(manageSelection, navigation)}
         />
+        <Stack.Screen name='ManageSelection' component={ManageSelectionsScreen} />
```
ztefanie, line 47: Why have you added this here? Isn't it enough to keep it in the HomeScreenNavigator as before?

src/routes/scope-selection/ScopeSelectionScreen.tsx
```diff
@@ -15,30 +19,60 @@ import { Discipline } from '../../constants/endpoints'
...
-const DisciplineContainer = styled.View`
-  margin: 0 ${props => props.theme.spacings.sm};
+const HighlightContainer = styled.Text<{ disabled: boolean }>`
+  flex-direction: row;
+  color: ${props => (props.disabled ? props.theme.colors.disabled : props.theme.colors.black)};
 `
 
-const ButtonContainer = styled.View`
-  margin: ${props => props.theme.spacings.md} auto;
+const BoldText = styled.Text`
```
ztefanie, line 30: Please use the `ContentTextBold` from /components/text

src/routes/scope-selection/ScopeSelectionScreen.tsx
```diff
@@ -15,30 +19,60 @@ import { Discipline } from '../../constants/endpoints'
...
+  font-weight: bold;
 `
 
-const StyledText = styled(ContentSecondary)`
-  text-align: center;
+const StyledScrollView = styled.ScrollView`
+  background-color: ${props => props.theme.colors.background};
```
ztefanie, line 35: This does not have any effect, does it? background has background color anyway?

src/routes/scope-selection/ScopeSelectionScreen.tsx
```diff
@@ -92,9 +133,41 @@ const ScopeSelectionScreen = ({ navigation, route }: IntroScreenProps): JSX.Elem
...
+                <Pressable
+                  key={profession.id}
+                  onPress={async () => {
+                    await pushSelectedProfession(profession.id).then(() => {
+                      navigation.navigate('ManageSelection')
+                    })
+                  }}
+                  disabled={disabled}>
+                  {highlightText(splitTextBySearchString(profession.title, searchString), disabled)}
+                  <ButtonSecondLine disabled={disabled}>{profession.parentTitle}</ButtonSecondLine>
```
ztefanie, line 160: This parentTitle is not displayed. Please remove or fix it

ztefanie (changes requested):
Sorry for taking so long to review.
I tested it mostly works as expected. :tada:

src/navigation/Navigator.tsx
LeandraH: No, I then can't actually navigate in there when opening the app after a fresh installation.
image

src/routes/scope-selection/ScopeSelectionScreen.tsx
LeandraH: True 😅

ztefanie (approved):
Almost there, only one small comment, please adjust the header of the screen:
921-header

@steffenkleinle @f1sh1918 Can you also please review here, as this PR is already open for a long time.

src/routes/scope-selection/ScopeSelectionScreen.tsx
```diff
@@ -1,44 +1,68 @@
...
+    normalize(discipline.title).toLowerCase().includes(normalize(searchKey.toLowerCase()))
+  )
+
+const highlightText = (textArray: [string] | [string, string, string], disabled: boolean): JSX.Element => (
+  <HighlightContainer disabled={disabled}>
+    <ContentTextLight>{textArray[0]}</ContentTextLight>
+    <ContentTextBold>{textArray[1]}</ContentTextBold>
+    <ContentTextLight>{textArray[2]}</ContentTextLight>
+  </HighlightContainer>
+)
```
steffenkleinle, line 64: Can we extract this along with the call of `splitTextBySearchString(profession.title, searchString)` to a separate `TextHighlighter` component or similar?

src/routes/scope-selection/ScopeSelectionScreen.tsx
```diff
@@ -1,44 +1,68 @@
...
 `
 
+const ButtonContainer = styled.View`
+  margin: ${props => props.theme.spacings.md} auto;
+`
+
+export const searchProfessions = (disciplines: Discipline[] | undefined, searchKey: string): Discipline[] | undefined =>
+  disciplines?.filter(discipline =>
+    normalize(discipline.title).toLowerCase().includes(normalize(searchKey.toLowerCase()))
+  )
```
steffenkleinle, line 56: I think this should go to a helper. Additionally, we should probably use our custom `normalizeSearchString` helper instead of plain `normalize`.

src/routes/scope-selection/ScopeSelectionScreen.tsx
```diff
@@ -92,9 +125,40 @@ const ScopeSelectionScreen = ({ navigation, route }: IntroScreenProps): JSX.Elem
...
+                  <HorizontalLine />
+                </Pressable>
+              )
+            })}
+          </ScopeContainer>
+        ) : (
+          <ServerResponseHandler error={error} loading={loading} refresh={refresh}>
+            <ScopeContainer>{disciplineItems}</ScopeContainer>
+          </ServerResponseHandler>
+        )}
```
steffenkleinle, line 161: I think we should extract this to a separate component to make it reusable/this file easier to read.

src/navigation/Navigator.tsx
steffenkleinle: @LeandraH do you have any clue why this wasn't needed before?

LeandraH: I assume it's needed now because with this PR you can navigate to the `ManageSelectionsScreen` from the `ScopeSelectionScreen`, and that one is in the `Navigator` and not in the `HomeStackNavigator`
https://github.com/digitalfabrik/lunes-app/pull/1003/files#diff-056a90d9f549141742c1c899d05ac3a087ef8b87b4f1fa7cdccee90d48bd9fa6R147

src/routes/scope-selection/ScopeSelectionScreen.tsx
LeandraH: I'm not sure that's necessary since the only place (at least that I can think of) that might re-use this is the word search, and that doesn't have any highlighting. So I'd consider an extraction to a separate component a premature optimization. What do you think?

steffenkleinle: Hmm for me its mostly about readability/modularity. But its a well meant suggestion so I am also okay with not doing that here. Just thought it reduces the complexity of this file and adheres to classic design principles.

src/routes/scope-selection/ScopeSelectionScreen.tsx
LeandraH: Great idea!

src/services/helpers.ts
```diff
@@ -241,3 +241,8 @@ export const splitTextBySearchString = (allText: string, highlight: string): [st
     allText.slice(indexAfterHighlight),
   ]
 }
+
+export const searchProfessions = (disciplines: Discipline[] | undefined, searchKey: string): Discipline[] | undefined =>
+  disciplines?.filter(discipline =>
+    normalizeStrings(discipline.title).toLowerCase().includes(normalizeStrings(searchKey.toLowerCase())),
```
steffenkleinle, line 247: :upside_down_face: This does not do `trim()` the search string. We could introduce a custom helper as in Integreat to handle everything and reuse it:
``` ts
export const normalizeString = (str: string): string => normalizeStrings(str).toLowerCase().trim()
```

src/routes/scope-selection/ScopeSelection.tsx
```diff
@@ -0,0 +1,119 @@
...
+  navigateToDiscipline: (item: Discipline) => void
+  navigateToManageSelection: () => void
+  selectedProfessions: number[] | null
+}
+
+const ScopeSelection = ({
+  queryTerm,
+  setQueryTerm,
+  showSearchResults,
+  setShowSearchResults,
```
steffenkleinle, line 61: :wrench: AFAICS `showSearchResults` is always depending on the length of the `queryTerm`, i.e. if the query term is not empty it is true and otherwise false. Therefore this is redundant and could be removed.

src/routes/scope-selection/ScopeSelection.tsx
```diff
@@ -0,0 +1,119 @@
...
+  navigateToDiscipline,
+  navigateToManageSelection,
+  selectedProfessions,
+}: ScopeSelectionProps): JSX.Element => {
+  const { data: disciplines, error, loading, refresh } = useLoadDisciplines({ parent: null })
+  const allProfessions = useLoadAllDisciplines()
+    .data?.filter(discipline => discipline.total_discipline_children === 0)
+    .map(discipline => formatDiscipline(discipline, {}))
```
steffenkleinle, line 69: :question: Isn't the result of these exactly the same if `parent: null` is passed (except the `filter`)? AFAICS `useLoadAllDisciplines` queries the same url for `parent: null`. Therefore this could be unified. Not 100% sure though as the logic in `useLoadDisciplines` isn't super straight forward.

src/routes/scope-selection/ScopeSelection.tsx
```diff
@@ -0,0 +1,119 @@
...
+        <ScopeContainer>
+          {filteredProfessions?.map(profession => {
+            const disabled = !!selectedProfessions?.includes(profession.id)
+            return (
+              <StyledPressable
+                key={profession.id}
+                onPress={async () => {
+                  await pushSelectedProfession(profession.id).then(() => {
+                    navigateToManageSelection()
+                  })
```
steffenkleinle, line 102: :wrench: Please avoid mixing `async/await` and `then`. 
```suggestion
                onPress={async () => {
                  await pushSelectedProfession(profession.id)
                  await navigateToManageSelection()
                })
```

src/routes/scope-selection/ScopeSelection.tsx
LeandraH: Good catch, thanks!

src/routes/scope-selection/ScopeSelection.tsx
LeandraH: It's not exactly the same. In the case of  `useLoadDisciplinces` with`parent: null`, we call the endpoint `/disciplines_by_level` but in the case of `useLoadAllDisciplines`, we call the endpoint `/disciplines`.
I could also put this into `useLoadDisciplines` as another special case but it really is already a pretty hard to understand file. What do you think?

steffenkleinle: Ahh sorry, I guess I missed that. Thanks for the explanation, nvm :)

steffenkleinle (approved):
Tested on android, good job!

Issue #1023, opened 2024-05-15
App crashing when opening vocabulary list


LeandraH:
### Describe the Bug

When opening the vocabulary list, the app crashes.

### Steps to Reproduce

1. Open the app.
2. Click on 'Suche'
3. See crash

### Additional Information

The field article was renamed to singular_article. This is probably where the problems are coming from.
https://github.com/digitalfabrik/lunes-cms/pull/518

LeandraH:
Fixed in the CMS

PR #1024, opened 2024-05-15
1023: Rename article field in API response

LeandraH:
### Short description

The app started crashing when opening the vocabulary list because the field `article` had been renamed to `singular_article` in the CMS. This renames that field also in the response parser.

### Proposed changes

- Renamed `article` to `singular_article` in `useLoadVocabularyItems`

### Side effects

- The app should not crash anymore when opening the vocabulary list.

### Resolved issues

Fixes: #1023 

---

### DIFF
--- a/release-notes/unreleased/1023-rename-article-field.yml
+++ b/release-notes/unreleased/1023-rename-article-field.yml
@@ -1,0 +1,6 @@
+issue_key: 1023
+show_in_stores: true
+platforms:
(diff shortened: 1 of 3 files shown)


svenseeberg (approved):

f1sh1918:
thx for the quick fix. Since this should be fixed in the backend, i think the pr can be closed

PR #1034, opened 2024-10-31
873: Use react-native-vision-camera and react-native-image-picker

steffenkleinle:
### Short description

Replaced a few libraries and fixed their usages.

### Proposed changes

- Replace the deprecated react-native-camera with react-native-vision-camera
- Replace react-native-image-crop-picker with react-native-image-picker (due to smaller size and since the old library was not working for me after the react-native upgrade)
- Fix react-native-audio-recorder-player by correctly handling short press edge case
- Some cleanup

### Side effects

- Hopefully none?

### Resolved issues

Fixes: #873 

---


LeandraH (approved):
Tested on real iOS

ztefanie (approved):
Code looks good. Not tested as mentioned

PR #1044, opened 2024-12-04
Update copyright

LeandraH:
### Short description

Just setting the copyright to update dynamically because it annoyed me that it's from 2023 when we're almost in 2025.


steffenkleinle (approved):

ztefanie (approved):
Thanks for updating. Not tested

Issue #1045, opened 2024-12-09
Fix bump version task 


ztefanie:
Currently the delivery workflow is failing. Remove the "create jira release" step and check if the delivery works again.

PR #1047, opened 2025-01-06
1045: Remove JIRA release

LeandraH:
### Short description

Since we disabled our JIRA space but hadn't removed JIRA releases from our workflow yet, the deployments started failing. This removes the JIRA releases.

### Proposed changes

- Removed the script for the JIRA releases plus the package `jira-client` that we used for it.
- Scrubbed mentions of JIRA
- Added an example of a cURL request to the documentation for triggering a release from the terminal.

### Side effects

- The pipeline should be marginally shorter.

### Testing

I ran the [pipeline on the branch](https://app.circleci.com/pipelines/github/digitalfabrik/lunes-app/3595/workflows/a0a628b4-63a1-400f-841c-184a390168ef/jobs/11845) and was informed that version bumps are only allowed on main (makes sense). But I'm not really willing to deploy unreviewed code on main in order to test something, so this is untested. Does anybody have an idea for how to test?

### Resolved issues

Fixes: #1045

---


steffenkleinle (approved):
Looks good, thanks!

Issue #1051, opened 2025-01-20
Fix loss of precision error in vocabulary list


lunars97:
### Describe the Bug

After react-native update going the first word by clicking on the button **"Nächstes Wort"** in vocabulary list is showing `Error: Exception in HostFunction: Loss of precision during arithmetic conversion: (long long) 9.090909090909092`
### Steps to Reproduce

1. Go to 'Vocabulary List'
2. Click on 'on the first word' and on the button **"Nächstes Wort"**
3. See error

### Expected Behavior

Next word should appear

### Actual Behavior

Error: Exception in HostFunction

### Additional Information

The error appeared here #1010 

<details>
  <summary>Traceback</summary>

```

```

</details>

PR #1052, opened 2025-01-20
1051: Fix loss of precision error in vocabulary list

lunars97:
### Short description

After upgrading react native to v0.74, the react-native-paper was also updated which resulted in the `loss precision error`. The `progress` prop of` ProgressBar `component caused this issue. The workaround is mentioned [here](https://github.com/facebook/react-native/issues/47635#issuecomment-2507431266)

### Proposed changes

- Change the `progress` prop to `animatedValue`

### Side effects

- should be none :)

### Resolved issues

Fixes: #1051 

---

### DIFF
--- a/src/components/ExerciseHeader.tsx
+++ b/src/components/ExerciseHeader.tsx
@@ -133,7 +133,7 @@ const ExerciseHeader = ({
     <>
       {showProgress && (
         <ProgressBar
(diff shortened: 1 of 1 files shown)


david-venhoff (approved):
Thanks, LGTM!

LeandraH (approved):
Tested on emulated iOS where I couldn't reproduce the original issue but this also didn't break anything 😅

PR #1054, opened 2025-01-24
1042: Add text to long term excercise

david-venhoff:
### Short description

This pr addresses #1042 , which solves two problems:

### Proposed changes

- Add a text to the long term repetition info modal
- Fix the pluralization of the `wordsToRepeat` label
- (Not part of the original issue) Disable the repetition button if there are no words to repeat

### Side effects

should be none

### Resolved issues

Fixes: #1042 

---


LeandraH (approved):
Very nice, tested on emulated iOS

Issue #995, opened 2023-09-08
Badge number in bottom bar needs to be updated


ztefanie:
Currently we are showing the number of words, that need repetition as a badge on the BottomTabNavigator. But this number is not updated, unless the app is closed and reopened or a screen without the bottom navigation bar is opnend and closed. The badge number should be updated correctly, when the value changes.

Number should be hidden, when it is zero

david-venhoff:
Partially a duplicate of #1040

PR #1055, opened 2025-01-28
Hide the repetition badge number if it is zero

david-venhoff:
### Short description

This pr hides the repetition badge number if it is zero. This fixes the non-duplicate part of issue #995.

### Proposed changes

- Hide number if zero

### Side effects

none

### Resolved issues

Fixes: #995

---


LeandraH (approved):
Nice! Tested on emulated iOS

Issue #1067, opened 2025-03-12
Wrong title for "Vokabelliste" and "Vokabeln üben"


david-venhoff:
### Describe the Bug

In #1062 the tab "Sammlung" now contains both "Meine Vokabeln" and "Favoriten", so it makes not much sense that clicking on "Vokabelliste" in "Meine Vokabeln" shows  "Sammlung" as header. The back button also says "Sammlung".

### Expected Behavior

Instead maybe show something like "Vokabelliste" for the header and "zurück" for the back button

PR #1062, opened 2025-03-04
Merge tabs "Meine Wörter" and "Favoriten"

david-venhoff:
### Short description

Merges the two tabs into a new tab "Sammlung", which has two top bar tabs for "Meine Wörter" and "Favoriten".

### Proposed changes

- Add the material top tabs dependency
- Create a patch for react-native-tab-view, because the 6.x.x versions are not properly supported anymore. More context: https://github.com/react-navigation/react-navigation/issues/11989#issuecomment-2584247479
- Remove the old favorites tab

### Side effects

- Also resolves #939, because there are now fewer tabs

### Resolved issues

Fixes: #997
Fixes: #982 
Fixes: #939 

---

PR #1068, opened 2025-03-12
Fix wrong labels in user vocabulary

david-venhoff:
### Short Description

In #1062 I forgot to update the labels in the user vocabulary screens, but they did not make too much sense anymore.

### Proposed Changes

- Change the labels from "Sammlung" to be more specific and user more generic back labels

### Side Effects

- Should be none

### Resolved Issues

Fixes: #1067

---


LeandraH (approved):
Nice, good catch!

I would probably prefer putting the back button higher in the tab, it looks a little lost at the moment. What do you think?

image

david-venhoff:
Good idea, for some reason the padding was correct on android, but not on ios 🤔 

But adding `headerStatusBarHeight: 0` seems to make it work on both ios and android.

Issue #1040, opened 2024-11-26
Update long term exercise screens correctly


ztefanie:
### Describe the Bug
* Fix the navigation bar badge: After finishing exercises or adding new words, the badge is not always updated correctly.
* Fix the navigation to the repetition screen, should always be updated, not working correctly.

### Steps to Reproduce
* Add new words to the repetition exercise. See both the footer and the screen are not updated as expected. When closing and reopening the app, the view is displayed correctly.

### Expected Behavior
Screen and badge are updated immediately

### Actual Behavior
Screens do not update

david-venhoff:
Looks like the main problem is that our async store is not reactive. The current workaround is that we just refresh values from the store every time a focus event happens, which seems to go wrong here somehow. 

Maybe a more robust solution would be to switch to something like https://github.com/ammarahm-ed/react-native-mmkv-storage?

david-venhoff:
Actually, I can only reproduce this issue when I add words with the devtools to fill the repetition excercise with data. So it is rather low prio imo

david-venhoff:
For future reference, I have create a proof of concept branch that switches to reactive sync storage, which avoids the need for any focus effect hacks: https://github.com/digitalfabrik/lunes-app/tree/experiment/sync-storage

PR #1074, opened 2025-04-01
Add Storage abstraction over asyncStorage

david-venhoff:
### Short Description

Currently, we are directly using `AsyncStorage` to read and write from storage. However, this causes some issues:
- No reactivity: `AsyncStorage` is not reactive and requires manually reloading data when they have changed. This is not always possible though.
- Every read is async, which requires components to handle an additional 'Data have not loaded yet' state, which they should not have to concern themselves with

To improve on this situation, this pr adds an abstraction over async storage: A new class `StorageCache`, which provides sync read access and automatically notifies react when a value has changed. Because `StorageCache` keeps a cache of every storage value, every storage write has to go through this class to avoid cache invalidation. Also it has to be loaded once at app startup, which could make it theoretically slightly slower.

### Proposed Changes

- Add storage context provider
- Remove legacy storage apis
- move the remaining legacy storage apis to `storageUtils.ts`
- adjust tests by introducing a new `renderWithStorage` function, which allows to set storage values without having to use a mock
- Migrate favorites storage once at app startup, instead of on every access

### Side Effects

- Ideally none, except for no more bugs like #1040

### Testing

Well, pretty much the entire app D:

### Resolved Issues

Fixes: #1040

---


src/services/Storage.tsx
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  private notifyListeners = (key: keyof Storage) => {
+    this.listeners.get(key)?.forEach(listener => {
+      listener()
+    })
+  }
+}
+
+export const StorageCacheContext = createContext<StorageCache>(StorageCache.createForTesting())
```
steffenkleinle, line 142: 🙃 I think the `createForTesting` naming here is a little confusing as we use it for production as well. 
```suggestion
export const StorageCacheContext = createContext<StorageCache>(newDefaultStorage())
```

src/services/Storage.tsx
```diff
@@ -0,0 +1,180 @@
...
+export const getStorageItem = async <T extends keyof Storage>(key: T): Promise<Storage[T]> =>
+  getStorageItemOr(getStorageKey(key), defaultStorage[key])
+
+const setStorageItem = async <T extends keyof Storage>(key: T, value: Storage[T]): Promise<void> => {
+  await AsyncStorage.setItem(getStorageKey(key), JSON.stringify(value))
+}
+
+// https://github.com/react-native-async-storage/async-storage/issues/401#issuecomment-2508924008
+export class StorageCache {
+  private readonly listeners: Map<string, Set<() => void>> = new Map()
```
steffenkleinle, line 84: ❓ Why did you decide to go for a listener-based approach instead of keeping the storage in a `useState`? Just out of curiosity.

steffenkleinle (commented):
Great refactoring, simplifies using the async storage a lot! Not tested.

src/routes/repetition/RepetitionScreen.tsx
```diff
@@ -75,16 +74,15 @@ type RepetitionScreenProps = {
 const RepetitionScreen = ({ navigation }: RepetitionScreenProps): ReactElement => {
   const [isModalVisible, setIsModalVisible] = useState<boolean>(false)
   const { repeatWords, repeatNow, wordsToRepeat, yourLearningProgress, infoModalContentText } = getLabels().repetition
-  const { data: numberOfWordsNeedingRepetition, refresh: refreshNumberOfWordsNeedingRepetition } = useLoadAsync(
-    RepetitionService.getNumberOfWordsNeedingRepetitionWithUpperBound,
-    undefined,
+  const repetitionService = useRepetitionService()
+  const numberOfWordsNeedingRepetition = useMemo(
```
LeandraH, line 78: 🤔 I'm not sure that `useMemo` is necessary

src/routes/write-exercise/services/AbstractWriteExerciseService.ts
```diff
@@ -14,7 +15,8 @@ export default abstract class AbstractWriteExerciseService {
     public navigation: StackNavigationProp<RoutesParams, 'WriteExercise'>,
     public setCurrentIndex: React.Dispatch<React.SetStateAction<number>>,
     public setIsAnswerSubmitted: React.Dispatch<React.SetStateAction<boolean>>,
-    public setVocabularyItemWithResults: React.Dispatch<React.SetStateAction<VocabularyItemResult[]>>, // eslint-disable-next-line no-empty-function
+    public setVocabularyItemWithResults: React.Dispatch<React.SetStateAction<VocabularyItemResult[]>>,
+    public storageCache: StorageCache, // eslint-disable-next-line no-empty-function
```
LeandraH, line 19:
```suggestion
    public storageCache: StorageCache,
    // eslint-disable-next-line no-empty-function
```
Just to make it clear at first glance that the eslint problem is in the next line, not the storage one (I know you didn't write it, but it did confuse me 😅 )

src/routes/write-exercise/services/RepetitionWriteExerciseService.ts
```diff
@@ -1,10 +1,32 @@
...
 
 class RepetitionWriteExerciseService extends AbstractWriteExerciseService {
+  public repetitionService: RepetitionService
+
+  constructor(
+    route: RouteProp<RoutesParams, 'WriteExercise'>,
+    navigation: StackNavigationProp<RoutesParams, 'WriteExercise'>,
+    setCurrentIndex: React.Dispatch<React.SetStateAction<number>>,
+    setIsAnswerSubmitted: React.Dispatch<React.SetStateAction<boolean>>,
+    setVocabularyItemWithResults: React.Dispatch<React.SetStateAction<VocabularyItemResult[]>>, // eslint-disable-next-line no-empty-function
```
LeandraH, line 19:
```suggestion
    setVocabularyItemWithResults: React.Dispatch<React.SetStateAction<VocabularyItemResult[]>>,
    // eslint-disable-next-line no-empty-function
```

src/services/Storage.tsx
```diff
@@ -0,0 +1,180 @@
...
+   * @param key The key of the storage item
+   */
+  getItem = <T extends keyof Storage>(key: T): Readonly<Storage[T]> => this.storage[key]
+
+  /**
+   * Returns a storage item that may be modified
+   *
+   * @param key The key of the storage item
+   */
+  getMutableItem = <T extends keyof Storage>(key: T): Storage[T] => JSON.parse(JSON.stringify(this.getItem(key)))
```
LeandraH, line 116: 🤔 I'm not sure the `getMutableItem` is necessary on top of `getItem` with `setItem`. But that might be a bit of React instinct in me, to only have one-way data binding.

LeandraH (approved):
Very cool, nicely done! That looks a lot cleaner to me! Tested on real iOS

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
```diff
@@ -86,8 +88,8 @@ const ChoiceExerciseScreen = ({
   }, [results, currentWord])
 
   const onExerciseFinished = async (results: VocabularyItemResult[]): Promise<void> => {
-    const progress = await getExerciseProgress()
-    await saveExerciseProgress(disciplineId, exerciseKey, results)
+    const progress = storageCache.getItem('progress')
```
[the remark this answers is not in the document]
david-venhoff: I think no, because this is a callback (To my understanding, running hooks in a callback is invalid)

src/services/Storage.tsx
david-venhoff: The advantage of the listener based approach is that it is more granular. For example, if the storage was kept as a simple state, then setting `storage.foo = new_value()` would not only rerender the current component, but also all other components that read unrelated properties of storage. 
The listener-based approach allows to keep the state updates separate: If i update `storage.foo`, only components that read `storage.foo` will update.

src/services/Storage.tsx
```diff
@@ -0,0 +1,180 @@
...
+  nextUserVocabularyId: number
+  favorites: Favorite[]
+}
+
+/**
+ * The default storage contains the default values that will be used if nothing else
+ * is stored yet on the device.
+ * It is also useful for testing, to mock the actual storage implementation.
+ */
+export const newDefaultStorage = (): Storage => ({
```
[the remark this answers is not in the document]
david-venhoff: The problem is that `setItem` modifies the `Storage` object (using `this.storage[key] = value`). With the `newDefaultStorage` function, we can simply reset the storage cache after running a test (using `newDefaultStorage()`, so that it cannot interfere with other tests.

src/routes/home/components/ProfessionDetails.tsx
```diff
@@ -31,16 +31,15 @@ const ProfessionDetails = ({
   navigateToDiscipline,
   navigateToNextExercise,
 }: ProfessionDetailsProps): ReactElement | null => {
-  const { data: progress, refresh: refreshProgress } = useReadProgress(discipline)
+  const progress = useReadProgress(discipline)
   const { data: nextExerciseData, refresh: refreshNextExercise } = useLoadNextExercise(discipline)
 
-  const disciplineAlreadyStarted = progress !== null && progress !== 0
+  const disciplineAlreadyStarted = progress !== 0
```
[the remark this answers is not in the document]
david-venhoff: Done. Do you know why it was allowed to be undefined in the first place? As far as I know, undefined values are ignore anyways in `JSON.serialize`, so they would not exist after reopening the app.

src/services/storageUtils.ts
```diff
@@ -0,0 +1,164 @@
...
+  newUserVocabularyItem: VocabularyItem,
+): Promise<boolean> => {
+  const userVocabulary = getUserVocabularyItems(storageCache.getItem('userVocabulary'))
+  const index = userVocabulary.findIndex(item => JSON.stringify(item) === JSON.stringify(oldUserVocabularyItem))
+  if (index === -1) {
+    return false
+  }
+  userVocabulary[index] = newUserVocabularyItem
+  await storageCache.setItem('userVocabulary', userVocabulary)
+  return true
```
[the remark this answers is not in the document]
david-venhoff: I don't know, probably the return value was used in the past but not anymore? 
Guess I will just remove it :)

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
```diff
@@ -50,6 +51,7 @@ const ChoiceExerciseScreen = ({
   route,
   exerciseKey,
 }: SingleChoiceExerciseProps): ReactElement => {
+  const storageCache = useContext(StorageCacheContext)
```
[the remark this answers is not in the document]
david-venhoff: Do you imagine something like this for the implementation?

https://www.typescripttutorial.net/typescript-tutorial/typescript-function-overloadings/

src/services/Storage.tsx
```diff
@@ -0,0 +1,180 @@
...
+    case 'cmsUrlOverwrite':
+      return 'cms'
+    case 'customDisciplines':
+      return 'customDisciplines'
+    case 'userVocabulary':
+      return 'userVocabulary'
+    case 'nextUserVocabularyId':
+      return 'userVocabularyNextId'
+    case 'favorites':
+      return 'favorites-2'
```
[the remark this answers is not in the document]
david-venhoff: Yeah, pretty ugly 😅 
Actually, I think I can just replace that with an object. That is probably more idiomatic: https://github.com/digitalfabrik/lunes-app/pull/1074/commits/08dc4328c31e8588c206f04efdc3352750ef0588

src/services/Storage.tsx
```diff
@@ -0,0 +1,180 @@
...
+  cmsUrlOverwrite: null,
+  customDisciplines: [],
+  userVocabulary: [],
+  nextUserVocabularyId: 1,
+  favorites: [],
+})
+const defaultStorage = newDefaultStorage()
+
+// eslint-disable-next-line consistent-return
+export const getStorageKey = (key: keyof Storage): string => {
```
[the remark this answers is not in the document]
LeandraH: Yup, you're right, good point

src/services/Storage.tsx
LeandraH: Aaaah, I see, it's really more of a copy of the item that we get there, which is why that result is mutable but the actual item is not. Got it, thank you! Yeah, I don't see a more elegant solution either :(

LeandraH (approved):
Very cool!

src/services/Storage.tsx
steffenkleinle: I think `createDummy` kinda transports the same message, thats why I put `newDefaultStorage()` as my suggestion. But this is just a minor remark, so just pick what ever you feel like fits the most.

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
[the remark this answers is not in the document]
steffenkleinle: True, missed that. But why do you need to call that in the callback itself? Shouldn't it be enough/updated just fine if you use it outside of the callback?

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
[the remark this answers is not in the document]
david-venhoff: Alright, I just created a different hook `useStorageCache`

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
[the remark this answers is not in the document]
david-venhoff: Yes, but is there any advantage to doing that? To me, both solutions feel about the same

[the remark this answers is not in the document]
steffenkleinle: Hmm, I just like things being consistent. When looking at this I wonder why this is implemented this way and whether there is a difference, especially if you are new to the project. However, obviously very minor and I was just curious why this was implemented this way.

steffenkleinle (approved):
Not tested

src/services/Storage.tsx
david-venhoff: Hmm, I think I did not fully understand your suggestion. Do you suggest that we should change the return type of `newDefaultStorage` to `StorageCache`? That would require making the constructor of `StorageCache` public, but in that case we could also just write `new StorageCache(newDefaultStorage())`. But I would like to keep the constructor private, to emphasize that there are only two ways of creating storage cache: One method that creates a dummy storage cache and one method that uses AsyncStorage.

My reasoning for`createDummy` is that it is alright because the "dummy" storage cache is only used as a default value and should never be actually accessed in production

steffenkleinle: I missed that its private, sorry. Sounds good as described 👍

david-venhoff: Since this is not too important I will go ahead and merge the pr, just wanted to give my reasoning

david-venhoff:
Thanks a lot for your reviews!

PR #1076, opened 2025-04-10
Fix for Xcode 16.3

LeandraH:
### Short Description

The latest Xcode version (16.3) doesn't work with older versions of React-Native, this PR bumps us up to a supported one (https://github.com/facebook/react-native/releases/tag/v0.76.9, more info [here](https://github.com/facebook/react-native/pull/50431) and [here](https://github.com/facebook/react-native/issues/50411).

### Proposed Changes

- Upgrade React-Native to version 0.76.9
- Upgrade Sentry to the latest version

### Side Effects

- Hopefully none

### Testing

Pretty much check the entire app 😬 

### Resolved Issues

This resolves the issue that I can run the iOS app on my machine again, where I stupidly already updated Xcode to the latest version.

---


package.json
```diff
@@ -2,7 +2,6 @@
   "name": "lunes",
   "version": "0.0.1",
   "private": true,
-  "packageManager": "yarn@3.2.2",
```
david-venhoff, line 5: Also why do we not need this field anymore? Sorry if it's a stupid question, but I could not find much information about it online

david-venhoff (approved):
Thanks a lot, looks good to me and still works in the emulators :)

Only question I have is why did you remove the yarnrc file and the `packageManager` field in the package.json files?

src/navigation/HomeStackNavigator.tsx
```diff
@@ -2,6 +2,7 @@ import { createStackNavigator } from '@react-navigation/stack'
 import React from 'react'
 import { useTheme } from 'styled-components'
 
+import { Theme } from '../constants/theme'
```
steffenkleinle, line 5: 🔧 Importing this from `native`avoids the need for the typecast below:
```suggestion
import { useTheme } from 'styled-components/native'

```

src/navigation/HomeStackNavigator.tsx
```diff
@@ -19,7 +20,7 @@ const Stack = createStackNavigator<RoutesParams>()
 const HomeStackNavigator = (): JSX.Element | null => {
   const options = screenOptions(useTabletHeaderHeight())
   const { manageSelection, overview } = getLabels().general.header
-  const theme = useTheme()
+  const theme = useTheme() as Theme
```
steffenkleinle, line 23:
```suggestion
  const theme = useTheme()
```

src/navigation/UserVocabularyStackNavigator.tsx
```diff
@@ -20,7 +21,7 @@ const UserVocabularyStackNavigator = (): JSX.Element | null => {
   const headerHeight = useTabletHeaderHeight(hp('7.5%'))
   const options = screenOptions(headerHeight)
   const back = getLabels().general.back
-  const theme = useTheme()
+  const theme = useTheme() as Theme
```
steffenkleinle, line 24: 🔧 Same as above
```suggestion
  const theme = useTheme()
```

f1sh1918:
@LeandraH did you check that the ios builds work in the pipeline. as far as i can see this xcode image does not exists
At least when i try it for entitlementcard it tells me that it doesn't exist

f1sh1918 (changes requested):
xcode 16.3.0 is not released for circleCI yet afaics.
Please use xcode 16.2.0 and run a `commit_main` pipeline to check if it still builds

src/navigation/HomeStackNavigator.tsx
LeandraH: Nice, thanks!

ios/Gemfile.lock
```diff
@@ -283,22 +299,21 @@ GEM
...
   activesupport (>= 6.1.7.5, != 7.1.0)
   cocoapods (>= 1.13, != 1.15.1, != 1.15.0)
+  concurrent-ruby (< 1.3.4)
   fastlane
   fastlane-plugin-browserstack
   xcodeproj (< 1.26.0)
 
 BUNDLED WITH
-   2.4.7
+   2.1.4
```
f1sh1918, line 319: :question: can you explain why we have to downgrade that much?

ios/PrivacyInfo.xcprivacy
```diff
@@ -10,6 +10,7 @@
 			<key>NSPrivacyAccessedAPITypeReasons</key>
 			<array>
 				<string>C617.1</string>
+				<string>3B52.1</string>
```
f1sh1918, line 13: :question: whats this change about?

f1sh1918 (commented):
thx for the update. I still can not find a successfully running `commit_main` where i can see if `build_ios` still runs without issues @LeandraH

ios/PrivacyInfo.xcprivacy
LeandraH: It kept being added when I built the app. The description says: "Declare this reason to access the timestamps, size, or other metadata of files or directories that the user specifically granted access to, such as using a document picker view controller." I think it's from the image picker.

ios/Gemfile.lock
LeandraH: That's a good question, thanks, I'll see if I can change that.

LeandraH:
@f1sh1918 Thank you, can you check again? I updated the bundler and ran a `build_ios` workflow [here](https://app.circleci.com/pipelines/github/digitalfabrik/lunes-app/3780/workflows/9a45df13-4f4d-4ba5-b2be-226f3fdc22ce/jobs/12216).

f1sh1918 (approved):
Looks good. Thx for the update

Issue #1072, opened 2025-03-24
App freezes when entering a word without article in WriteExercise


david-venhoff:
### Describe the Bug

In the write exercise, the app freezes when a single word (no article) is entered as answer after the submit button was pressed.

Noticed on android.

### Steps to Reproduce

1. Enable Repetition exercise 
2. Add a word to repetition
3. Click on "Jetzt wiederholen"
4. Enter just a single word, for example "a"
5. Click submit
6. Notice that the app cannot be interacted with anymore

This also happens in the already published WriteExercise

### Expected Behavior

The app shouldn't freeze

### Additional Information

The freeze is caused by `MissingArticlePopover`. If I change the code to not show the popup, everything works fine.

PR #1078, opened 2025-04-25
Fix app freeze when entering a word without article

david-venhoff:
### Short Description

Our popover library does not support reacts new architecture on older versions (https://github.com/SteffeyDev/react-native-popover-view/releases/tag/v6.0.0)

### Proposed Changes

- Update the popover library

### Side Effects

- Should be none

### Testing

Just need to test that the missing article popover shows up.
I also wanted to write a test for that, but for some reason the popover does not show up at all in the dom 🤷 

### Resolved Issues

Fixes: #1072 

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "react-native-pager-view": "^6.7.0",
     "react-native-paper": "5.12.5",
     "react-native-permissions": "^5.1.0",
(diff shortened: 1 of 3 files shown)


steffenkleinle (approved):
Not tested

Issue #1079, opened 2025-05-02
Word cut off on the favorite word view


LeandraH:
### Describe the Bug

When looking at a favorited word in the new version (where own words and favorites are merged), the word itself is cut off from the bottom.

### Steps to Reproduce

1. Be on a dev version of the app on the main branch
2. Add a word to your favorites
3. Go to Sammlung > Favoriten > your word
4. See that the word is cut off

### Expected Behavior

I would expect the word to have enough space to fit in there.

### Additional Information

![Image](https://github.com/user-attachments/assets/802496ca-dec3-4c3a-b233-4a0e69c0a1c2)

PR #1080, opened 2025-05-06
1079: Fix word cut off on favorite view on IOS

david-venhoff:
### Short Description

The favorite view had the same problem as the user vocabulary view (Fixed in commit 895bbec853b3433202daffaf01ee4268d4ed77da)

This pr applies the same fix to the favorites stack navigator.

### Proposed Changes

- In the favorite stack navigator, explicitly set the header height to 0, so that the content does not have unnecessary padding applied to it.

### Side Effects

- Should not change anything on android devices

### Resolved Issues

Fixes: #1079 

---

### DIFF
--- a/src/components/VocabularyDetail.tsx
+++ b/src/components/VocabularyDetail.tsx
@@ -7,7 +7,6 @@ import WordItem from './WordItem'
 
 const ItemContainer = styled.View`
   margin-top: ${props => props.theme.spacings.xl};
-  height: 10%;
   width: 85%;
   align-self: center;
 `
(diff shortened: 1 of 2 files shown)


LeandraH (approved):
It looks good on the bigger devices but when I test it on an old Pixel with Android API 25, it's still a little cut off. I'm not sure how far back we want to support the app though, I'll leave that up to you :)

image

david-venhoff:
@LeandraH Good catch, I have now also removed the manual limit of the height to 10% of the screen height. As far as I understand, that is not needed

Issue #1083, opened 2025-05-14
Speaker button not greyed out in exercises


david-venhoff:
### Describe the Bug

In exercises, the speaker button is not greyed out, even when it is disabled. This makes it impossible to see whether it will work when clicked:

![Image](https://github.com/user-attachments/assets/68789020-d815-473b-81a3-590640c05510)

### Steps to Reproduce

1. Go to Exercise
2. Before answering, click the speaker button
3. Notice that nothing happens

### Expected Behavior

The speaker button should be greyed out if it is disabled

PR #1084, opened 2025-05-14
Fix volume icon not greyed out correctly

david-venhoff:
### Short Description

Turns out we were already using the correct colors for the icon, but they were fully covered by the background of the icon…

This pr simplifies the svg icon by removing the background from it.

### Proposed Changes

- Remove background from assets/images/volume-up-circle-icon.svg

### Side Effects

- The colors that were already being applied to the background should now show up

### Resolved Issues

Fixes: #1083 

---


LeandraH (approved):
Good find! Not tested

Issue #1082, opened 2025-05-13
Email from Apple: Invalid purpose string value for NSLocationWhenInUseUsageDescription


LeandraH:
### Describe the Bug

I got an email from Apple that our Lunes build has an issue:

Hello,

We noticed one or more issues with a recent delivery for the following app:

  Lunes
  App Apple ID 1562834995
  Version 2025.5.0
  Build 100000097

    Although delivery was successful, you may want to correct the following issues in your next delivery. Once you've corrected the issues, upload a new binary to App Store Connect.
  
	ITMS-90738: Invalid purpose string value -  The “” value for the NSLocationWhenInUseUsageDescription key isn’t allowed in “Lunes.app”. Your app’s code references one or more APIs that access sensitive user data, or the app has one or more entitlements that permit such access. The Info.plist file for the “Lunes.app” bundle should contain a NSLocationWhenInUseUsageDescription key with a user-facing purpose string explaining clearly and completely why your app needs the data. For details, visit: https://developer.apple.com/documentation/uikit/protecting_the_user_s_privacy/requesting_access_to_protected_resources.

    Apple Developer Relations

### Additional Information

<img width="817" alt="Image" src="https://github.com/user-attachments/assets/92c36750-f6e4-461c-97d7-05e35d8009fd" />

david-venhoff:
@LeandraH Do you have an idea why this key exists at all in our `Info.plist`? Lunes does not actually use location information, right?

If so, the solution would be to just delete this key?

LeandraH:
I have no idea, we really shouldn't be using the location, and if we need to, we certainly shouldn't be using an empty string as an explanation, it's very odd. I would expect deleting the key to fix the problem, yes.

PR #1094, opened 2025-05-24
Remove unused ios permission

david-venhoff:
### Short Description

- We don't need the location permission for the lunes app, and the android variant does also not declare it
- Not sure why it was here in the first place though

### Proposed Changes

- Remove the permission from the ios manifest

### Side Effects

- should be none

### Testing

App should still work

### Resolved Issues

Fixes: #1082

---


LeandraH (approved):
Looking at the git blame, I get the feeling that it was copied from somewhere when setting up the app in the first place 😅

Issue #1098, opened 2025-06-16
Listening to audios while IOS is in silent mode


HuguetteAfi:
### Describe the Bug

When silent mode is activated in IOS, the audio of the vocabulary is not played. @f1sh1918 did some troubleshooting finding out that there is no possibility to play the sound with the current library anyway, there would only be the possibility to find out with an extra library that silent-switch is activated and to point it out. https://www.npmjs.com/package/react-native-volume-manager/v/1.8.0

You would probably need a snackbar or something like that where you can display a message when user:sounds in the volume symbol
https://www.npmjs.com/package/react-native-volume-manager/v/1.8.0#usesilentswitch-hook

### Steps to Reproduce

When IOS user want ro listen to the audio of a word while on silent mode, they can't hear anything.

### Expected Behavior

IOS User should obtain a notication that they are in the silent mode.

HuguetteAfi:
Is the problem with android to0?

PR #1100, opened 2025-06-18
1098: Play sound in silent mode on iOS

LeandraH:
### Short Description

This fixes the problem that the words were not being read aloud on iOS if the device was on silent mode.

### Proposed Changes

- Set the category of our audio to playback, which ensures that the silent mode is ignored.

### Side Effects

- Should be none

### Testing

Use an iPhone, neither @david-venhoff nor I managed to reproduce the problem in the simulator. Set the iPhone to silent mode with the switch on the side. Then go to e.g. the Suche and tap on the loudspeaker symbol next to any word. You should now hear it pronounced.

### Resolved Issues

Fixes: #1098 

---

### DIFF
--- a/release-notes/unreleased/1098-play-sound-in-silent-mode-on-ios.yml
+++ b/release-notes/unreleased/1098-play-sound-in-silent-mode-on-ios.yml
@@ -1,0 +1,5 @@
+issue_key: 1098
+show_in_stores: true
+platforms:
+  - ios
+de: Die Vokabeln können jetzt auch im Lautlos-Modus vorgelesen werden


david-venhoff:
As far as I am aware, the app should not play sound automatically, without requiring a click on the speaker icon, so imo it is okay to bypass the silent mode (But I don't have a hard opinion on it). 

@HuguetteAfi Do you have an opinion on this?

steffenkleinle:
Just in case you change your mind and want to ignore the silent mode in the future, there also seems to be a solution by the library to do so:
https://github.com/ak1394/react-native-tts?tab=readme-ov-file#controls-the-ios-silent-switch-behavior

PR #1101, opened 2025-06-27
Detect muted audio and display mute icon

david-venhoff:
### Short Description

This pr tries to detect the silence mode and muted volume and displays a different icon and an error message if that is the case.

### Proposed Changes

- Add `react-native-volume-manager` dependency
- Add context that keeps track of volume and silent switch state
- Display a different icon if the sound is muted
- Display an error message in the vocabulary exercise views

### Side Effects

- The iPhone emulator does not report the current sound, which makes it impossible to test this feature on it
- I did not add an error message to the list view ("Suche") and the audio recorder, because it did not find a good place to add the error message. That should not be a problem though I think

### Testing

Mostly needs testing on a real iPhone

### Resolved Issues

Fixes: #1098 

---


f1sh1918:
@david-venhoff the detection works properly.
I guess you haven't finished implementation yet. Some kind of info in addition to the changed sound icon would be great.
We may have some iPhone or iPad which you could have for testing purposed. I think that would make sense, since for LUNES you also have to test camera, etc at some point

david-venhoff:
Alright, thanks for testing!
Yes, I will still have to complete this pr.
I think having an iPhone would be useful, maybe that's something we can talk about at the next conference

src/components/AudioPlayer.tsx
```diff
@@ -97,11 +99,15 @@ const AudioPlayer = ({ audio, disabled, isTtsText = false }: AudioPlayerProps):
...
   return (
     <VolumeIcon
-      disabled={disabled || !isInitialized}
+      disabled={disabled || !isInitialized || isSilent}
       isActive={isActive}
       onPress={handleSpeakerClick}
       testID='audio-player'>
-      <VolumeUpCircleIcon width={theme.spacingsPlain.lg} height={theme.spacingsPlain.lg} />
+      {isSilent ? (
+        <VolumeDisabled width={theme.spacingsPlain.md} height={theme.spacingsPlain.md} />
```
f1sh1918, line 107: Just a small recommendation. I think we started once with using React Native Paper which is similar to MUI
There are several nice and lean component such as a tooltip.
Not sure if that fits design needs

```
        <Tooltip enterTouchDelay={0} title={getLabels().general.error.deviceIsMuted} leaveTouchDelay={2600}>
        <VolumeDisabled width={theme.spacingsPlain.md} height={theme.spacingsPlain.md} />
        </Tooltip>
```

f1sh1918 (approved):
Nice implementation 👍 
Works like a charm
Tested on iPad

Issue #1088, opened 2025-05-23
Move words to long term memory exercise after each level


david-venhoff:
### Describe the Problem

Currently, this only happens after level 3

PR #1102, opened 2025-06-27
Move words to long term memory exercise after first level

david-venhoff:
### Short Description

This pr changes the behavior of when new words are added to the long term memory exercise. Before, this only happened when every level was completed and now it will happen after the second level (every level except the word list)

### Proposed Changes

- Move the functionality to add words to the long term memory exercise to the `saveExerciseProgress` function
- Adjust tests
- Adjust when the repetition reminder is shown

### Side Effects

- Should be none

### Resolved Issues

Fixes: #1088

---

### DIFF
--- a/release-notes/unreleased/1088-move-words-to-long-term-memory-exercise-after-each-level.yml
+++ b/release-notes/unreleased/1088-move-words-to-long-term-memory-exercise-after-each-level.yml
@@ -1,0 +1,6 @@
+issue_key: 1088
+show_in_stores: true
+platforms:
+  - android
+  - ios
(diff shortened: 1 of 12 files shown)


f1sh1918:
Alright. Our prior intention for the long term exercises was to build them upon the regular exercises

f1sh1918:
> In our team call we have decided to go ahead with this, because that will allow users to choose if they want to continue the normal exercises or if the want to do the repetition, depending on what they prefer.
> 
> @f1sh1918 If you are okay with this, could you do a review on this again?

Well thats your decision. I already reviewed and tested it

f1sh1918 (approved):
Tested on ios. 👍

Issue #1104, opened 2025-07-02
Increase android target api to 35


david-venhoff:
### Describe the Problem

Google sent a warning that the target api must be increased to 35 until the 31.08.

### Describe Possible Alternatives

If we do nothing, we will not be able to submit updates until we increase the target sdk

PR #1105, opened 2025-07-04
Update target sdk to 35

david-venhoff:
### Short Description

This is required by google, so that we can continue pushing updates

### Proposed Changes

- Update the target sdk to 35
- Update `react-native-safe-area-context`
- Fix calculation of bottom tab bar height

### Side Effects

- As a side effect of android sdk 35, edge to edge navigation is now enabled by default. To me everything looks fine (except the one small bug with the height calculation of the bottom tab bar height that I fixed), but worst case we could also disable it again
- Potentially other side effects of the new sdk, but I did not find any

### Resolved Issues

Fixes: #1104

---

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         buildToolsVersion = "35.0.0"
         minSdkVersion = 24
         compileSdkVersion = 35
-        targetSdkVersion = 34
+        targetSdkVersion = 35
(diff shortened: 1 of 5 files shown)


f1sh1918 (approved):
@david-venhoff i updated `podfile.lock` for ios (Safe Area Context updates are missing)

Issue #1106, opened 2025-07-07
Android 15 Navigation bar overlapping layout


f1sh1918:
### Describe the Bug

Android navigation buttons are overlapping (android 15 only)

### Steps to Reproduce

1. Enable navigation buttons for android
2. Start the app
3. See navigation buttons overlaying (see screenshot)

### Expected Behavior

Buttons will not overlay over the app layout

### Additional Information

<img width="1440" height="3120" alt="Image" src="https://github.com/user-attachments/assets/efdae83d-fc56-44bf-a98c-4cbc3a4ee9d7" />

david-venhoff:
This is caused by #1105, because the edge to edge mode is now enabled by default which means we will be allowed to draw on the entire screen. But I am not sure what the bug is. Even the android settings do that in android 15

david-venhoff:
Ah, it definitely needs some more padding at the bottom

PR #1109, opened 2025-07-11
Fix overlap of content on android

david-venhoff:
### Short Description

This pr fixes some problems caused by #1105 where the content starts to overlap the system navigation bar on android.

### Proposed Changes

- For each screen not part of the bottom tab navigator, add some bottom padding manually

### Side Effects

- Should be none

### Testing

The modified screens should have some more padding, on no screen should content overlap with the android navigation bar

### Resolved Issues

Fixes: #1106 

---


f1sh1918:
Hey @david-venhoff,
we have a component thats called `RouteWrapper` which was likely designed to handle `Layout` issues and is reusable.
This component has a `Container` which is a `SafeAreaView` and there should be handled needed paddings i think

PR #1110, opened 2025-07-14
1106: Android content overlay

f1sh1918:
### Short Description

With android 15 in some parts, `SafeArea` will get handled differently and we need to define some paddings for particular screens to avoid overlapping

### Proposed Changes

- add `shouldSetBottomInset` and `shouldSetTopInset` to `RouteWrapper`
- use the inset on routes if needed

### Side Effects

- likely none

### Testing

- start application with Android Bottom Navigation instead of gesture navigation
- Check `HomeScreen`, `Scope`, `ProfessionSelection` and `VocabularyList` for content overlapping or other glitches
- Test with android and ios

### Resolved Issues

Fixes: #1106

---

### DIFF
--- a/src/components/RouteWrapper.tsx
+++ b/src/components/RouteWrapper.tsx
@@ -1,5 +1,6 @@
 import React, { ReactElement, ReactNode } from 'react'
 import { Platform, StatusBar } from 'react-native'
+import { useSafeAreaInsets } from 'react-native-safe-area-context'
 import styled from 'styled-components/native'
 
 import theme from '../constants/theme'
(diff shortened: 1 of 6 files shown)


f1sh1918:
> I initially decided to implemented the fix on each screen individually so that the content of the lists is visible behind the system navigation buttons. But most users probably would not notice the difference and I think that this solution is nicer to maintain and a bit cleaner code-wise, so lgtm.
> 
> I also added a commit to fix the overlay menu (which I missed in my original pr), by adding some top padding there as well.

Alright thx for the fix of the `Sidebar`

PR #1116, opened 2025-07-18
Optimize startscreen layout

david-venhoff:
# Description

Cleans up the startscreen layout to make the ui a bit cleaner and to make it easier to add more content to the startscreen in the future.

<p align="middle">
image
image
</p>

### Proposed Changes

- Show discipline cards in a horizontal list
- Add edit and new discipline icons
- Remove 'Lernbereiche verwalten' from the menu, since it is not needed anymore
- Also show custom disciplines in the list, next to normal disciplines
- If there is only one profession, show it full-width

### Deviations from the design
- Use a pencil icon instead of an edit icon, because that better matches the behavior
- Keep 'Module ansehen' button, because there is also a 'Start' button
- Show single profession full-width

### Side Effects

- Should only affect the home screen

### Testing

Test that the home screen still works and all previous functionality is still accessible

### Resolved Issues

Fixes: #926 

---


f1sh1918 (changes requested):
Maybe i've overseen something but if i delete all professions, there is no way to add one again?

image

david-venhoff:
Oof, good catch!
Should no be fixed in f39e45ba91488a2c69d55db6abd2de8dad44be8d

Issue #1127, opened 2025-08-01
Remove browserstack from ci


david-venhoff:
### Describe the Problem

We don't need it anymore and there's no reason to spend circleci credits on it

### Describe Your Preferred Solution

Remove browserstack from ci and fastlane

### Describe Possible Alternatives

Do nothing

PR #1130, opened 2025-08-08
CI: Remove `upload_to_browserstack`

david-venhoff:
### Short Description

We don't use it very often, so there is no reason to spend circleci credits on it

### Proposed Changes

- Remove all code related to browserstack

### Side Effects

- Should be none, except that new builds are not uploaded to browserstack anymore.
- If we want to use browserstack in the future, we can simply upload a build manually

### Resolved Issues

Fixes: #1127

---


LeandraH (approved):
Looks great!

PR #1135, opened 2025-08-15
Handling of missing tts engine

david-venhoff:
### Short Description

Some fixes regarding the tts functionality

### Proposed Changes

- Add a `query` to android manifest so that there are no more false negatives about whether tts functionality is available
- Only initialize the tts engine once, so that we don't get spammed with errors when debugging
- Only open the google play store when the user clicks on the icon

### Side Effects

- Should be none

### Testing

Test that the sound player and tts player still work on android and ios

### Resolved Issues

Fixes: #959 

---


android/app/src/main/AndroidManifest.xml
LeandraH, line 1: Ooooh, nice catch, I wonder if the lack of this intent declaration is one of the reasons why we keep getting TTS errors in the Integreat app...

LeandraH (approved):
Nice job, tested on real iOS

android/app/src/main/AndroidManifest.xml
david-venhoff: I really can't say I understand what exactly it does. On the android emulator it worked perfectly without the query for me, but on my personal phone (which is set up a bit nonstandard) tts never works without it, and adding the query fixed it...

So i guess either way its a good idea to also add it to integreat :)

PR #1150, opened 2025-08-30
Prune dependencies

david-venhoff:
### Short Description

prune some dependencies we don't appear to be using anymore. Also update dependencies where no code changes are required.

### Proposed Changes

Pruned dependencies:
- react-native-masked-view: Don't know what it was used for, appears unused now
- @sentry/types: Not required anymore
- react-native-reanimated: Unused
- react-native-vector-items: Unused

### Side Effects

- Slightly smaller download size
- Slightly faster compile times

### Testing

Unfortunately I did not have time to test this yet, will do so after the vacation. I published this pr in case somebody already wants to review it

### Resolved Issues

Fixes: /

---


LeandraH (commented):
Wow, I like those side effects!

Looks like `react-native-masked-view` was added erroneously to begin with 🤪 https://github.com/digitalfabrik/lunes-app/commit/28a56da080536da7fd9c48c18a27c6c0d43f2834

But also, does this run for you? I only got it to run on emulated or real iOS once I downgraded `react-native-screens`, which seems to not support RN 76: https://github.com/software-mansion/react-native-screens/issues/3156

LeandraH (approved):
Yup, tested on simulated and real iOS, works :)

Issue #1155, opened 2025-09-18
Update react native to 0.77


david-venhoff:
### Describe the Problem

We should update react native to 0.77 for googles 16kb pages policy

### Describe Your Preferred Solution

Update react native to 0.77

### Describe Possible Alternatives

Alternatively, we could also directly update to 0.78 or newer, which brings in support for react actions, which would also be a nice to have.

Issue #1152, opened 2025-09-12
Make android app 16kb page size compliant


steffenkleinle:
### Describe the Problem

Google requires apps to be 16kB page size compliant starting 01.11.2025. While its possible to request an extension in the Google Play Console, we should make our app compliant.
More information: https://play.google.com/console/u/0/developers/5269030016285360668/app/4973131027931283146/policy-center/issues/4989329769159092809/details

### Describe Your Preferred Solution

Libraries that do not support 16 KB:
base/lib/arm64-v8a/libVisionCamera.so
base/lib/arm64-v8a/libappmodules.so
base/lib/arm64-v8a/libc++_shared.so
base/lib/arm64-v8a/libfbjni.so
base/lib/arm64-v8a/libgesturehandler.so
base/lib/arm64-v8a/libhermes.so
base/lib/arm64-v8a/libhermestooling.so
base/lib/arm64-v8a/libimagepipeline.so
base/lib/arm64-v8a/libjsi.so
base/lib/arm64-v8a/libnative-filters.so
base/lib/arm64-v8a/libnative-imagetranscoder.so
base/lib/arm64-v8a/libreact_codegen_rnpicker.so
base/lib/arm64-v8a/libreact_codegen_rnscreens.so
base/lib/arm64-v8a/libreact_codegen_rnsvg.so
base/lib/arm64-v8a/libreact_codegen_safeareacontext.so
base/lib/arm64-v8a/libreactnative.so
base/lib/arm64-v8a/libreanimated.so
base/lib/arm64-v8a/librnscreens.so
base/lib/arm64-v8a/libworklets.so
base/lib/x86_64/libVisionCamera.so
base/lib/x86_64/libappmodules.so
base/lib/x86_64/libc++_shared.so
base/lib/x86_64/libfbjni.so
base/lib/x86_64/libgesturehandler.so
base/lib/x86_64/libhermes.so
base/lib/x86_64/libhermestooling.so
base/lib/x86_64/libimagepipeline.so
base/lib/x86_64/libjsi.so
base/lib/x86_64/libnative-filters.so
base/lib/x86_64/libnative-imagetranscoder.so
base/lib/x86_64/libreact_codegen_rnpicker.so
base/lib/x86_64/libreact_codegen_rnscreens.so
base/lib/x86_64/libreact_codegen_rnsvg.so
base/lib/x86_64/libreact_codegen_safeareacontext.so
base/lib/x86_64/libreactnative.so
[...]

PR #1156, opened 2025-09-19
Update react-native to 0.77

david-venhoff:
### Short Description

This pr updates react native to 0.77.3, in preparation for #1152.

Also, I accidentally commit this to main 🙈 
The changes on main are reverted now, and I modified the branch protection rules so that even administrators cannot push to main, to prevent such accidents in the future. Sorry for that!

### Proposed Changes

- Update AppDelegate.mm as required
- Don't use swift, since it is not required yet
- Pin react-native-svg, to avoid a compilation error
- Remove react-native-vector icons, which is unused and also caused upgrade problems

### Side Effects

Ideally none

### Testing

Please test that the main dependencies still work, I found no problems so far

### Resolved Issues

Fixes: #1155 

---


LeandraH (approved):
Nice! Tested on real iOS

PR #1167, opened 2025-10-06
Small fixes to the repetition wordlist design

david-venhoff:
### Short Description

@nikolahoff found two issues with #1159 

### Proposed Changes

- Remove the shadow
- Add small padding

### Side Effects

Should be none

### Resolved Issues

Fixes: #1165 

---


LeandraH (approved):
Nice, tested in emulated iOS

PR #1169, opened 2025-10-06
Revert "Move release notes to 2025.10.0"

david-venhoff:
### Short Description

This reverts commit fd9259496899980ba96bc3c688e89f261ec02fce.

Reason is that we will not promote this beta release to production, so the release notes will have to be moved to the `unreleased` folder again, so that the release notes for the next full release are correct

~I will create another pr to change the release workflow, so that reverting the release notes will not have to be done manually again~
Not sure actually what a better workflow would look like... Ideally it should be easy to create an arbitrary number of pre-releases and then a final release that contains the release notes of all pre-releases. How does the app-team do releases?

### Proposed Changes

- Revert the commit 

### Side Effects

none

### Resolved Issues

Fixes: /

---


david-venhoff:
Alright, I guess the current workflow is fine for now, I don't expect additional beta release to happen too often.

Issue #1141, opened 2025-08-29
Meta: Migrate to new cms api model


david-venhoff:
### Describe the Problem

When the cms api gets reworked, we'll need to migrate to it.

Depends on #1071, so that users will know why their progress is lost

PR #1173, opened 2025-10-11
Use api v2

david-venhoff:
### Short Description

Pr to switch to the v2 api of the cms.
This pr removes all v1 endpoints and converts them to the v2 of the api.
This also brings in some model changes and migrations to migrate from the old data format to the new format, but
the users should not see any significant changes, except where absolutely required. For example, api v2 has no concept of job categories anymore, so the job selection screen will now show a list of all jobs instead.

This pr also changes the technical terms used in the app code to be consistent with the new api v2 terms. The changes should be as follows:
- `Discipline` -> `Job` or `Unit`, depending on the context
- `VocabularyItem` -> `Word`
- `Profession` -> `Job`

### Proposed Changes

- Add localhost cms server url to simplify implementation & testing
- Remove custom disciplines #1187 
- Migrate the words endpoint to v2 #1180 
- Migrate the jobs endpoint to v2 #1183 
- Migrate the sponsors endpoint to v2 #1194 
- Migrate the units-of-jobs and words-of-units endpoints to v2 #1191 
- Migrate the feedback endpoint #1195 
- Simplify the job search screen: #1212 
- Some refactors (#1223, #1222, #1225, ...)

### Side Effects

- The job list is now more annoying to use, because a first time user has to scroll down to the end to click the confirm button. I think we should merge this pr rather sooner than later, and this is something that can be cleaned up in a later pr. When this is merged, I'll open an issue for it.

### Testing

- Test that the migration works: Start the app in the previous version, fill it with some data, switch to this pr and verify that all jobs, progress, favorites, wordNodeCards and custom vocabularyItems still exist
- Test that having custom disciplines and then migrating to this pr does not break anything

### Resolved Issues

Fixes: #1141 

---

### DIFF
--- a/src/models/VocabularyItem.ts
+++ b/src/models/VocabularyItem.ts
@@ -52,6 +52,18 @@ export const isUserVocabularyItem = (vocabularyItem: VocabularyItem): vocabulary
 export const areVocabularyItemIdsEqual = (
   vocabularyItemId1: VocabularyItemId,
   vocabularyItemId2: VocabularyItemId,
-): boolean => JSON.stringify(vocabularyItemId1) === JSON.stringify(vocabularyItemId2)
+  // eslint-disable-next-line consistent-return
+): boolean => {
+  switch (vocabularyItemId1.type) {
+    case 'lunes-standard':
+      return vocabularyItemId2.type === 'lunes-standard' && vocabularyItemId1.id === vocabularyItemId2.id
+    case 'user-created':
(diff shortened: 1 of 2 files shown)


david-venhoff:
Still to do:
- [ ] Wait for the production CMS to migrate all old jobs
- [ ] Add release note
- [ ] Do some testing and decide on whether it is feasible to keep user progress
- [ ] Otherwise, add note about potentially lost progress

Issue #1143, opened 2025-08-29
New Api: Migrate endpoints for jobs 


david-venhoff:
### Describe Your Preferred Solution

When the new api is implemented on the cms, we will need to migrate our endpoints for jobs.
This will also include updating all the model type definitions and migrating the storage.
Additionally, a lot of screens will change or need to be updated, as some concepts have been removed / simplified

PR #1183, opened 2025-10-15
Migrate endpoints for jobs

david-venhoff:
### Short Description

Migrates getting all jobs and a specific job by id to use api v2.
The new api has no more top level disciplines, so we can remove the `ScopeSelection` screen and instead display a list of all jobs.
The new api also has no more concept of custom disciplines (Eigene Lernbereiche) until we decide what to do with them. For now I just fake a network error if some part of the code tries to load a custom discipline, but that will of course have to change still.

It might also make sense to combine the job search with the normal job list now to get a more unified ui, but that's something for another pr.

### Proposed Changes

- Rename professions to jobs to match the new CMS terminology
- Combine the `ScopeSelection` and `DisciplineSelection` screens
- Update the following components which used the old api: `JobSelectionScreen`, `JobCard`, `SelectionItem`
- Remove the now unused `useLoadDiscipline` hook
- Don't remove `useLoadDisciplines` for now, because that is still used to get all units of a job

### Side Effects

- As with all other api migration prs, this leaves the app in a non-functional state, for this reason this pr does not merge into main

### Testing

~Currently the required api endpoint is not on the cms test system yet, so testing will require either waiting or a local cms installation~
The test server can now be used for testing

### Resolved Issues

Fixes: #1143 

---

### DIFF
--- a/src/routes/home/__tests__/HomeScreen.spec.tsx
+++ b/src/routes/home/__tests__/HomeScreen.spec.tsx
@@ -16,8 +16,6 @@ jest.mock('../../../services/helpers', () => ({
 }))
 jest.mock('@react-navigation/native')
 jest.mock('../../../hooks/useReadProgress')
-jest.mock('../../../hooks/useLoadDisciplines')
-jest.mock('../../../hooks/useLoadDiscipline')
 jest.mock('../../../services/CmsApi')
 jest.mock('../components/HomeScreenHeader', () => {
   const Text = require('react-native').Text


src/services/CmsApi.ts
```diff
@@ -0,0 +1,44 @@
...
+      apiKey: string
+    }
+  | {
+      disciplineId: number
+    }
+
+export const getJob = async (id: JobId): Promise<Discipline> =>
+  !isTypeLoadProtected(id)
+    ? transformJobsResponse(await getFromEndpoint<JobResponse>(Endpoints.job(id.disciplineId)))
+    : Promise.reject(new Error(NetworkError)) // TODO: Add support back to the cms
```
LeandraH: This made it pretty hard to test since the app crashed on opening for me 🙃

Issue #1185, opened 2025-10-20
New Api: Migrate endpoint for units


david-venhoff:
### Describe the Problem

We will need to migrate to the new endpoint to get all units of a job

### Describe Your Preferred Solution

Migrate the api

### Describe Possible Alternatives

none

david-venhoff:
Completed by #1191

Issue #1172, opened 2025-10-11
New api: Migrate enpoint for all words of a unit


david-venhoff:
### Describe the Problem

migrate to the new endpoint for getting all words of a unit

PR #1191, opened 2025-10-28
APIv2: Add endpoints for units of jobs and words of units

david-venhoff:
### Short Description

This pr updates the endpoints used for getting the units of a job and the words of a unit.
I decided to update both in a single pr because these endpoints are used together a lot.

### Proposed Changes

- Rename many instances of `Discipline` to either `Job` or `Unit`
- Add the endpoints
- Add new hooks to access the endpoints
- Remove now outdated hooks
- Make `useReadProgress` async, because the api no longer returns all units of a job
- Create Unit model in new directory `models/unit.ts`
    - Differentiate between standard units and user vocabulary units in the type system, so that the ids cannot be mixed up anymore accidentally
    - This uncovered some bugs where ids of standard units and user vocabulary units got mixed up!
    - For example, in user vocabulary exercises, progress is ignored. However, it still got updated and would have changed standard units with the same id.

### Side Effects

- This should make the app usable again! Only missing is feedback and sponsors :)

### Testing

Can be tested on the test cms now

### Resolved Issues

Fixes: #1172 
Fixes: #1185 

---

### DIFF
--- a/src/components/DisciplineListItem.tsx
+++ b/src/components/DisciplineListItem.tsx
@@ -3,6 +3,7 @@ import * as Progress from 'react-native-progress'
 import { heightPercentageToDP as hp } from 'react-native-responsive-screen'
 import styled from 'styled-components/native'
 
+import { EXERCISES } from '../constants/data'
 import { Discipline } from '../constants/endpoints'
 import labels from '../constants/labels.json'
 import theme from '../constants/theme'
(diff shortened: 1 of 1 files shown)


src/hooks/useReadProgress.ts
```diff
@@ -1,10 +1,13 @@
...
 import { Discipline } from '../constants/endpoints'
 import { getProgress } from '../services/helpers'
+import { Return, useLoadAsync } from './useLoadAsync'
 import useStorage from './useStorage'
 
-const useReadFromAsyncStorage = (profession: Discipline | null): number => {
-  const [progress] = useStorage('progress')
-  return getProgress(progress, profession)
-}
+const useReadFromAsyncStorage = async ({ job, progress }: { job: Discipline; progress: Progress }): Promise<number> =>
```
sascha11110: Is there a reason why `type Discipline` is still called `Discipline`?

sascha11110 (approved):
Code Review ✅
Test on iOS ✅

Issue #1193, opened 2025-10-31
New Api: Migrate feedback


david-venhoff:
### Describe the Problem

Currently the new cms API has no endpoinst for feedback, so we either need to use the old endpoint or add one to the new api

david-venhoff:
Fixed by #1195

PR #1195, opened 2025-10-31
1193 migrate feedback

david-venhoff:
### Short Description

This pr migrates the feedback endpoint to the new API, which should be the last endpoint that needs to be migrated.

### Proposed Changes

- Add endpoint to `CmsAPI.ts`
- Add feedback model
- Fix bug where it was possible to create feedback about user vocabulary & units

### Side Effects

should be none

### Testing

Can be tested on the lunes test CMS server

### Resolved Issues

Fixes: #1193 

---


LeandraH (approved):
Nice! One question, where can I give feedback for jobs?

Tested on simulated iOS

Issue #1196, opened 2025-11-04
Misleading behaviour for selecting a new profession


nikolahoff:
### Describe the Problem
if a user wants to add a new profession via the search function it is not clear if the selected profession is already added or not. The searched profession is greyed out after selecting. It is not clear is the profession is already added by then.

<img width="1179" height="2556" alt="Image" src="https://github.com/user-attachments/assets/6d4fbe13-5574-4075-b703-dd9fab2ec560" />
<img width="1179" height="2556" alt="Image" src="https://github.com/user-attachments/assets/0e27fcc2-ae77-4a99-a862-0519a59e3380" />
<img width="1179" height="2556" alt="Image" src="https://github.com/user-attachments/assets/6421b83d-6cd4-41a4-a4ac-07598242b863" />

### Describe Your Preferred Solution

We need to implement a success information to the user after selecting the profession. ie. "Pflegefachmann / frau was added to your professions."
### Describe Possible Alternatives

Define the userflow for search - enter profession - select profession - show overview screen

### Additional Information

s. screenshot

nikolahoff:
@david-venhoff please leave a comment when it is done! Thanks:)

PR #1212, opened 2025-11-20
Simplify profession search screen

david-venhoff:
### Short Description

Now that there are no more groups of professions, there is no reason to not use the normal job list when searching. This also improves the UI.

<p>
image
image
</p>

### Proposed Changes

- Remove the `FilteredJobList` component
- Add its functionality to the job list

### Side Effects

- Should be none

### Resolved Issues

Fixes: #1196 

---

### DIFF
--- a/src/services/__tests__/helpers.spec.ts
+++ b/src/services/__tests__/helpers.spec.ts
@@ -18,7 +18,6 @@ import {
   getProgress,
   getSortedAndFilteredVocabularyItems,
   searchJobs,
-  splitTextBySearchString,
   willNextExerciseUnlock,
 } from '../helpers'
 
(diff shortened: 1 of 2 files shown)


src/routes/job-selection/JobSelection.tsx
```diff
@@ -5,93 +5,31 @@ import { CheckCircleIconGreen } from '../../../assets/images'
 import { JobListItem } from '../../components/DisciplineListItem'
 import SearchBar from '../../components/SearchBar'
 import ServerResponseHandler from '../../components/ServerResponseHandler'
-import { ContentTextBold, ContentTextLight } from '../../components/text/Content'
 import { Discipline } from '../../constants/endpoints'
 import useLoadAllJobs from '../../hooks/useLoadAllJobs'
-import useStorage, { useStorageCache } from '../../hooks/useStorage'
-import { getLabels, searchJobs, splitTextBySearchString } from '../../services/helpers'
```
LeandraH: :wrench: Looks like we can delete `splitTextBySearchString` 🥳

Issue #1207, opened 2025-11-12
[T41K-2] Refactor user-vocabulary-units


david-venhoff:
### Describe the Problem

The code quality for generating units is suboptimal.

### Describe Your Preferred Solution

Rewrite it to be a bit cleaner to read and avoid the global state shenanigans

### Describe Possible Alternatives

Do nothing and accept technical debt

PR #1223, opened 2025-11-26
Refactor user vocabulary units

david-venhoff:
### Short Description

Small refactor to the 'splitVocabularyToUnits.ts' file.
I wanted to avoid ugly rebases, so I hope it is okay that this merges into `use-api-v2` 😁 

### Proposed Changes

- Don't use global state
- Rely less on mutation

### Side Effects

The app behavior should not change

### Resolved Issues

Fixes: #1207 

---


LeandraH:
> I wanted to avoid ugly rebases, so I hope it is okay that this merges into `use-api-v2` 😁

Yes, I'm very happy with that :)

Issue #1208, opened 2025-11-12
[T41K-3] Change the name of `VocabularyItem` to `Word`


david-venhoff:
### Describe the Problem

To better match the new terminology. 

### Describe Your Preferred Solution

Can also be used as an opportunity to slightly refactor this type:
* Remove `alternatives`, since this field is not sent out anymore
* Combine `type` and `id`, as is done in `Unit` to improve type safety
* Probably makes sense to do together #1176 to simplify the migration

### Describe Possible Alternatives

Do nothing and accept technical debt

PR #1225, opened 2025-11-28
Refactor `VocabularyItem`

david-venhoff:
### Short Description

This pr moves the `VocabularyItem` type to the models directory and adds a refactor to its id.

I deviated a bit from the issue (#1208):
- Did not change the name from `VocabularyItem` to `Word`: Even though `VocabularyItem` does not match the new terminology exactly, it might be a better name than `Word`, because it also has a field called `word`. Also renaming it would produce an even bigger diff, since we'd probably also want to rename all variable names then.
- Did not remove the `alternatives` field. I figured it was not causing much additional complexity and might be useful again in the future

I think this should be the last issue which touches a lot of files :)

### Proposed Changes

- Move `VocabularyItem` to the `models` directory
- Extract all fields that make up its id into the new id field. This will improve type safety and make it harder to misuse words, e.g. by accidentally allowing submitting feedback about custom vocabulary or by accidentally using the same id for different words in list views.
- Make favorites be the same type as `VocabularyItemId`
- Store user vocabulary with its id, just to make it more ergonomic
- Add a migration

### Side Effects

- Should be none

### Resolved Issues

Fixes: #1208

---

### DIFF
--- a/src/routes/choice-exercises/components/SingleChoiceExercise.tsx
+++ b/src/routes/choice-exercises/components/SingleChoiceExercise.tsx
@@ -19,7 +19,7 @@ import {
 } from '../../../constants/data'
 import { useStorageCache } from '../../../hooks/useStorage'
 import { StandardUnitId } from '../../../models/Unit'
-import VocabularyItem, { AlternativeWord } from '../../../models/VocabularyItem'
+import VocabularyItem, { AlternativeWord, VocabularyItemTypes } from '../../../models/VocabularyItem'
 import { RoutesParams, VocabularyItemResult } from '../../../navigation/NavigationTypes'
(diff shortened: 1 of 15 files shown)


src/routes/choice-exercises/components/SingleChoiceExercise.tsx
```diff
@@ -155,7 +155,7 @@ const ChoiceExerciseScreen = ({
         currentWord={currentWord}
         numberOfWords={count}
         feedbackTarget={
-          vocabularyItem.type === 'user-created' ? undefined : { type: 'word', wordId: vocabularyItem.id }
+          vocabularyItem.id.type === 'lunes-standard' ? { type: 'word', wordId: vocabularyItem.id } : undefined
```
LeandraH: 🔧 Same here

Issue #1184, opened 2025-10-20
[T41K-5] Hooks are named inconsistently


david-venhoff:
### Describe the Bug

We have some common prefixes of the names of our hooks:
- `useGet*`
- `useLoad*`
- `useRead*`
- `use*`

However, there is no difference between `useGet*` and `useLoad*`, both names indicate that the hook has a loading state were no data is loaded yet. I would propose to consistently use only `useLoad` to indicate that a hook needs to load some data asynchronously.

Also our `useRead*` hooks seem to follow no convention at all, because some return data immediately and some have to load asynchronously. For this reason, I'd also suggest to rename `useRead*` hooks to use either the `useLoad*` or just `use*` convention

PR #1226, opened 2025-12-01
Make hook naming more consistent

david-venhoff:
### Short Description

This pr adapts a simple naming convention for hooks, so that their naming is a bit more consistent:
- If a hook is async (returns the `Return<T>` type), it should be named `useLoadValue`
- Otherwise, it should be named `useValue`

### Proposed Changes

- Rename existing hooks that don't match this convention
- Delete `useReadNextExercise`, because it is apparently unused

### Side Effects

- Should be none

### Resolved Issues

Fixes: #1184 

---

### DIFF
--- a/docs/conventions.md
+++ b/docs/conventions.md
@@ -11,8 +11,9 @@ We use [typescript-eslint](https://github.com/typescript-eslint/typescript-eslin
 
 ### Hints
 
-- Preferr arrow functions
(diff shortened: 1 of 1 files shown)


LeandraH (approved):
Great idea, thank you! Maybe we can also put this somewhere in the documentation?

Not tested

Issue #1227, opened 2025-12-03
Upgrade react native to 0.78


david-venhoff:
### Describe the Problem

This version brings some nice performance improvements

### Describe Possible Alternatives

None. We will have to do it eventually

### Related Issues

#1155

Issue #1201, opened 2025-11-10
Flaky test


david-venhoff:
### Describe the Bug

Found this in a failed CI run, see https://app.circleci.com/pipelines/github/digitalfabrik/lunes-app/4070/workflows/d18c74bb-90d0-42a2-8dd2-b62037ceb04b/jobs/12806:

<details>

```
FAIL src/routes/repetition/__tests__/RepetitionWordListScreen.spec.tsx
  ● RepetitionWordListScreen › should correctly remove words

    expect(received).toBeNull()

    Received: {"_fiber": {"_debugHookTypes": null, "_debugNeedsRemount": false, "_debugOwner": [FiberNode], "_debugSource": null, "actualDuration": 0, "actualStartTime": -1, "alternate": [FiberNode], "child": [FiberNode], "childLanes": 0, "deletions": null, "dependencies": null, "elementType": "Text", "flags": 0, "index": 0, "key": null, "lanes": 0, "memoizedProps": [Object], "memoizedState": null, "mode": 0, "pendingProps": [Object], "ref": null, "return": [FiberNode], "selfBaseDuration": 0, "sibling": null, "stateNode": [Object], "subtreeFlags": 0, "tag": 5, "treeBaseDuration": 0, "type": "Text", "updateQueue": null}}

      56 |     fireEvent.press(confirmButton)
      57 |
    > 58 |     await waitFor(() => expect(queryByText(wordNodeCards[0].word.word)).toBeNull())
         |                  ^
      59 |
      60 |     const newDeleteButtons = getAllByTestId('delete-button')
      61 |     expect(newDeleteButtons).toHaveLength(1)

      at Object.<anonymous> (src/routes/repetition/__tests__/RepetitionWordListScreen.spec.tsx:58:18)
```

</details>

### Steps to Reproduce

Could not reproduce it locally

### Expected Behavior

The test should always succeed

PR #1229, opened 2025-12-03
Update React native to 0.78

david-venhoff:
### Short Description

Since there are currently not many important issues, I decided to upgrade react native again, as this will need to be done anyways at some point.
The changelog is here: https://reactnative.dev/blog/2025/02/19/react-native-0.78
The new testing library version also includes a [new function](https://oss.callstack.com/react-native-testing-library/docs/api/misc/async#waitforelementtoberemoved) that might fix a flaky test, though I am not sure about it because I could never reproduce the test failing locally.
React 0.78 also brings in support for the [react compiler](https://react.dev/learn/react-compiler/introduction), which automatically memoizes components for improved performance 🎉 

### Proposed Changes

- Upgrade React native to 0.78
- Replace all usages of `JSX.Element` with `ReactElement`, because `JSX` was removed from the global namespace
- Try to fix the flaky test
- Enable the React compiler

### Side Effects

- The tests run now faster, for me about 3 seconds less. I think this caused by the tests now running in parallel (https://oss.callstack.com/react-native-testing-library/docs/migration/v13#concurrent-rendering-by-default).
- React 19 removed the global `JSX` type: https://react.dev/blog/2024/04/25/react-19-upgrade-guide#the-jsx-namespace-in-typescript
- The app should run overall a bit faster due to the react compiler, but I did not test this

### Resolved Issues

Fixes: #1227 
Fixes: #1201 

---

### DIFF
--- a/src/components/DeletionModal.tsx
+++ b/src/components/DeletionModal.tsx
@@ -1,4 +1,4 @@
-import React, { type JSX } from 'react'
+import React, { ReactElement } from 'react'
 import styled from 'styled-components/native'
 
 import { getLabels } from '../services/helpers'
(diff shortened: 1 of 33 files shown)


LeandraH (approved):
Nice! Tested on real iOS

PR #1263, opened 2026-01-28
Experiment/native api voice recognition

david-venhoff:
### Short Description

Experimental implementation for the voice recognition training mode.
It is only implemented on android right now.
In general it works quite well actually, but there are some words which just cannot be understood by androids voice recognition. To work around this, we can supply biasing strings which hint to the speech recognizer which word we expect to hear, but unfortunately this is only supported on modern androids (Tiramisu or higher).
Also some emulators seem to be broken, api 31 worked for me.

### Proposed Changes

- Add custom module for speech recognition (I tried to find a good react native library for this but failed. However, I think using a custom module is fine, since it can be a relatively shallow wrapper around the system apis)
- Implement voice training mode


david-venhoff:
Current status: I would consider this experiment more or less failed, because android and ios only provide an api to convert audio into a list of matches, while we actually need an api that determines how close the audio is to expected word. The problem is that many of the more specialized words used in lunes simply cannot be recognized by android, even with hints / biases.

Issue #1248, opened 2026-01-23
Update React native to 0.79


david-venhoff:
### Describe the Problem

We should update to a more recent react native version. 0.79.
 Should make the app start on android faster and improve dx

### Additional Information

https://reactnative.dev/blog/2025/04/08/react-native-0.79

Issue #1148, opened 2025-08-30
Upgrade react-navigation to 7.x


david-venhoff:
### Describe the Problem

We are currently using react navigation 6.x, which is no longer actively maintained

david-venhoff:
Related #1015

PR #1265, opened 2026-01-28
1248: Update RN to 79

LeandraH:
### Short Description

This PR updates a couple of packages, in particular `react-native` to 0.79 and `react-navigation` to version 7.

### Proposed Changes

- Updated `react-native` to 0.79
- Updated `@react-navigation` to versions 7
- Removed `react-native-tab-view` and its patch because it's not necessary anymore with `@react-navigation/material-top-tabs` version 7
- Replaced `AppDelegate.mm` with `AppDelegate.swift`, mostly to make future updates easier (the upgrade helper switch to Swift)
- Added some mocks because react-navigation didn't like the previous imports and removed some others

### Side Effects

- Should be none

### Testing

Go through the entire app, test in particular if the nested tabs under "Sammlung" work (I had some problems there). I sometimes got errors about missing routes but I assume that is the combination of the new API and the old app.

### Resolved Issues

Fixes: #1248 
Fixes: #1148 

---

### DIFF
--- a/src/constants/labels.json
+++ b/src/constants/labels.json
@@ -99,7 +99,8 @@
       "sentence": {
         "title": "Satzbausteine",
         "description": "Die Vokabel richtig anwenden"
-      }
+      },
+      "comingSoon": "Kommt bald"
     },
     "tryLater": "Später versuchen",
     "cheat": {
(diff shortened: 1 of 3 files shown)


david-venhoff:
Tested on android emulator, found two more place where navigation does not work yet:

image

- the "jetzt wiederholen" button in the image
- The back button In exercise for custom user vocabulary (and also "jetzt wiederholen")

Issue #1271, opened 2026-02-02
Development environment only: iOS app crashes when completing an exercise unit


LeandraH:
### Describe the Bug

This occurs on real and emulated iOS in dev mode, but not on Android or in prod mode.

When an exercise is completed (it seems to be specifically when calling `navigation.navigate('ExerciseFinished', { lots of params })`), Xcode pauses the app and gives me an error that says more or less that there is an empty string (in the C++ code). My first investigation suggests that the problem is the big object that we pass in though the Navigator, which react-navigation seems to serialize over the native bridge. iOS apparently has a limit on the size of objects that can be sent over the bridge and then triggers an abort statement (which is apparently the error I am seeing).

### Steps to Reproduce

1. Be on emulated or real iOS, with a development version of the app
2. Go to 'Module ansehen'
3. Choose a module where you have already looked at the Wortliste (or look at it first)
4. Solve all the exercises
5. See that the app becomes unresponsive and Xcode says it's paused

### Expected Behavior

The ExerciseFinished screen shows up

### Additional Information

<img width="880" height="115" alt="Image" src="https://github.com/user-attachments/assets/b32bdafe-94af-43c1-9dcf-f4cc67902ce2" />

Issue #1145, opened 2025-08-29
[T41K-12] Learning mode: Select correct image


david-venhoff:
### Describe the Problem

We want a new learning mode where users have to select the correct image given a word / audio.

PR #1273, opened 2026-02-03
1145 Implement image training exercise

david-venhoff:
### Short Description

This pr implements the image training mode.
image

### Proposed Changes

- Refactor the `ExerciseFinishedScreen`, so that it can be used from both training exercises and normal exercises
- Add the ImageTrainingScreen
- Add TrainingFinishedScreen which will also be used by the other training exercises
- Add code to use the words-of-job endpoint of the cms

### Side Effects

- The exercise finished screen now looks slightly different, for example I added some padding so that the progress bar is not right on the edge
- iOS should no longer crash when navigating to the exercise finished screen

### Testing

Test that the mode works and that it matches the [design](https://www.figma.com/design/4bW33nU6zZ6ySeTqrM37Jq/Lunes_Master?node-id=3269-1071&t=LKxBDngVyuhROtyK-1)

### Resolved Issues

Fixes: #1145
Fixes: #1271 
Fixes: #1277 

---

### DIFF
--- a/src/routes/training/__tests__/ImageTrainingScreen.spec.tsx
+++ b/src/routes/training/__tests__/ImageTrainingScreen.spec.tsx
@@ -35,6 +35,7 @@ describe('ImageTrainingScreen', () => {
         name: 'Test job',
         icon: 'icon',
         numberOfUnits: vocabularyItems.length,
+        migrated: false,
       },
     },
   }


src/routes/training/ImageTrainingScreen.tsx
```diff
@@ -0,0 +1,203 @@
...
+  vocabularyWithoutCurrentItem.splice(state.currentIndex, 1)
+
+  const choices = shuffleArray(vocabularyWithoutCurrentItem).slice(0, 3)
+  const insertIndex = Math.floor(Math.random() * (choices.length + 1))
+  choices.splice(insertIndex, 0, state.vocabularyItems[state.currentIndex])
+  return { ...state, choices: choices.map(choice => ({ src: choice.images[0], key: choice.id })) }
+}
+
+export const initializeState = (vocabularyItems: VocabularyItem[]): State => {
+  const shuffled = shuffleArray(vocabularyItems).splice(0, MAX_TRAINING_REPETITIONS)
```
LeandraH:
```suggestion
  const shuffled = shuffleArray(vocabularyItems).slice(0, MAX_TRAINING_REPETITIONS)
```
`splice` edits the original array (which is fine here but not what we want to do), slice does not.

src/routes/training/ImageTrainingScreen.tsx
LeandraH: Oooof, do we need the reducer? We've usually been using a couple of `useState` variables for complex routing like this 😅

src/routes/training/ImageTrainingScreen.tsx
```diff
@@ -0,0 +1,203 @@
...
+import { Route, RoutesParams } from '../../navigation/NavigationTypes'
+import { getLabels, shuffleArray } from '../../services/helpers'
+import { reportError } from '../../services/sentry'
+import ImageGrid, { ImageGridItem, ImageGridItemState } from './components/ImageGrid'
+import TrainingExerciseContainer from './components/TrainingExerciseContainer'
+import TrainingExerciseHeader from './components/TrainingExerciseHeader'
+
+type State = {
+  vocabularyItems: VocabularyItem[]
+  currentIndex: number
```
LeandraH: I don't quite understand, is this the index of the correct image?

src/routes/training/ImageTrainingScreen.tsx
```diff
@@ -0,0 +1,203 @@
...
+  completed: boolean
+}
+
+const initializeChoices = (state: Omit<State, 'choices'>): State => {
+  if (state.vocabularyItems.length === 0) {
+    return { ...state, choices: [] }
+  }
+
+  const vocabularyWithoutCurrentItem = [...state.vocabularyItems]
+  vocabularyWithoutCurrentItem.splice(state.currentIndex, 1)
```
LeandraH: I find a filter more legible, something like
```
const vocabularyWithoutCurrentItem = [...state.vocabularyItems].filter((_, index) => index !== state.currentIndex)
```

src/constants/labels.json
```diff
@@ -88,9 +88,16 @@
     },
     "training": {
       "train": "Vokabeln trainieren",
+      "finished": {
+        "askContinue": "Möchtest du eine weitere {} Übung machen?",
```
LeandraH: I know that's what the design says but I believe grammatically correct would be {}-Übung.

LeandraH (approved):
Wow, this looks great, I'm really impressed! I find it a very hard to understand some of the code (probably largely because it's a complex feature) so I tried to find things to make it clearer to me 😅

src/routes/training/ImageTrainingScreen.tsx
```diff
@@ -0,0 +1,204 @@
...
+  switch (action.type) {
+    case 'selectAnswer': {
+      if (state.answer?.isCorrect) {
+        return state
+      }
+      const isCorrect = action.key === state.vocabularyItems[state.currentVocabularyItemIndex].id
+      return {
+        ...state,
+        correctAnswersCount:
+          isCorrect && state.answer === null ? state.correctAnswersCount + 1 : state.correctAnswersCount,
```
LeandraH: Maybe we can still add a `isFirstAttempt = state.answer === null` here

src/routes/training/ImageTrainingScreen.tsx
```diff
@@ -0,0 +1,204 @@
...
+  if (state.vocabularyItems.length === 0) {
+    return { ...state, choices: [] }
+  }
+
+  const vocabularyWithoutCurrentItem = state.vocabularyItems.filter(
+    (_, index) => index !== state.currentVocabularyItemIndex,
+  )
+
+  const choices = shuffleArray(vocabularyWithoutCurrentItem).slice(0, 3)
+  const insertIndex = Math.floor(Math.random() * (choices.length + 1))
```
LeandraH: Maybe `indexOfCorrectAnswer` or something like that to make it clearer that this isn't the function that does the inserting?

Issue #1071, opened 2025-03-20
[T41K-8] Add info on module-overview screen about progress might not be shown


nikolahoff:
### Describe the Problem

There were problems about wrong vocabularylists in some modulegroups. So therefore we need to restructure the CMS. After reorganizing the process of some modules might not be available anymore.

### Describe Your Preferred Solution

We need an information displayed to the user on the module overview page.

### Describe Possible Alternatives

---

### Related Issues

---

### Additional Information

--

nikolahoff:
:round_pushpin:

nikolahoff:
@david-venhoff the design for an information about missing progress data can be found [here](https://www.figma.com/design/4bW33nU6zZ6ySeTqrM37Jq/Lunes_Master?node-id=1980-213&t=dWvO0kqrefAX7q8m-1)

HuguetteAfi:
@david-venhoff I would only display the warning in the specific jobs targeted then and not in general.

PR #1274, opened 2026-02-04
1071: Note about Progress Reset

LeandraH:
### Short Description

This PR adds a note to the `UnitSelectionScreen` that progress on this job may have been lost if the job has been migrated since the user selected it.

### Proposed Changes

- Add a note about possibly lost progress with a modal explaining further with the possibility to contact us (I used the existing feedback modal for that)
- Add a list to the `Storage` of jobs that are selected and not migrated
- Check in the `UnitSelectionScreen` if the current job is on that list and is migrated; if so, show the note
- Renamed and added a color for better consistency

### Side Effects

- None

### Testing

Manually add an ID to the `notMigratedSelectedJobs` (I suggest 28, from the Anlagenmechaniker:in für Sanitär-, Heizungs- und Klimatechnik, it's currently migrated and the first result in the job list)

### Resolved Issues

Fixes: #1071 

---

### DIFF
--- a/src/routes/UnitSelectionScreen.tsx
+++ b/src/routes/UnitSelectionScreen.tsx
@@ -10,7 +10,7 @@ import RouteWrapper from '../components/RouteWrapper'
 import ServerResponseHandler from '../components/ServerResponseHandler'
 import Title from '../components/Title'
 import useLoadUnits from '../hooks/useLoadUnits'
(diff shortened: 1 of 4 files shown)


david-venhoff (approved):
Thanks!
Looks very good to me, I only found a small bug in the migration

Issue #1146, opened 2025-08-29
Learning mode: complete a sentence


david-venhoff:
### Describe Your Preferred Solution

The user will be given a list of words and arrange them so they will complete a correct sentence.

PR #1275, opened 2026-02-09
1146 sentence training

david-venhoff:
### Short Description

Implement sentence training. Based off 1145-learning-mode-image-selection because it re-uses some code added by this pr.
image

### Proposed Changes

- Update CmsAPI to load the example sentences
- Update the VocabularyItem type to store them
- Add a BottomSheet component
- Implement the training screen

### Side Effects

- Should be none

### Testing

Test that the mode works, all the new jobs in the test cms already contain example sentences

### Resolved Issues

Fixes: #1146 

---

### DIFF
--- a/src/components/BottomSheet.tsx
+++ b/src/components/BottomSheet.tsx
@@ -82,7 +82,8 @@ const BottomSheet = ({ visible, ...props }: BottomSheetProps): ReactElement => {
         <ModalBody
           backgroundColor={backgroundColor ?? theme.colors.backgroundHigh}
           bottomPadding={bottomPadding}
-          style={{ transform: [{ translateY: offset }, { perspective: 1000 }] }}>
+          style={{ transform: [{ translateY: offset }, { perspective: 1000 }] }}
+        >
           {children}
         </ModalBody>
       </ModalContainer>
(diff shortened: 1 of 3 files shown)


david-venhoff:
The pr currently has one bug that the audio player does not work, because the urls returned by the cms are relative. But this is probably a CMS issue

src/routes/training/sentence/State.ts
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Converts a sentence into an array of words for the exercise
+ * Punctuation like dots and exclamation marks are removed from the words
+ * so that the exercise does not become too easy.
+ * Also makes sure that double spaces do not get converted into an empty word,
+ * because that has already happened in the CMS data.
+ */
+export const splitSentence = (sentence: string): string[] =>
+  sentence
+    .replace(/[.!]/g, '')
```
LeandraH: This feels a bit fragile to me, but I guess punctuation in German is not going to change fundamentally anytime soon 😂 Still, I'd like to add `,?;`. Maybe even an m dash if the sentences are created by AI 😁

david-venhoff: Although, alternatively we could split "Purpur-Sonnehut" into two words? 🤔 

Not sure whether it is better to keep it together or split it

Issue #1278, opened 2026-02-10
Make 'Sprechstudio' button greyed out


david-venhoff:
### Describe the Problem

Since we won't finish this training mode before the next exercise, we should clearly indicate that the button is not functional yet.

### Additional Information

Context: https://chat.tuerantuer.org/digitalfabrik/pl/rg9fcaoo8bgkmni8xhue71xbsa

PR #1279, opened 2026-02-10
1278 Indicate when a training mode is not yet ready

david-venhoff:
### Short Description

This is mainly for the 'Sprechstudio' mode, which will likely not yet be finished when we do the next release.
image

### Proposed Changes

- Add some opacity and a label "coming soon" to not yet implemented training modes

### Side Effects

- should be none

### Testing

Should display correctly

### Resolved Issues

Fixes: #1278 

---


JoeyStk (approved):
Thank you very much for your PR! 🎉
This looks very good, tested on Android 12.

I have two nits you could think about, but since they are really minor, I decided to approve already and you can decide for yourself, if you want to further look into them :)

Issue #1214, opened 2025-11-21
Remove levels 3 & 4


david-venhoff:
### Describe the Problem

For didactic reason we don't want to support the levels 3 and 4 anymore. 

### Describe Your Preferred Solution

Remove these two levels. 

### Describe Possible Alternatives

Do nothing

PR #1284, opened 2026-02-16
1214: Delete levels 3&4

JoeyStk:
### Short Description

This PR deletes the article exercise and writing exercise. Because the two classes have been referenced by other classes, I also had to delete those dependencies.

### Proposed Changes

- Delete writing exercise & dependencies
- Delete article exercise & dependencies
- Refactored a little bit 
- Adjust a few small things to match the new design

### Side Effects

- Has been tested on Android (v12) and iOS emulator
- There are some possible side effects when levels 3 &4 have been worked on in the past and have now been deleted. For me it caused an ExceptionError and the app wouldn't open anymore. @david-venhoff and I came to the conclusion however, that fixes this wouldn't be part of this starter ticket, that's why I left out for now 

### Resolved Issues

Fixes: #1214 
Fixes: #1300

---


JoeyStk:
@david-venhoff We tried to work in all of the feedback in the old PR. Do you think you could have a look at this PR? :) That would be great :) Thank you :)

src/routes/exercise-finished/ExerciseFinishedScreen.tsx
```diff
@@ -195,12 +162,7 @@ const ExerciseFinishedScreen = ({ navigation, route }: ExerciseFinishedScreenPro
           </MessageContainer>
         </RoundedBackground>
 
-        <Button
-          label={buttonText}
-          iconLeft={buttonText === getLabels().results.action.repeat ? RepeatIcon : undefined}
```
david-venhoff: Was the icon removed on purpose?

src/routes/write-exercise/services/RepetitionWriteExerciseService.ts
david-venhoff: With the current changes, the repetition exercise does not update its word node cards any more, because the updating happened here

JoeyStk:
Thank you for your review! 

> Oh, and also add a changelog note please :)

Do you mean as a release note? If so, should be added now :)

> Can you still reproduce this? I couldn't and I am not sure what would cause this.

No, seems to be a leftover from the last PR, also couldn't reproduce this anymore :)

LeandraH:
Sorry, my bad, didn't think about my own PR when we were dealing with the merge conflicts here yesterday 😓

src/routes/choice-exercises/components/SingleChoiceExercise.tsx
```diff
@@ -62,12 +64,13 @@ const ChoiceExerciseScreen = ({
   )
   const { vocabularyItem, numberOfTries, result } = results[currentWord]
   const [answers, setAnswers] = useState<Answer[]>(vocabularyItemToAnswer(vocabularyItem))
+  const repetitionService = RepetitionService.fromStorageCache(storageCache)
```
david-venhoff:
```suggestion
  const repetitionService = useRepetitionService()
```

david-venhoff (approved):
Great, thanks!

I found some small bugs, but nothing major

david-venhoff:
I think we can get rid of some dependencies :tada:

- react-native-popover-view
- react-native-keyboard-aware-scroll-view

Issue #1294, opened 2026-02-18
Sidebar does not close after navigating back


david-venhoff:
### Describe the Bug

When going to the settings screen and then navigating back, the sidebar is still open.
Also on android I get weird animation where the settings menu slides up from the bottom of the screen, instead of fading in, as it does in the latest public release. 

### Steps to Reproduce

1. Go to Settings
2. Navigate back
4. Notice that the sidebar is still open

### Expected Behavior

The sidebar should be closed

### Related Issues

Looks like #1046 has re-appeared.

### Additional Information

Might be caused by the upgrade of react navigation.

PR #1295, opened 2026-02-18
Fix sidebar navigation

david-venhoff:
### Short Description

I noticed a small navigation issue that we missed in the migration to react native navigation 7, where the sidebar does not close after clicking any button inside it.

### Proposed Changes

- use `popTo` instead of `navigate`, as suggested here: https://reactnavigation.org/docs/upgrading-from-6.x/#the-navigate-method-no-longer-goes-back-use-popto-instead

### Side Effects

- Should be none

### Testing

The sidebar should close again after clicking a button inside it, e.g. `Settings` and it should be closed when pressing the back button in the settings menu.

### Resolved Issues

Fixes: #1294 

---


LeandraH (approved):
Nicely done, tested on emulated iOS

Issue #1281, opened 2026-02-11
Update popup notification text leading to the "Vertiefungsübung" exercise


HuguetteAfi:
### Describe the Problem

The text leading to the "Vertiefungsübung" exercise does not reflect its purpose and should be updated to match the content of the updated tooltip (see image).

### Describe Your Preferred Solution

New text example: „Einige Wörter waren schwer, keine Sorge!" (Titel), „Unter „Vertiefen" kannst du diese Vokabeln gezielt wiederholen. “ (Text)

Also, it should only pop up if there were incorrect answers.

### Additional Information

<img width="590" height="1278" alt="Image" src="https://github.com/user-attachments/assets/04df09cb-40b3-4e80-a0b0-e3b48318672a" />

Issue #1280, opened 2026-02-11
Update Info tooltip text for "Vertiefungsübung" exercise


HuguetteAfi:
### Describe the Problem

The info-tooltip should be added and updated to provide clearer guidance on the exercise, for now user don't get a clear prewiew to the structure of the activity. 

### Describe Your Preferred Solution

- Add an Info-Tooltip next to „Wörter wiederholen" stating „Wiederhole hier Vokabeln, die dir beim Kennenlernen und Üben bisher schwer gefallen sind."
- Update Info-Tooltip next to „Dein Lernfortschritt" stating „Jede Vokabel durchläuft 7 Phasen. Zu Beginn wiederholst du sie häufiger, später in größeren Abständen."

### Additional Information

- Only text update; no UX or design changes required since details were already discussed with @hauf-toni 
- Functionality of the popup remains unchanged.

<img width="590" height="1278" alt="Image" src="https://github.com/user-attachments/assets/17e3a911-5f1d-484a-bf48-eaaacf36f36d" />

<img width="375" height="753" alt="Image" src="https://github.com/user-attachments/assets/edea357a-84ed-4fc8-b30e-c9d14214b137" />

Issue #1276, opened 2026-02-10
Adjust "Vertiefungsübung": focus on incorrect vocabulary + multiple-choice format


HuguetteAfi:
### Describe the Problem

Due to the release of two new features, the didactic context of the "Vertiefungsübung" has changed.
In its current form (free text input for all vocabulary items across multiple intervals), the value of the exercise is no longer fully clear.
Instead of removing the exercise completely, we want to introduce a short-term, pragmatic solution that keeps the feature useful while a long-term concept is being worked on in parallel.

### Describe Your Preferred Solution

**Goal**
- Preserve the repetition aspect of the exercise
- Reduce cognitive load for users
- Ensure the exercise complements the new features rather than competing with them

### Proposed Change (Short-Term Solution)
- Only incorrectly answered vocabulary items are included in the deepening exercise
- The repetition format is changed to multiple-choice - An image is displayed, four answer options are shown, one option is correct (Level 1)
- The exercise is still presented in multiple repetition intervals, aligned with the existing spacing logic

PR #1297, opened 2026-02-19
1276: Only add incorrect words to long-term-memory exercises

LeandraH:
### Short Description

This PR changes when words are added to the long-term-memory exercises: only when they've been answered incorrectly at least once.

The ticket also requests that the repetition exercise be the `WordChoiceExercise` but @JoeyStk and I are implementing that in #1284 

### Proposed Changes

- If a word exercise is answered incorrectly at least once, that word is added to the repetition service
- Not a change: if a word is favourited, it will still be added to the repetition service, waiting for feedback from Huguette here
- If an incorrect word is already in the repetition service, it has so far been ignored. Now, it will be reset to the first section
- The modal telling users about the repetition exercises at the end of an exercise will only be shown when a word has been answered incorrectly
- The wording on that modal has been changed
- The wording on the modal on the repetition screen has been changed, and a new one added
- The styling of the repetition screen has been slightly refactored

### Side Effects

- The lower headline on the repetition screen ("Dein Lernfortschritt") is now centered. I don't know why it wasn't before.

### Testing

Do a couple of exercises, check that words that are answered correctly are not added to the repetition service, but words that have been answered incorrectly at least once have been. Check the modals.

### Resolved Issues

Fixes: #1276 
Fixes: #1280 
Fixes: #1281 

---


david-venhoff (approved):
Awesome, Thanks!

The only question I have is whether we should also now reset the progress of a word node card if it has been answered wrong in the repetition exercise, for consistency?

Issue #1190, opened 2025-10-27
[T41K-6] Refactor Exercises


david-venhoff:
### Describe the Problem

Currently the code for exercise is a bit hard to read. 

### Describe Your Preferred Solution

We should probably simplify the code for exercises a bit, before adding new exercises

### Describe Possible Alternatives

Do nothing

PR #1305, opened 2026-02-27
Refactor exercises

david-venhoff:
### Short Description

Some cleanup to the exercises. I tried to avoid changing the logic of `WordChoiceExercise` as much as possible (to not accidentally introduce new bugs), while also getting rid of the `useEffect` calls.

### Proposed Changes

- Remove some unused code and props
- Rename `SingleChoiceExercise` to `WordChoiceExercise` for consistency
- Combine the two tests for word choice exercises. I think it was not on purpose that we had two test files for the same screen
- Remove some `useEffect` calls which should better be avoided with react
- Simplify navigation params
- Simplify navigation back from exercises: We now simply call `navigation.pop()`, instead of passing around a navigation action

### Side Effects

- The behavior when going back in exercises is now slightly different and more consistent than before. (Before this pr, when navigating back from an exercise, you would sometimes end up in the last screen, sometimes at the home screen, etc. Now it is always the last screen)

### Testing

Test that the exercises still work and that the code is more understandable to you than before (And feel free to suggest additional refactorings, if you find any)

### Resolved Issues

Fixes: #1190

---


src/routes/choice-exercises/components/WordChoiceExercise.tsx
```diff
@@ -30,6 +29,45 @@ import { getLabels, moveToEnd, shuffleArray } from '../../../services/helpers'
...
+const generateFalseAnswers = (
+  vocabularyItems: VocabularyItem[],
+  correctVocabularyItem: VocabularyItem,
+  answersCount: number,
+): Answer[] => {
+  const shuffledWrongAnswers = shuffleArray(vocabularyItems.filter(it => it.id !== correctVocabularyItem.id))
+  return shuffledWrongAnswers.slice(0, answersCount - 1)
+}
+
+const vocabularyItemToAnswer = (vocabularyItems: VocabularyItem[], vocabularyItem: VocabularyItem): Answer[] => {
```
LeandraH: 💭 Could we maybe rename this to something like `getChoices` or `getAnswerChoices`? Or `generateChoices`?

src/routes/choice-exercises/components/WordChoiceExercise.tsx
```diff
@@ -151,7 +170,7 @@ const ChoiceExerciseScreen = ({
     const isCorrect = correctAnswers.some(it => isAnswerEqual(it, clickedAnswer))
     await updateResult(numberOfTries + 1, isCorrect)
 
-    if (!isRepetitionExercise && exerciseKey >= FIRST_EXERCISE_FOR_REPETITION && !isCorrect) {
+    if (!isRepetitionExercise && !isCorrect) {
       const repetitionService = RepetitionService.fromStorageCache(storageCache)
```
LeandraH: Same here

src/routes/write-exercise/components/MissingArticlePopover.tsx
LeandraH: Uhm, this is gone, right?

src/routes/exercise-finished/ExerciseFinishedScreen.tsx
```diff
@@ -107,7 +107,7 @@ const ExerciseFinishedScreen = ({ navigation, route }: ExerciseFinishedScreenPro
         feedbackColor={resultColor}
         FeedbackIcon={ResultIcon}
         message={message}
-        onBack={() => navigation.dispatch(closeExerciseAction)}
+        onBack={() => navigation.pop()}
       >
         <Button label={buttonText} buttonTheme={BUTTONS_THEME.contained} onPress={() => navigationAction()} />
```
LeandraH:
```suggestion
        <Button label={buttonText} buttonTheme={BUTTONS_THEME.contained} onPress={navigationAction} />
```

LeandraH (approved):
Nice job, it's a great idea to try to keep things more maintainable! Tested on emulated iOS

Issue #1286, opened 2026-02-16
Revisit the design of the exercises


JoeyStk:
### Describe the Problem

In the latest design there have been some small changes to the existing design.

### Describe Your Preferred Solution
- [x] Change view of module list (s. screenshots 1a, 1b)
- [x] Change feedback in module list (s. screenshot 2a, 2b, 2c)
- [x] Change size of Lunes icon on home screen (s. screenshot 3)
- [x] Change feedback screen (increase font size, change position of circle) (s. screenshot 4)

### Screenshot 1:
<img width="307" height="575" alt="Image" src="https://github.com/user-attachments/assets/21f6e3a8-fa55-4066-bcb9-81f2e9c57200" />
should look like this:

<img width="321" height="575" alt="Image" src="https://github.com/user-attachments/assets/c76e33c6-d001-46a1-8611-f67d9eae686e" />

### Screenshot 2: 

<img width="300" alt="height" src="https://github.com/user-attachments/assets/da40724d-95d1-4bfb-91fe-c981589241cf" />

should look like this: 

<img width="337" height="581" alt="Image" src="https://github.com/user-attachments/assets/9b5f2091-c472-48c1-8731-3dc97c4fa1aa" /> and

<img width="337" height="581" alt="Image" src="https://github.com/user-attachments/assets/106e9eaa-fd15-4c74-8bc0-0f324114e306" />

### Screenshot 3:
<img width="300" alt="height" src="https://github.com/user-attachments/assets/f5bd1b26-979f-4895-9915-09707633842f" />
should look like this:

<img width="321" height="567" alt="Image" src="https://github.com/user-attachments/assets/2ccc1e1d-a590-4a16-a2aa-8a21c28ae2c0" />

### Screenshot 4: 
<img width="300" alt="height" src="https://github.com/user-attachments/assets/dfb7326c-d3ca-4b67-b446-52b0728b18ec" />
should look like this:

<img width="320" height="613" alt="Image" src="https://github.com/user-attachments/assets/b7e833b4-b2ab-4134-90fe-0c1f7fca418b" />

PR #1318, opened 2026-03-09
1286: Finish redesign

JoeyStk:
### Short Description

In this PR I revisited the new design template and changed the last remaining differences between design and implementation.

### Proposed Changes

- Increase size of the logo
- Change the look of the exercise selection
- Change feedback badge

### Side Effects

- I hope none

### Resolved Issues

Fixes: #1286 

---


src/routes/home/components/HomeScreenHeader.tsx
```diff
@@ -18,7 +18,7 @@ const HeaderStyle = styled.View`
 const SmileIconStyle = styled.Pressable`
   position: absolute;
   left: 50%;
-  margin-left: -${hp('3%')}px;
+  margin-left: -${hp('5%')}px;
   top: 50%;
```
david-venhoff: I think the item is centered vertically in the design. One (although ugly) way to do that would be to replace this line with `bottom: -${hp('6%')}px;` (half of the icon size)

JoeyStk:
Thank you for your reviews! I hope that I could in-cooperate all of your feedback. However, during my conclusive testing round, I noticed that this PR seems to reintroduce [this bug](https://github.com/digitalfabrik/lunes-app/pull/1319), (see screenshot below)
WhatsApp Image 2026-03-14 at 11 45 36

I'll investigate this further in a bit

PR #1327, opened 2026-03-16
Fix github release creation

david-venhoff:
### Short Description

The command to create a new github release was not executed, because it was wrapped in single quotes (''), which prevent all string interpolation in bash. The fix to just remove them also matches what the integreat app does: https://github.com/digitalfabrik/integreat-app/blob/ba803418a36c77892336abbda5d8bc337c3db05f/.circleci/src/jobs/notify_release.yml#L21

### Proposed Changes

- Remove the single quotes

### Side Effects

- should be none

### Testing

lets test it in the next beta release 😁 

### Resolved Issues

Fixes: /

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -523,7 +523,7 @@ jobs:
             - run:
                 command: |
                     echo "Creating release ${NEW_VERSION_NAME}..."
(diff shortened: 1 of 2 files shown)


JoeyStk (approved):
Wonderful! Thank you very much! This should fix it :)

Issue #1262, opened 2026-01-27
[DT-APP-Y-1] GDPR User Controls


sascha11110:
**Scope:** Implement data export and deletion UI in Settings.

**Tasks:**
1. Add "Export My Data" button in Settings
   - Calls CMS export endpoint
   - Shows JSON or offers download
2. Add "Delete My Data" button in Settings
   - Confirmation dialog
   - Calls CMS deletion endpoint
   - Resets local tracking consent
3. Show installation ID in Settings (for support purposes)

**References:**
- Settings screen: `lunes-app/src/routes/settings/SettingsScreen.tsx`

PR #1375, opened 2026-03-31
Add GDPR user controls

david-venhoff:
### Short Description

Since we are already releasing tracking functionality in the next release, it probably makes sense to also include the GDPR controls. 

### Proposed Changes

- Add settings for exporting and deleting analytics data
- Add a new screen for viewing and copying the exported data
- Add some tests

### Side Effects

 - Should be none

### Testing

Can be tested with a local cms server, or on the test server once it has updated to include the changes in develop

### Resolved Issues

Fixes: #1262 

---


src/routes/settings/SettingsScreen.tsx
```diff
@@ -48,29 +70,64 @@ const SettingsScreen = (): ReactElement => {
     await setAnalyticsConsent({ consentGiven: newConsentGiven, consentDate: new Date().toISOString() })
   }
 
+  const onDeleteAnalyticsData = async (): Promise<void> => {
+    try {
+      await deleteAnalyticsData(installationId ?? '')
```
LeandraH: Does that mean that if we don't have an `installationId`, we call 'analytics/data//'? Though I guess there shouldn't really be a case of not having an `installationId`?

Same for getting the data in the export screen, I guess?

PR #1381, opened 2026-04-07
Move release notes back to unreleased

david-venhoff:
### Short Description

If I understood correctly, we will make another beta release, which will again require the release notes to be in unreleased.

### Proposed Changes

- Move release notes back to unreleased

### Side Effects

None

### Resolved Issues

Fixes: /

---


LeandraH (approved):
Ah yes. Ooops 😅

PR #1395, opened 2026-04-08
1386: Remove react-native-responsive-screen

### LINKED ISSUE #1386: Remove react-native-responsive-screen
### Describe the Problem

We currently use [`react-native-responsive-screen`](https://github.com/marudy/react-native-responsive-screen) for quite a bit of sizing which is not ideal when the screen size changes (e.g. because of someone turning the phone) because it doesn't recalculate. We can probably replace its uses mostly with `useWindowDimensions` (which is recalculated), or other built-in functions.

I would like to not have `react-native-responsive-screen` in our code base any more, and all of its uses replaced by native methods.

We could leave it until something really breaks. Let's not.

LeandraH:
### Short Description

This PR removes [`react-native-responsive-screen`](https://github.com/marudy/react-native-responsive-screen) which is not actively maintained, and not the best use for our code anyway.

### Proposed Changes

- Remove from the constants and replace with the same values for a screen height of 800px
- Add a constant for 28px which seems to be a common icon size in our code, adapted three icons to that size
- Replace other occurrences with functionally similar code
- Refactoring

### Side Effects

- When tapping outside the modal, it now closes. That's not a side effect, but it doesn't strictly speaking belong here, it just bothered me.
- There are a lot of somewhat random numbers in the code now. I tried to clean it up a bit already, but there are still more.

### Testing

Go through the entire app and see if sizes look weird. Sorry!

### Resolved Issues

Fixes: #1386 

---


src/components/ModalSkeleton.tsx
```diff
@@ -44,25 +38,34 @@ type ModalSkeletonProps = {
 
 const ModalSkeleton = ({ visible, onClose, testID, children }: ModalSkeletonProps): ReactElement => {
   const { keyboardHeight, isKeyboardVisible } = useKeyboard()
-  const screenHeight = useScreenHeight()
   const theme = useTheme()
-  const onCloseKeyboard = () => isKeyboardVisible && Keyboard.dismiss()
+  const handleBackdropPress = (): void => {
```
LeandraH, line 44: I found it annoying, and unusual behavior that the modal didn't close when I tapped outside, so I changed that.

src/routes/home/components/NextExerciseCard.tsx
LeandraH, line 1: Turns out we don't use that anymore.

src/components/Button.tsx
```diff
@@ -29,14 +29,9 @@ const ThemedButton = styled.Pressable<ThemedButtonProps>`
       border-width: 1px;
     `};
   flex-direction: row;
-  padding: ${props => `${hp('1.5%')}px ${props.theme.spacings.sm}`};
-  ${props =>
-    !props.fitToContentWidth &&
-    css`
-      width: ${wp('70%')}px;
-    `};
+  padding: 12px ${props => props.theme.spacings.sm};
```
david-venhoff, line 33:
```suggestion
  padding: ${props => props.theme.spacings.sm};
```

💭 Maybe we can remove the hardcoded 12px and use the same value for both axis?

src/components/Button.tsx
```diff
@@ -81,6 +77,7 @@ const Button = (props: ButtonProps): ReactElement => {
     fitToContentWidth,
     testID = 'button',
   } = props
+  const buttonWidth = fitToContentWidth ? undefined : screenWidth * BUTTON_WIDTH_RATIO
```
david-venhoff, line 86: 💭 Maybe we could try to refactor the button a bit in the future to remove the `fitToContentWidth` parameter and make it the default behavior. But that seems way out of scope for this pr

src/components/Button.tsx
```diff
@@ -111,8 +108,7 @@ const Button = (props: ButtonProps): ReactElement => {
       backgroundColor={getBackgroundColor()}
       onPress={onPress}
       disabled={disabled}
-      style={style}
-      fitToContentWidth={!!fitToContentWidth}
+      style={[{ width: buttonWidth }, style]}
```
david-venhoff: 🙃 For consistency maybe the width should go into the styled button

src/components/DisciplineListItem.tsx
```diff
@@ -27,21 +26,21 @@ type JobListItemProps = {
...
+  width: ${props => props.theme.sizes.defaultIcon}px;
+  height: ${props => props.theme.sizes.defaultIcon}px;
 `
 
 const IconContainer = styled.View`
   position: absolute;
-  top: ${hp('1.75%')}px;
-  left: ${hp('1.75%')}px;
+  top: 14px;
+  left: 14px;
```
david-venhoff:
```suggestion
  align-self: center;
```

src/components/FeedbackBadge.tsx
```diff
@@ -26,8 +25,8 @@ const BadgeText = styled(HintSecondary)`
 `
 
 const BadgeIcon = styled.View`
-  height: ${hp('3.5%')}px;
-  max-height: ${hp('2.75%')}px;
+  height: ${props => props.theme.sizes.defaultIcon}px;
+  max-height: 22px;
```
david-venhoff:
```suggestion
```
💭 Since this effectively just reduces the vertical padding, maybe we can remove this hardcoded value and then set the vertical padding in BadgeContainer to 0 (currently xxs)

src/components/Header.tsx
```diff
@@ -9,43 +8,43 @@ const Wrapper = styled.View`
...
+  width: 24px;
+  height: ${props => props.theme.sizes.defaultIcon}px;
 `
 
 const SmileIconStyle = styled.Pressable`
   position: absolute;
   left: 50%;
-  margin-left: -${hp('3%')}px;
-  top: ${hp('7.5%')}px;
+  margin-left: -24px;
```
david-venhoff:
```suggestion
  align-self: center;
```
Since the width is 64px, the correct offset would be -32px to center it. But we can also just reuse the align-self trick :) In that case we also need to remove the `left: 50%` part I think

src/components/ModalSkeleton.tsx
```diff
@@ -44,25 +38,34 @@ type ModalSkeletonProps = {
...
+      <StyledPressable onPress={handleBackdropPress}>
         <Overlay>
-          <ModalContainer
-            bottomPosition={isKeyboardIosVisible ? keyboardHeight : undefined}
-            height={isKeyboardVisible ? screenHeight - keyboardHeight : undefined}
-          >
-            <Icon onPress={onClose}>
-              <CloseIcon width={theme.spacingsPlain.md} height={theme.spacingsPlain.md} />
-            </Icon>
-            <ScrollView persistentScrollbar contentContainerStyle={{ alignItems: 'center' }}>
```
david-venhoff, line 63: Removing the `alignItems: 'center'` here causes the feedback boxes in the feedback modal to become uncentered. I think we could either add this back or modify the feedback modal to increase the width of the feedback text inputs.
Edit: looks like this also causes the buttons in many other modals to become uncentered, including the feedback modal, the job remove modal and the gdpr data deletion modals.

src/components/Title.tsx
```diff
@@ -9,7 +8,7 @@ import { HeadingText } from './text/Heading'
 import { SubheadingText } from './text/Subheading'
 
 const Container = styled.View`
-  min-height: ${hp('7%')}px;
+  min-height: 56px;
```
david-venhoff:
```suggestion
```
I think we can remove this, I couldn't find a screen where this had an effect

src/components/VocabularyItemImageSection.tsx
```diff
@@ -10,7 +9,7 @@ import ImageCarousel from './ImageCarousel'
 
 const AudioContainer = styled.View`
   position: absolute;
-  bottom: ${hp('-2.25%')}px;
+  bottom: -18px;
```
david-venhoff:
```suggestion
  bottom: -16px;
```
I think 16px should be the exact value needed for centering this, the audio player is 32px

src/components/VocabularyListItem.tsx
```diff
@@ -11,18 +10,18 @@ import { ContentTextLight } from './text/Content'
...
 const StyledTitle = styled(ContentTextLight)<{ articleColor: string }>`
   border-radius: ${props => props.theme.spacings.xs};
   margin-bottom: ${props => props.theme.spacings.xxs};
   background-color: ${props => props.articleColor};
   align-self: flex-start;
-  width: ${hp('5%')}px;
+  width: 40px;
   overflow: hidden;
-  height: ${hp('2.5%')}px;
+  height: 20px;
```
david-venhoff:
```suggestion
 overflow: hidden;
 padding: 0 ${props => props.theme.spacings.xs}
```
I think we can replace the hardcoded size with some padding

src/constants/theme/sizes.ts
```diff
@@ -1,7 +1,6 @@
-import { widthPercentageToDP as wp } from 'react-native-responsive-screen'
-
 const SIZES = {
-  smallImage: `${wp('32%')}px`,
+  smallImage: '32%',
```
david-venhoff: 💭 What do you think about choosing a fixed size here? Looks like '128px' might be a decent value.
IMHO it would make sense to stay consistent with the changes to the spacings, but I am open to other opinions :)

src/constants/theme/sizes.ts
```diff
@@ -1,7 +1,6 @@
...
+  defaultIcon: 28,
```
david-venhoff, line 3: 🙃 Maybe this should go into a new `sizesPlain` const, so that we can use '28px' here? That would be more consistent with spacings.ts.

src/navigation/BottomTabNavigator.tsx
```diff
@@ -46,27 +46,31 @@ const BottomTabNavigator = (): ReactElement | null => {
   const numberOfWordsNeedingRepetition = repetitionService.getNumberOfWordsNeedingRepetition()
 
   const renderHomeTabIcon = ({ focused }: { focused: boolean }) =>
-    focused ? <HomeIconWhite width={hp('5%')} height={hp('5%')} /> : <HomeIconGrey width={hp('5%')} height={hp('5%')} />
+    focused ? (
+      <HomeIconWhite width={HOME_TAB_ICON_SIZE} height={HOME_TAB_ICON_SIZE} />
+    ) : (
+      <HomeIconGrey width={HOME_TAB_ICON_SIZE} height={HOME_TAB_ICON_SIZE} />
+    )
```
david-venhoff: 🙃 We could update the home icons viewBox to "6 6 19 19" to remove the padding around the svg, so that this special case of width and height won't be needed

src/navigation/BottomTabNavigator.tsx
```diff
@@ -78,11 +82,11 @@ const BottomTabNavigator = (): ReactElement | null => {
...
           tabBarStyle: {
             backgroundColor: theme.colors.primary,
-            minHeight: hp('7%') + insets.bottom,
+            minHeight: TAB_BAR_HEIGHT + insets.bottom,
             paddingBottom: insets.bottom,
           },
-          tabBarItemStyle: { height: hp('7%'), padding: theme.spacingsPlain.xs },
-          tabBarLabelStyle: { fontSize: hp('1.5%'), paddingLeft: isTablet() ? hp('1%') : 0 },
+          tabBarItemStyle: { height: TAB_BAR_HEIGHT, padding: theme.spacingsPlain.xs },
+          tabBarLabelStyle: { fontSize: TAB_LABEL_FONT_SIZE, paddingLeft: isTablet() ? theme.spacingsPlain.xs : 0 },
```
david-venhoff, line 82: 💭 I think we can remove most of the styling here and have the bar look better than before 😄 

```ts
        screenOptions={{
          headerShown: false,
          tabBarActiveTintColor: theme.colors.background,
          tabBarStyle: {
            backgroundColor: theme.colors.primary,
          },
          tabBarLabelStyle: { fontSize: TAB_LABEL_FONT_SIZE, paddingLeft: isTablet() ? theme.spacingsPlain.xs : 0 },
        }}
```

src/routes/exercises/StandardExercisesScreen.tsx
```diff
@@ -29,7 +28,7 @@ const BadgeWrapper = styled.View`
 `
 
 const ListItemResizer = styled.View`
-  width: ${wp('85%')}px;
+  width: 85%;
```
david-venhoff: 💭 We could also use `width: 100%` and then some horizontal spacing to get rid of this number, but it does not matter much probably

src/routes/home/components/Card.tsx
```diff
@@ -18,7 +17,7 @@ const Box = styled.Pressable<{ width?: number }>`
   justify-content: space-between;
   margin: ${props => props.theme.spacings.sm};
   padding: 0 ${props => props.theme.spacings.sm};
-  min-height: ${hp('28%')}px;
+  min-height: 224px;
```
david-venhoff:
```suggestion
```
🙃 This does not seem to be needed anymore, since job cards are now quite high by default

src/routes/home/components/HomeScreenHeader.tsx
```diff
@@ -11,7 +10,7 @@ const Wrapper = styled.View`
 `
 const HeaderStyle = styled.View`
   background-color: ${props => props.theme.colors.primary};
-  height: ${hp('7%')}px;
+  height: 56px;
```
david-venhoff: 🙃 We can remove this if we also remove the `height: 100%` in the `MenuIconContainer` to have the height be calculated automatically

src/routes/repetition/RepetitionScreen.tsx
```diff
@@ -37,7 +36,7 @@ const TextContainer = styled.Text`
   color: ${props => props.theme.colors.primary};
   font-family: ${props => props.theme.fonts.contentFontRegular};
   font-weight: ${props => props.theme.fonts.lightFontWeight};
-  font-size: ${hp('1.8%')}px;
+  font-size: 14px;
```
david-venhoff: 💭 Can we maybe reuse the small font size or the default font size from the theme?

src/routes/repetition/RepetitionScreen.tsx
```diff
@@ -57,7 +56,7 @@ const EmptyStateContent = styled.View`
 const EmptyStateSubtitle = styled.Text`
   color: ${props => props.theme.colors.primary};
   font-family: ${props => props.theme.fonts.contentFontRegular};
-  font-size: ${hp('1.8%')}px;
+  font-size: 14px;
```
david-venhoff: Same here

david-venhoff (approved):
Really cool! I am very happy to see `react-native-responsive-screen` go 😄 

During the review I tried to find places where we can remove hardcoded sizes which is why there are now a bunch of suggestions, but feel free to only pick what makes sense to you :)

Looks like `useScreenHeight.ts` is now also unused and can be removed :tada:

I noticed a small inconsistency where components sometimes access theme directly and sometimes use `useTheme` to access it. But I don't think that is particularly important right now.

src/components/Button.tsx
LeandraH: I like the idea but it doesn't look good. I'm going to assume that Toni will come up with a good solution in her mini design system that will include spacing, though, so I will leave it for now.

src/components/ModalSkeleton.tsx
LeandraH: Ouch, good catch, thanks!

src/constants/theme/sizes.ts
LeandraH: Hmm, I'll leave it like this for now because it's only two values at the moment. But if we get more, I agree that we should split into two objects.

PR #1398, opened 2026-04-09
1397: Check indexed access on arrays

### LINKED ISSUE #1397: App crashes when doing a word choice exercise on a unit without any words
### Describe the Bug

It can happen by accident that units without words are published, so we should handle them gracefully.
One example, currently `Mechatroniker:in -> Elektrische Netzwerke und Widerstände` has no words

### Steps to Reproduce

1. Go to `Mechatroniker:in -> Elektrische Netzwerke und Widerstände -> Übung`
2. Notice app crash

### Expected Behavior

The app should not crash and handle the empty exercise gracefully

LeandraH:
### Short Description

I didn't know that TS doesn't enforce checks for empty array automatically, so I added a flag for TS to do that. And then had Claude fix a loooooot of files.

I tried to keep them separate in each commit, the really important commit is only the last one. Sorry about the huge PR for what should be a small feature!

### Proposed Changes

- Added the flag `noUncheckedIndexedAccess` to our `tsconfig.json`
- Added a `!` to all tests
- Added a `!` to all places where we access our own arrays (e.g. `ARTICLES`)
- Added a `!` to places where we access an index in an array by having previously found that index in the array
- Added actual safety checks in the few places where it was necessary (including in the `wordChoiceExerciseScreen`)

### Side Effects

- TS will remind us to think about whether accessing an array by index is safe

### Testing

Go to Mechatroniker:in -> Elektrische Netzwerke und Widerstände -> Übung, see that the app doesn't crash

Note: I didn't put in anything special into the empty page because it really should not be happening. Any disagreements with that?

### Resolved Issues

Fixes: #1397 

---


david-venhoff (approved):
LGTM. I verified that it fixes the crash but did not review every individual change. 

One concern I have is that the exclamation marks cause a lot of noise and might lead to us just automatically adding it after every array access without thinking what could go wrong.
Maybe we should open a follow up issue to see where we can remove the exclamation marks? For example we could (ab-)use the type system a bit to have types like NonEmptyArray or FixedSizeArray<T, 5>

LeandraH:
> One concern I have is that the exclamation marks cause a lot of noise and might lead to us just automatically adding it after every array access without thinking what could go wrong. Maybe we should open a follow up issue to see where we can remove the exclamation marks? For example we could (ab-)use the type system a bit to have types like NonEmptyArray or FixedSizeArray<T, 5>

That's a very good point, I looked into this some more and learned that when we type an array just as const (e.g. `const ARTICLES = [...] as const` instead of `export const ARTICLES: Readonly<ArticleType[]> = [...] as const`), TypeScript infers as much as it can by itself, including the length of the array. So letting TS do some more of its own typing actually makes a bunch of the `!` unnecessary 🥳 

I'll still fix that in this PR

PR #1399, opened 2026-04-09
1396: Improve release workflow

### LINKED ISSUE #1396: Improve release workflow
### Describe the Problem

I noticed two issues with our current release workflow:
- Before a new pre release, we need to manually move back the release notes
- The promote ci run marked the first pre-release as latest, instead of the most recent one

- Maybe move the release notes only after promotion?
- Fix the app toolbelt to only mark the most recent release as latest

Do nothing

### Additional Information

I have manually marked our release as latest in the github releases section

david-venhoff:
### Short Description

Fixes two small issues of our current release workflow.

### Proposed Changes

- Move release notes only when running the promote job, so that we don't have to move them back every time we do a pre-release
- Update the app toolbelt to fix a small bug we had with our github release (https://github.com/digitalfabrik/app-toolbelt/pull/55)

### Side Effects

- Should be none

### Testing

We'll probably only see whether this works at the next release 🤷 

### Resolved Issues

Fixes: #1396 

---


LeandraH (approved):
Looks good, thanks for figuring that out!

PR #1400, opened 2026-04-09
1387: Remove react-native-image-zoom-viewer

### LINKED ISSUE #1387: Remove react-native-image-zoom-viewer
### Describe the Problem

We use the archived package `react-native-image-zoom-viewer` in the `ImageCarousel`. That's dangerous because it doesn't get any security fixes. If we want to keep the ability to zoom in on images, we should replace this package.

@HuguetteAfi @hauf-toni Should we keep the ability to zoom? I didn't know until today that we have that 😅 

EDIT: As agreed in the comments, just remove the zoom.

I would like to not have `react-native-image-zoom-viewer` in our code base anymore, one way or another.

Possibly just deleting the zoom option.

david-venhoff:
### Short Description

It was decided that the niche functionality of zooming is not worth keeping a deprecated library.
To re-implement the pager functionality, this pr uses the already existing library `react-native-pager-view`.
Also removes the `minimized` prop to the image carousel. I don't know what it was used for.

### Proposed Changes

- Remove the library
- Use the pager library to re-implement the swiping functionality
- Remove unused `minimized` prop

### Side Effects

I changed the image resize mode to 'contain', from 'cover' so that images are not cut off, which might cause some more edges. If this is not wanted, I can revert that change.

### Testing

Test that images are still shown and that swiping still works. The simplest way to do so is to create a custom word with multiple images or to find a training exercise where the cms returns multiple images for a single word.

### Resolved Issues

Fixes: #1387

---

### DIFF
--- a/jest.setup.ts
+++ b/jest.setup.ts
@@ -1,6 +1,5 @@
 import mockAsyncStorage from '@react-native-async-storage/async-storage/jest/async-storage-mock'
 import '@testing-library/jest-native/extend-expect'
-import { Component } from 'react'
 // @ts-expect-error file is js
 import mockRNDeviceInfo from 'react-native-device-info/jest/react-native-device-info-mock'
 import 'react-native-gesture-handler/jestSetup'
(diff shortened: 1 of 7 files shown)


JoeyStk (approved):
Thank you very much! Looks good, and I could find any issues while testing! Good job :)

PR #1405, opened 2026-04-15
1404: Use distinct package id for debug builds on android

### LINKED ISSUE #1404: Simplify debugging on a real device
### Describe the Problem

Right now it is quite annoying to debug on my smartphone because I have to uninstall the actual lunes app each time I want to run a devlopment version.

I think the simplest solution here is to configure android to build the debug app under a different id (e.g. `app.lunes.debug`), so that android will treat the full-release version and the debug version as two distinct apps.

- Do nothing and uninstall the release version each time I want to do some development on my smartphone
- Make the debug version greater than the release version: But then I wouldn't be able to downgrade back to the release version
- Maybe something entirely different?

david-venhoff:
### Short Description

This enables installation alongside the production version of lunes to make it easier to debug on personal devices.
Also changes the name to Lunes (Debug) and adds a slight tint to the icon so that the different build variants can be distinguished more easily.

### Proposed Changes

- add a suffix to the package id so that android treats the debug builds as a distinct app to the release build
- Change the name and icon of debug builds so that they won't be confused with a potential existing production installation of lunes

### Side Effects

- This does not change anything on ios, because I don't know if such a problem also exists on iOS. @LeandraH do you have an opinion on this?

### Resolved Issues

Fixes: #1404 

---

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -111,6 +111,7 @@ android {
     buildTypes {
         debug {
             signingConfig signingConfigs.debug
+            applicationIdSuffix = ".debug"
         }
         release {
             // Caution! In production, you need to generate your own keystore file.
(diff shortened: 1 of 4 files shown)


JoeyStk (approved):
Thank you very much! Makes sense and looks good! :)

PR #1408, opened 2026-04-22
Fix the CI

david-venhoff:
### Short Description

The CI broke last week, because the runner was no longer able to access our private app signing repository.
The reason for that appears to be that there is a mismatch between the key that we configured at [circle ci](https://app.circleci.com/settings/project/github/digitalfabrik/lunes-app/ssh) (Section "Additional SSH Keys") and the fingerprint expected in our circle ci .yml files.
I have no idea why we suddenly have this mismatch since last week (First failing commit is 3586bf6da9af52489e07bd17169d5014cbe03124 on April 15th). The only explanation I can think of is that someone modified the circle ci key for some reason without notifying us about that?

### Proposed Changes

- Update the expected fingerprint with the fingerprint of the key specified in circle ci

### Side Effects

- should be none

### Testing

I ran a commit_main workflow for this pr which completed successfully: https://app.circleci.com/pipelines/github/digitalfabrik/lunes-app/4543/workflows/31568908-3fa4-4b40-9fe4-4495bf148ad6

### Resolved Issues

Fixes: our ci

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -688,7 +688,7 @@ workflows:
                     - check
                     - bump_version
         when:
-            or:
+            and:
                 - << pipeline.parameters.run_commit >>
                 - equal:
                     - main
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -688,7 +688,7 @@ workflows:
                     - check
                     - bump_version
         when:
-            and:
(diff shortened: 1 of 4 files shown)


JoeyStk (approved):
Thank you very much! LGTM!

PR #1414, opened 2026-04-27
1413: Use strings as exercise keys

### LINKED ISSUE #1413: Handle new exerciseKeys
### Describe the Problem

In https://github.com/digitalfabrik/lunes-cms/pull/709/changes, we changed the exercise keys to strings from numbers. We need to update that in the App.

Note: While we will lose some data if there is a mismatch between the CMS and App releases here, the user won't notice it, it will be sent to Sentry if tracking is enabled. So we'll probably get some annoying errors on Sentry that we can't do anything about, but we won't be bugging the user.

Update the code to use strings for the exerciseKeys and add a migration.

Nope.

LeandraH:
### Short Description

We changed the exercise keys in the CMS to use strings instead of numbers, so I updated that in the app code.

### Proposed Changes

- Updated the exercise keys to use strings
- Slight refactoring

### Side Effects

- When there's a mismatch between the app and the CMS here, the CMS will reject the tracking posts. That won't be shown to the user but sent to Sentry (if tracking is enabled, which it has to be if we're trying to track a user action)

### Testing

Open the app against the CMS on develop. Have tracking enabled. See that you don't get 500 errors anymore when doing actions that we track.

### Resolved Issues

Fixes: #1413 

---


src/constants/data.ts
david-venhoff:
```suggestion
export const EXERCISE_FOR_REPETITION = ExerciseKeys.wordChoiceExercise
```
🙃 Maybe we should rename this constant, since the exercise keys are no longer orderable so that we don't accidentally reintroduce something like `exercise >= FIRST_EXERCISE_FOR_REPETITION`

src/routes/exercise-finished/ExerciseFinishedScreen.tsx
```diff
@@ -43,7 +43,8 @@ const ExerciseFinishedScreen = ({ navigation, route }: ExerciseFinishedScreenPro
   const score = calculateScore(results)
 
   const { exercise: notNeededForNavigation1, results: notNeededForNavigation2, ...navigationParams } = route.params
-  const repeatExercise = (): void => navigation.popTo(EXERCISES[exercise].screen, { ...navigationParams })
+  const repeatExercise = (): void =>
+    navigation.popTo(EXERCISES.find(exerciseItem => exerciseItem.key === exercise)!.screen, { ...navigationParams })
```
david-venhoff:
```suggestion
  const repeatExercise = (): void => navigation.popTo(EXERCISES[exercise].screen, { ...navigationParams })
```

💭 Since exercises no longer have a numeric id, we could maybe just change the defintion of `EXERCISES` to something like the following to simplify access:
```ts
export const EXERCISES: Record<ExerciseKey, Exercise> = {
  [ExerciseKeys.vocabularyList]: {
    key: ExerciseKeys.vocabularyList,
    title: labels.exercises.vocabularyList.title,
    description: labels.exercises.vocabularyList.description,
    level: 0,
    icon: TrainingSentences,
    screen: 'VocabularyList',
  },
  [ExerciseKeys.wordChoiceExercise]: {
    key: ExerciseKeys.wordChoiceExercise,
    title: labels.exercises.wordChoice.title,
    description: labels.exercises.wordChoice.description,
    level: 1,
    icon: TrainingSpeech,
    screen: 'WordChoiceExercise',
  },
}
```

src/services/helpers.ts
```diff
@@ -123,20 +122,14 @@ export const getNextExercise = async ({ progress, job }: GetNextExerciseParams):
   if (!firstUnfinishedUnit) {
     return {
       unit: units[0]!,
-      exerciseKey: 0,
+      exerciseKey: ExerciseKeys.vocabularyList,
     } // TODO #965: show success that every exercise is done
```
david-venhoff: Unrelated, but maybe this is a good opportunity to remove the comment, since #965 is not planned any more

david-venhoff (approved):
Nice refactoring :)
LGTM

src/constants/data.ts
LeandraH: Good point, thanks!

src/routes/exercise-finished/ExerciseFinishedScreen.tsx
LeandraH: Great idea!

PR #1415, opened 2026-04-27
Fix color of settings icon in header

LeandraH:
### Short Description

Sorry, I apparently deleted the color of the settings icon during my recent styling updates. This puts it back in, so that you have a white settings wheel on a dark blue background, instead of a black one on a dark blue background.

### Testing

Open the app and see that you can see the settings wheel

---


david-venhoff (approved):
:)

PR #1419, opened 2026-04-28
1261: Add tracking for repetition of exercises

### LINKED ISSUE #1261: [DT-APP-5-1] Exercise Repetition Tracking
**Scope:** Track when exercises are started.

**Tasks:**
1. Track each exercise entry with session-scoped counter
2. Increment counter for same exercise+unit combination within session
3. Reset counters on new session
4. Event: `{ type: 'exercise_start', exercise_type, unit_id, session_count }`

**References:**
- Exercise entry points: `lunes-app/src/navigation/HomeStackNavigator.tsx`

LeandraH:
### Short Description

This PR adds tracking for repeating exercises

### Proposed Changes

- Send a tracking event to the CMS whenever an exercise is started (they are aggregated and counted in the CMS)
- Pull the management of the session ID into a separate helper function (it was only used inside the session tracking hook before)

### Side Effects

- Should be none

### Testing

Probably still against a local CMS running develop, start an exercise (word list or word choice), go into the settings to see your tracked events, see the repetition event

### Resolved Issues

Fixes: #1261 

---

### DIFF
--- a/src/hooks/__tests__/useTrackSession.spec.ts
+++ b/src/hooks/__tests__/useTrackSession.spec.ts
@@ -2,12 +2,17 @@ import { act, renderHook } from '@testing-library/react-native'
 import { mocked } from 'jest-mock'
 import { AppState, AppStateStatus } from 'react-native'
 
-import { generateUniqueId, trackEvent } from '../../services/AnalyticsService'
+import { trackEvent } from '../../services/AnalyticsService'
+import { getCurrentSessionId, rotateSessionId } from '../../services/SessionService'
 import useTrackSession from '../useTrackSession'
 
 jest.mock('../../services/AnalyticsService', () => ({
   trackEvent: jest.fn(),
-  generateUniqueId: jest.fn(() => 'session-1'),
+}))
+
+jest.mock('../../services/SessionService', () => ({
+  getCurrentSessionId: jest.fn(() => 'session-1'),
+  rotateSessionId: jest.fn(),
 }))
 
 jest.mock('../useStorage', () => ({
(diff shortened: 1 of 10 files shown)


src/hooks/__tests__/useTrackSession.spec.ts
```diff
@@ -55,8 +60,12 @@ describe('useTrackSession', () => {
...
   })
 
-  it('should track session_end and generate new session ID when transitioning to background', () => {
-    mocked(generateUniqueId).mockReturnValueOnce('session-1').mockReturnValueOnce('session-2')
+  it('should track session_end and rotate session ID when transitioning to background', () => {
+    mocked(getCurrentSessionId)
+      .mockReturnValueOnce('session-1')
+      .mockReturnValueOnce('session-1')
+      .mockReturnValueOnce('session-2')
+      .mockReturnValueOnce('session-2')
```
JoeyStk, line 68: Why do we need to do this? :)

JoeyStk (approved):
Thank you very much! I think this looks good, and I don't think that my question is going to be a blocker, therefore I'm already going to approve this :)

src/hooks/__tests__/useTrackSession.spec.ts
LeandraH: Because we want to return different mocked values (we usually can only return one mocked answer with every call). If we want to change it, we have to do the mockReturnValueOnce we have here. We need to do that here because getCurrentSessionId is called at the beginning and at the end of a session, so twice per session here. It's quite the implementation detail but I didn't really find a better way of testing it 😢

PR #1420, opened 2026-05-05
Hotfix: Fix white screen on restart

LeandraH:
### Short Description

The situation is described here:
https://chat.tuerantuer.org/digitalfabrik/pl/fft3ow8e5pr57kezbsb1syonay

It looks like we might be having a problem with our migrations: On a fresh installation, the storage version is never written to the AsyncStorage (only if `lastVersion !== STORAGE_VERSION`, which is not the case on a fresh installation, because we set `lastVersion` to be `STORAGE_VERSION` on a fresh installation). When the app then opens a second time, we try to migrate again, since there is no version in the AsyncStorage. This then breaks when trying to migrate the shape of the `WordNodeCard` because the old shape is already gone.

### Proposed Changes

- Always write the version to the AsyncStorage
- Update `migrate4To5` to also work if that migration has already been applied

### Side Effects

- We hopefully don't have white screens for our users anymore

### Testing

I haven't been able to reproduce this issue at all, so this is a pretty theoretical exercise to me.

### Resolved Issues

It's a hotfix

---

### DIFF
--- a/src/services/storageUtils.ts
+++ b/src/services/storageUtils.ts
@@ -250,7 +250,7 @@ export const migrate3To4 = async (): Promise<void> => {
 // Replaces full VocabularyItem stored in WordNodeCards with only the VocabularyItemId
 export const migrate4To5 = async (): Promise<void> => {
   type OldWordNodeCard = {
-    word: { id: VocabularyItemId }
+    word?: { id: VocabularyItemId }
     section: WordNodeCard['section']
     inThisSectionSince: string
   }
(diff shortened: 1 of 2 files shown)


src/services/storageUtils.ts
```diff
@@ -339,9 +357,7 @@ export const migrateStorage = async (): Promise<void> => {
     await migrateApiEndpointUrl()
   }
 
-  if (lastVersion !== STORAGE_VERSION) {
-    await AsyncStorage.setItem('version', STORAGE_VERSION.toString())
-  }
+  await AsyncStorage.setItem('version', STORAGE_VERSION.toString())
```
svenseeberg, line 363: How does this add to the white screen issue? I suppose migrations won't work if the last and current version are not correctly detected and therefore running them on each start just introduces a performance penalty?

svenseeberg: I see, with the other changes the migrations are attempted on each start to finish pending ones?!

svenseeberg (approved):

src/services/storageUtils.ts
LeandraH: On a fresh installation, we're setting `const lastVersion = await getStorageVersion()`, and `getStorageVersion()` returns `STORAGE_VERSION` on a new installation, so effectively `const lastVersion = STORAGE_VERSION`. With this check, we then never write the current storage version (for newly installed apps). That's fine the first time around, but when opening the app a second time, we are then back to the situation that there is no version saved, so we run through the migrations again.

So my idea here is to fix both problems:
1. Write the storage version so that we don't run through the migration each time (not really for performance, more for not breaking if the migrations aren't written for that case, but the performance could become a factor too)
2. Rewrite the migrations to not break if a migration is applied a second time.

Fixing only one should be enough, but one more redundancy shouldn't hurt.

PR #1422, opened 2026-05-06
1421: Fix the color of the status bar

### LINKED ISSUE #1421: Wrong status bar color on the home screen
### Describe the Bug

Tested on android.
When opening the app, the status bar has white text color, which is correct. Then, after navigating to e.g. the search screen, the status bar text becomes black, which is also correct. But then going back to the home screen leaves the text black, which causes the status bar to be unreadable due to bad contrast.

Image

### Steps to Reproduce

1. Go to Home screen initially, notice white text
2. Click on 'Search', notice black text
3. Go back, notice that the text is still black

### Expected Behavior

The status bar text color should have good contrast to be readable.
### Additional Information

I think react leaves both status bar components mounted and then the most recent status bar wins. We should probably unmount a status bar if it is not rendered

david-venhoff:
### Short Description

When there are multiple status bars mounted, react native always lets the most recently mounted 'win'. The behavior we want, however, is that only the currently active status bar is used.

### Proposed Changes

- Only render the status bar if the current screen has focus, to prevent multiple status bars from "fighting" with each other

### Side Effects

- Should be none

### Testing

Test that the color of the status bar now changes correctly when going back to the home screen

### Resolved Issues

Fixes: #1421

---

### DIFF
--- a/jest.setup.ts
+++ b/jest.setup.ts
@@ -43,6 +43,12 @@ jest.mock('@react-navigation/elements', () => ({
   useHeaderHeight: jest.fn().mockImplementation(() => 200),
 }))
 
+jest.mock('@react-navigation/native', () => ({
+  ...jest.requireActual('@react-navigation/native'),
+  useIsFocused: () => true,
+  useFocusEffect: jest.fn(),
+}))
+
 beforeEach(() => {
   jest.clearAllMocks()
   mockAsyncStorage.clear()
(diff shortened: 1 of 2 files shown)


LeandraH (approved):
I can't reproduce the issue on emulated iOS or Android but it seems like a reasonable solution?

david-venhoff:
Weird, I could reliably reproduce it on my smartphone and emulator

PR #1425, opened 2026-05-06
1423: Remove top padding from the tab bar

### LINKED ISSUE #1423: Navigation buttons are cut off on android
### Describe the Bug

On android the navigation buttons are cut off. I did not notice this before, because the issue is not really noticeable when the gesture navigation mode is enabled, but it looks quite bad with button navigation enabled:

Image

### Steps to Reproduce

1. Enable 3 button navigation in the system settings
2. Open app
3. Notice that the bottom tab navigation icons are cut off

### Expected Behavior

The icons should not be cut off

david-venhoff:
### Short Description

On some android configurations it caused the buttons to be cut off.

### Proposed Changes

- Remove the padding

### Side Effects

On all devices the padding is now reduced

### Testing

Test with an android device that is configured for 3-button navigation in the system settings

### Resolved Issues

Fixes: #1423 

---

### DIFF
--- a/src/navigation/BottomTabNavigator.tsx
+++ b/src/navigation/BottomTabNavigator.tsx
@@ -78,7 +78,6 @@ const BottomTabNavigator = (): ReactElement | null => {
           tabBarStyle: {
             backgroundColor: theme.colors.primary,
           },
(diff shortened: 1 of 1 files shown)


JoeyStk (approved):
Thank you very much! Looks good!

PR #1427, opened 2026-05-06
1409: Automatic screenshots

### LINKED ISSUE #1409: Automatically update screenshots in app stores
### Describe the Problem

We have so far added the screenshots in the app stores manually. Fastlane offers the possibility to do it automatically for [iOS](https://docs.fastlane.tools/getting-started/ios/screenshots/) and [Android](https://docs.fastlane.tools/getting-started/android/screenshots/). It seems like a good idea to do that, maybe with an environment variable that we set in CircleCI to decide if the screenshots should be done or not, to not waste tokens with every single release.

Have a new variable for updating the screenshots (default false)
If that variable is set to true, automatically take and upload the screenshots (it doesn't have to be done through Fastlane, but that seems like a good opportunity)

We could do it manually because it's not often (though that is part of the problem, nobody is responsible and has the right permissions)
We could use a different automation, maybe even in some way combined with e2e tests

### Related Issues

Well, we just published a major rewrite, and the screenshots aren't up to date...

https://apps.apple.com/us/app/lunes/id1562834995

https://play.google.com/store/apps/details?id=app.lunes&hl=en

david-venhoff:
### Short Description

Experimentally add [maestro](https://github.com/mobile-dev-inc/maestro) to automatic our screenshots.
This pr does not use the screenshot taking described in the fastlane docs, because that is platform dependent, which means we would have to maintain one script for taking screenshots on android and another script for taking screenshots on iOS, which seemed like too much effort to me.
I did not add CI integration yet, which means that the commands will have to be run locally, because I'd prefer to implement this in a separate pr if we decide to use this solution.

I also decided to store the screenshots in the repo instead of taking them automatically when uploading the app, so that we have a chance to make sure that everything looks good. The disadvantage is that this introduces some binary artifacts into the repo, which will increase its size.

The screenshot taking process should be mostly independent of the data on the CMS, but it will break if we make big changes to the app UI. The script should always be quick to adapt though.

### Proposed Changes

- Add the maestro automation tool to automatically take screenshots of interesting parts of the app
- Add a test mode which loads some demo data and makes sure that everything is deterministic
- Add the screenshots
- Instruct fastlane to upload the screenshots when we upload binaries to the app stores

### Side Effects

- Should be none

### Testing

Follow the instructions in `docs/screenshots` and make sure that everything works. I am also curious about whether you think this is a good approach.
I did not test whether the automatic fastlane upload works. I think we will just have to wait until the next release to see if it works.

### Resolved Issues

Fixes: #1409 

---


tools/screenshots-ios
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+OUTPUT_DIR="ios/fastlane/screenshots/de-DE"
+
+# Pin the simulator's status bar so screenshots are reproducible
+trap 'xcrun simctl status_bar booted clear' EXIT
+xcrun simctl status_bar booted override \
+  --time "9:41" \
```
LeandraH, line 9: 😍

src/services/screenshotData.ts
```diff
@@ -0,0 +1,45 @@
...
+const getProgress = async (): Promise<Progress> => {
+  const units = await getUnitsOfJob({ type: 'standard', id: jobId })
+  const unitProgress = { [ExerciseKeys.wordChoiceExercise]: 1, [ExerciseKeys.vocabularyList]: 1 }
+  return Object.fromEntries(units.map(unit => [unit.id.id, unitProgress]))
+}
+
+export const seedScreenshotData = async (storageCache: StorageCache): Promise<void> => {
+  // Make every call site that reaches for randomness deterministic so the
+  // captured screens look identical across runs. Only effective until the app is restarted.
+  Math.random = () => 0
```
LeandraH, line 36: Thank you for the explanation about the restart but it's still freaking me out to rewrite a global function 😅  Could we reverse the assignment at the end of the function?

LeandraH (approved):
I just love watching automation tests run 😍 

This is very very cool, and I think it's a great solution to get the screenshots!

About using Maestro instead of Fastlane, it looks to me like we are still maintaining two screenshot files 🙃 but I also don't know how much more duplication would be in Fastlane files. Either way, I'm happy to go with this!

Tested on emulated iOS and Android

david-venhoff:
> About using Maestro instead of Fastlane, it looks to me like we are still maintaining two screenshot files 

What I meant is that we only need one file that instructs which actions should be taken in the app (.maestro/screenshots.yaml). With the alternative solution we would have needed one such implementation for ios and one for android, while making sure that they don't diverge too much.

src/services/screenshotData.ts
david-venhoff: Not really, because the function only sets up the demo data, but we need the deterministic `Math.random` behavior during the entire screenshotting process.
I guess we could add another function that would be somehow called at the end of the screenshotting process, but that is probably not worth the effort, since an app restart would do the same.
But I definitely agree that it is very ugly 😅

PR #1436, opened 2026-05-19
1435: Add tracking for the training exercises

### LINKED ISSUE #1435: Add tracking to training exercises
### Describe the Problem

We want to also track training exercises. There already is a [CMS issue](https://github.com/digitalfabrik/lunes-cms/issues/699) for it, and a [PR](https://github.com/digitalfabrik/lunes-cms/pull/765). We should use that to also send actual tracking information over.

Send tracking information to the CMS for the three training exercises for the duration, the dropout point, and the repetitions.

LeandraH:
### Short Description

This PR adds tracking (dropout, repetition, duration) to the three training exercises

### Proposed Changes

- Change the tracking payload to use exerciseKey as created [here](https://github.com/digitalfabrik/lunes-cms/pull/765)
- Rename ExerciseKey to StandardExerciseKey, which cause half the file changes (sorry!)
- Track dropout, repetition and duration in the three training exercises

### Side Effects

- Should be none

### Testing

Test against CMS branch develop, have tracking enabled, go into a few training exercises, see in the settings > Nutzerdaten exportieren that there are now tracking data for the training exercises

### Resolved Issues

Fixes: #1435 

---


src/hooks/useTrackDropout.ts
```diff
@@ -1,39 +1,42 @@
...
+  vocabularyItemId?: number,
 ): { markCompleted: () => void } => {
   const storageCache = useStorageCache()
   const isCompletedRef = useRef(false)
 
-  const markCompleted = (): void => {
+  const markCompleted = useCallback((): void => {
     isCompletedRef.current = true
-  }
+  }, [])
```
david-venhoff:
```suggestion
  const markCompleted = (): void => {
isCompletedRef.current = true
  }
```

🙃 We don't actually need `useCallback` for those things anymore, because the react compiler will handle them automatically (https://react.dev/learn/react-compiler/introduction#what-does-react-compiler-do)

src/routes/choice-exercises/components/WordChoiceExercise.tsx
```diff
@@ -115,7 +116,15 @@ const WordChoiceExercise = ({
...
   const count = vocabularyItems.length
 
-  const { markCompleted } = useTrackDropout(navigation, unitId, state.currentWord, count)
+  const exerciseKey: StandardExerciseKeyPayload | null = useMemo(
+    () =>
+      unitId !== null
+        ? { type: 'exercise', exercise_type: StandardExerciseKeys.wordChoiceExercise, unit_id: unitId.id }
+        : null,
+    [unitId],
+  )
```
david-venhoff: 💭 This is a bit unfortunate, because react does not actually guarantee that `useMemo` memoizes the value and instead says that `useMemo` should only be used as a performance optimization: https://react.dev/reference/react/useMemo#preventing-an-effect-from-firing-too-often

But I also don't really know any better way of making sure that the `useTrackDrop` hook does not re-run every time :(
I guess the current approach is the best we can do without making the code much less readable

david-venhoff (approved):
Very nice, LGTM

src/routes/choice-exercises/components/WordChoiceExercise.tsx
david-venhoff: Although maybe we could create a new hook like `useStandardExerciseKey` (and mirrored `useTrainingExerciseKey`) that internally uses `useRef`? That would be guaranteed to be correct and also make it a bit more explicit why the exercise keys should not change.

src/hooks/useTrackDropout.ts
LeandraH: Awesome, good catch, thank you for continuously reminding me! I'm trying to remember it 😅

src/routes/choice-exercises/components/WordChoiceExercise.tsx
LeandraH: Great idea!

PR #1442, opened 2026-05-31
Don't throw when running browserstack on main

### LINKED ISSUE #1410: Don't throw when running a Browserstack delivery without PR
### Describe the Problem

When starting the pipeline for a Browserstack delivery from main, we get an error in the pipeline because it tries to write a link to the PR for the connected branch. We should not get that error.

I would like to not get an error thrown when starting a Browserstack delivery from main. It would be even nicer if I get a Browserstack link, but that is optional.

We could leave it the way it is.

david-venhoff:
### Short Description

This pr fixes the CI failure that happens when creating a browserstack build on main

### Proposed Changes

- Don't try to post a message on a non-existing pull request
- Instead, post the message to mattermost

### Side Effects

- none

### Testing

We'll see if this works in the next browserstack run

### Resolved Issues

Fixes: #1410 

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ commands:
             failure_mentions:
                 default: '@david.venhoff @leandra.hahn @johannes.stock'
                 type: string
-            only_for_branch:
+            on_failure_only_for_branch:
                 default: main
                 type: string
             success_mentions:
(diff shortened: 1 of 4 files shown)


LeandraH (approved):
Looks great! 🥳

PR #1443, opened 2026-06-03
1437: Match behavior in training exercises to that in standard exercises

### LINKED ISSUE #1437: Have the behavior of training exercises match the standard exercises better
### Describe the Problem

I noticed a few discrepancies in behavior between the training exercises and the standard exercises:
- when trying to leave a training exercise early, I don't have to confirm that I want to leave; I do have to confirm in a standard exercise
- we have a Solve All / Fail All button in the speech training exercise but not in the image or sentence training ones
- the number of possible retries is different
- we can skip a word in all exercises but in the standard exercises, that word gets sent to the end of the stack, in the training exercises, it is not checked again
- there is no feedback button in the training exercises but it is in the header in the standard exercises

Have the behavior be the same between standard and training exercises

Nobody has complained so far

LeandraH:
### Short Description

This PR makes a couple of small changes in the training exercises to match the standard exercises better.

### Proposed Changes

- When trying to leave a training exercise early, I am now asked to confirm
- In the header of the training exercises, we now have the button for sending feedback
- There are "Solve All" and "Cheat All" buttons in the training exercises (in dev mode)
- All exercises have the same number of max retries
- When skipping a word, it is now presented later in the exercise, as opposed to not being presented again
- Some refactoring

### Side Effects

- I refactored the VocabularyItemBuilder to have unique IDs for easier testing but that shouldn't have any side effects

### Testing

Go through the training exercises, see that you're asked to confirm when leaving, see that there is a feedback button in the header, see how many retries you can get, see that a skipped word is shown again later, see the cheat buttons (in dev mode)

### Resolved Issues

Fixes: #1437 

---


src/routes/training/ImageTrainingScreen.tsx
```diff
@@ -63,31 +77,63 @@ export type Action =
...
   | { type: 'nextWord' }
+  | { type: 'skip' }
+  | { type: 'cheatAll'; result: SimpleResult }
 
 // eslint-disable-next-line consistent-return
 export const stateReducer = (state: State, action: Action): State => {
   switch (action.type) {
     case 'selectAnswer': {
-      if (state.answer?.isCorrect) {
+      const hasReachedMaxAttempts = state.incorrectAttemptsForCurrentWord >= NUMBER_OF_MAX_RETRIES
```
david-venhoff, line 87: 💭 Could you add a comment to the `NUMBER_OF_MAX_RETRIES` constant that it does not really make sense to increase more due to its usage in the ImageTrainingScreen

david-venhoff (approved):
Nice LGTM

src/routes/training/ImageTrainingScreen.tsx
LeandraH: Good idea!

PR #1457, opened 2026-06-23
Update readme for volunteering

LeandraH:
### Short Description

Just updating our Readme with a link to the new wiki page for volunteers, which will presumably be better maintained 🤪


JoeyStk (approved):
Good idea! Thank you very much for keeping an eye on the Read Me! 😊 
Looking good :)

PR #1459, opened 2026-06-30
1458: Pause analytics tracking background

### LINKED ISSUE #1458: Pause analytics tracking in background
### Describe the Problem
Currently we send analytics data to our cms backend.
We realized that sometimes we have `duration_in_seconds` of about 10000s or more which are more than 2 hours. This may occur because the user starts the unit, moves the app to background for some time and continues at a later point.
I already reproduced this behavior

- count only the time that the app is in foreground

f1sh1918:
### Short Description

Adjust the hook for time tracking, to only rely on time that the app is in foreground to avoid wrong time tracking if the user moved the app in background and continues the session later.

### Proposed Changes

- measure only time that the app is in foreground, adjust tests
- rename hook
- adjust imports

### Side Effects

- should be none

### Testing

Start local Lunes CMS backend:
- `export LUNES_CMS_ALLOWED_HOSTS=10.0.2.2` (needed for android) then run backend

Lunes App:
- allow analytics data
- switch to backend to local endpoint
- choose a profession
- go to f.e. "word list" stay so seconds there
- move app in background for a minute f.e.
- move app in foreground
- leave wordlist
- check your db table for the particular event `analytics_analyticsevent` 
Mine was f.e.
```
{"exercise_key": {"type": "exercise", "exercise_type": "word_list", "unit_id": 28}, "duration_seconds": 24}
```
### Resolved Issues

Fixes: #1458 

---


david-venhoff (approved):
Nice, LGTM

PR #1466, opened 2026-07-09
1403: Don't limit screen orientation on android

### LINKED ISSUE #1403: Don't limit screen orientation on android
### Describe the Problem

We have this line in our android manifest:
https://github.com/digitalfabrik/lunes-app/blob/3586bf6da9af52489e07bd17169d5014cbe03124/android/app/src/main/AndroidManifest.xml#L19

The google play console warns that this setting will no longer be respected by devices running android 16, so we should probably remove this setting and verify that the app works in landscape mode.
I just did a quick check and for multiple layout glitches, so it won't be as simple as just removing the flag.

Remove the line from the android manifest and fix all layout glitches in the landscape mode

- Do nothing
- Show some kind of warning in landscape mode

david-venhoff:
### Short Description

Removes the limitation that the app may not be used in landscape mode and slightly adapts the UI to make the app also usable in the landscape mode.
It does not look great, especially on smartphones, but it looks okay in landscape mode. I think that is okay, since the app is mostly intended to be used on smartphones.

### Proposed Changes

- Allow non-portrait usage in the android manifest
- Respect left and right insets in the route wrapper. This should prevent UI being behind androids navigation bar
- Make all screens scrollable that weren't before
- Adjust the layouts of the image training and speech training modes to reduce the image size in landscape mode

### Side Effects

- While this is mostly a fix for android compatibility, it should also improve ios landscape mode

### Testing

Test on a tablet and maybe a smartphone in landscape mode, the app should be usable and not look too bad

### Resolved Issues

Fixes: #1403 

---

### DIFF
--- a/src/routes/settings/SettingsScreen.tsx
+++ b/src/routes/settings/SettingsScreen.tsx
@@ -17,6 +17,8 @@ import { getLabels } from '../../services/helpers'
 import DebugModal from './components/DebugModal'
 import VersionPressable from './components/VersionPressable'
 
+const Root = styled.ScrollView``
+
 const SettingsHeading = styled(Heading)`
   padding: ${props => props.theme.spacings.xl};
   text-align: center;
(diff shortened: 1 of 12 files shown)


src/routes/settings/SettingsScreen.tsx
```diff
@@ -17,6 +17,8 @@ import { getLabels } from '../../services/helpers'
 import DebugModal from './components/DebugModal'
 import VersionPressable from './components/VersionPressable'
 
+const Root = styled.ScrollView``
```
LeandraH, line 20: ❓ Is this just for renaming or is there a functional reason here as well?

src/routes/training/components/ImageGrid.tsx
```diff
@@ -58,14 +56,19 @@ const StyledImage = styled.Image<{ width: number; padding: number; state: ImageG
...
   const paddingPx = theme.spacingsPlain.xs
   // let's be a bit defensive here so that there are no unwanted overflows in the flow layout
-  const imageWidth = Math.floor((availableWidth - 2 * NUM_COLUMNS * paddingPx) / NUM_COLUMNS - 1)
+  const imageWidth = Math.floor((availableWidth - 2 * numColumns * paddingPx) / numColumns - 1)
 
   const onGridLayout = (event: LayoutChangeEvent): void => {
     if (availableWidth !== event.nativeEvent.layout.width) {
       setAvailableWidth(event.nativeEvent.layout.width)
     }
+    setIsLandscape(event.nativeEvent.layout.width > event.nativeEvent.layout.height)
```
LeandraH: ❓ Why didn't you use `useWindowDimensions` here?

LeandraH (commented):
Nice, looks good in general. I decided to only test it in iOS so far (just to make your life difficult 🤪), and I found two fun bugs when the device decides for me to switch back into portrait mode:

https://github.com/user-attachments/assets/7274ec91-3341-42e7-9174-15b7f754764d

https://github.com/user-attachments/assets/ad158d2c-23cb-4771-afb9-90cbe30397b0

Oh, and it apparently decided to keep the videos in portrait mode. Yay. Sorry about that, the emulator was in landscape mode at the beginning and at the end of the videos.

src/routes/settings/SettingsScreen.tsx
david-venhoff: It is for making the settings screen scrollable. Before the content would just cut off at the bottom if the screen height was too small

src/routes/training/components/ImageGrid.tsx
david-venhoff: No reason in particular, I just noticed that we listen for size here anyways and can reuse the listener for that. I'll update it to `useWindowDimensions`

david-venhoff:
Good catch! I fixed our modalSkeleton, bottomSheet and the data collection popup. I left the modals for capturing an image and capturing audio when creating custom vocabulary, since this should not be very annoying to deal with and would require some more layout fixes

LeandraH (approved):
Looks good to me! I guess we might want to add a release note because this is a user-facing change? Then again, we might not want to encourage people to use the landscape orientation 😅

david-venhoff:
I think I agree that it is probably better not to advertise this feature 😄

PR #1468, opened 2026-07-09
1467: Fix job selection when keybard is open

### LINKED ISSUE #1467: Job selection with open keyboard: select job on first tap
### Describe the Problem

When searching in the jobs and then tapping on the one you found, the first tap just closes the keyboard, the second tap selects the job.

The first tap should close the keyboard AND select the job.

N/A

### Additional Information

N/A

LeandraH:
### Short Description

When searching in the job selection and then tapping on a job, this PR closes the keyboard and selects the job at once.

### Testing

Go into the job selection (by pressing the plus in the home screen), type something into the search field, tap on a job, see that it is selected on the first tap, not just on the second one.

### Resolved Issues

Fixes: #1467 

---


JoeyStk (approved):
Looks good! Not tested

PR #1470, opened 2026-07-10
1469: LLM code review

### LINKED ISSUE #903: LUN-475: Long term exercise service
# Stefanie Metzger - 20.2.2023, 14:06:08

Implement a long term exercise class (service) that contains all logic needed for the long term exercise, this should be implemented with the singleton pattern.

There are 7 sections

 

|section|needs repetition after|
| --- | --- |
|0|0 days|
|1|1 day|
|2|3 days|
|3|7 days|
|4|30 days|
|5|90 days|
|6|never|

 

Progress is saved in AsycStorage, implement methods for saving, modifying and deleting here and not in AsyncStorageService.

 

Class fields:
* words of each secction (there are 7 secctions in total
* for every word we need to save the word, the secction it is currently in, the date when it was added to this secction

Methods that should be implemented:
* get number of words that can currently be repeated.
* get number of words that can currently be repeated, with a upper bound, so if more than 10 words can be repeated return 10
* get words for next exercise (contains either all words that need to be repeated, if the number of these is <= 10 or 10 words that need repetition. Think of a valid algorithm, best would be to take these words in the next exercise that would have needed to be repeated most time ago relative to the time span in their section, e.g. if not all words that need repetition can be repeated in this exercise, because there are more than 10, wordA which is in section 2 and needed repetition 2 days ago and wordB which is in section 3 and needed repetiotion 3 days ago, wordA should be prefered over wordB)
* add words to the first section (after a module is finished)
* move word to the next section
* move word to the previous section
* get a list of numbers, with length = 7, which contains the numbers of words in each secction

 

Add a lot of tests as this feature is going to be really hard to debug.

**Environment**: -
**Linked issues:**
- Blocks: LUN-476
- Blocks: LUN-474
- Blocks: LUN-477
- Blocks: LUN-479

### LINKED ISSUE #1469: Implement LLM code review in pipeline
### Describe the Problem
Currently we are doing code reviews mostly manually or asking LLM to do one on a pr.
We could also have a basic LLM code review directly on our pipeline

- implement a basic code review on commit

N/A

### Additional Information

N/A

f1sh1918:
### Short description

Add basic code reviews to our commit workflow

### Proposed changes

- add python script that we use in salt
- adjust prompt to our project context
- add `get_access_token` script to avoid app-toolbelt

### How to test

- N/A

### Resolved issues

Fixes: #1469


.circleci/scripts/llm-pr-review.py
```diff
@@ -0,0 +1,369 @@
...
+COMMENT_MARKER = "<!-- llm-pr-review -->"
+
+# Maximum diff size to send to the LLM (200 KB). Larger diffs are truncated.
+MAX_DIFF_BYTES = 200_000
+
+SYSTEM_PROMPT = """
+You are a senior React Native/TypeScript engineer reviewing a pull request
+for lunes-app, the mobile vocabulary trainer app (iOS & Android, built with
+React Native and TypeScript) behind the Lunes project. The backend content
+(disciplines, training sets, words/units) is managed by the separate
```
JoeyStk:
```suggestion
(disciplines/jobs, training sets/units, words/units) is managed by the separate
```
disciplines and training sets are choice of words that we used in version 1, but don't use in version 2 anymore. Is this on purpose or by accident? :)

.circleci/scripts/llm-pr-review.py
```diff
@@ -0,0 +1,369 @@
...
+ Library. New or changed behavior (components, hooks, services) should
+ come with tests. If a PR changes a shared function's signature, check
+ that all call sites (including tests and mocks in `src/__mocks__`) were
+ updated.
+7. Obvious typos in code, comments, file paths, identifiers, and
+ documentation. Only flag clear spelling mistakes — do not nitpick
+ stylistic word choices.
+8. Commit message / PR title style. The repository convention (see
+ `docs/conventions.md` and `git log`) is:
+ - `<Issue key>: Your commit message`, e.g. "LUN-612: Add commit message
```
JoeyStk:
```suggestion
 - `<Issue key>: Your commit message`, e.g. "612: Add commit message
```
I don't think we use the Jira Style anymore, do we? 🤔 I think we might even remove it :)

JoeyStk (commented):
Thank you very much! This looks already really good. I just found two smaller things, we might want to think about again :)

.circleci/scripts/llm-pr-review.py
```diff
@@ -0,0 +1,369 @@
...
+   documentation" — the issue key prefix (Jira-style, e.g. `LUN-1234`, or a
+   plain issue number) should match the branch name's prefix.
+ - Additional context (why the change is necessary) goes after a blank
+   line in the body.
+ - The commit message must be generally useful: it should clearly
+   describe what changed and, in the body, why. Flag messages that are
+   vague ("fix stuff", "update"), tautological ("change X to X"), or that
+   don't explain a non-obvious change.
+ Dependabot commits are exempt from these rules.
+9. CircleCI config: `.circleci/config.yml` is auto-generated from
```
david-venhoff: Do we need this point? The ci already verifies that the config.yml is consistent with everything in `.circleci/src`

.circleci/scripts/llm-pr-review.py
david-venhoff: Looks like we forgot to remove this style from our conventions.md file

LeandraH:
I would still like to also give the AI a somewhat open possibility to review and see how that goes, give it some criteria like clean code, accessibility, maintainability, and ask it for its thoughts. It works pretty well with Claude, but might obviously be a total pain here, so maybe we can try it for a bit?

.circleci/scripts/llm-pr-review.py
f1sh1918: I mean I had to add it explicitly. Before it mentioned duplicated code

david-venhoff: In that case can we only remove the `Only check it for **consistency**` part?

.circleci/scripts/llm-pr-review.py
f1sh1918: i will do that :)

.circleci/scripts/llm-pr-review.py
f1sh1918: yes i will remove old wording

f1sh1918:
still waiting for another review

david-venhoff:
i think we should probably also mirror https://github.com/digitalfabrik/lunes-cms/pull/916 here

f1sh1918:
> digitalfabrik/lunes-cms#916

done @david-venhoff

david-venhoff (approved):
Thanks, lets give it a try 🎉

PR #1471, opened 2026-07-15
Fix inconsistent result screen for wordChoiceExercise

### LINKED ISSUE #992: Counter implementation is confusing
### Describe the Bug

counter implementation is somehow confusing (see documentation in Figma) 
counting starts with e.g. 1/11 and if 3 words are correctly answered -> counter shows 4/11
counter ist counting differently in round 3 than in round 1 and 2 within a module

Originally reported by Isabelle here: #948

### LINKED ISSUE #1446: Inconsistent result screen for wordChoiceExercise
### Describe the Bug

When answering most questions wrong initially, but getting them right later, the result screen looks pretty inconsistent:

Image

### Steps to Reproduce

1. Go to a wordChoiceExercise
2. Answer all words wrong initially, but get them right later
4. See the results screen

### Expected Behavior

The results screen should clearly explain my results

david-venhoff:
### Short Description

Instead of counting the words that have been answered correctly after any number of attempts, only count words that have been answered correctly initially.

### Proposed Changes

- Adapt the way how we count correct words in the result screen

### Side Effects

- Should be none

### Testing

Go to a word choice exercise, initially answer all words wrong and then correctly. 
Previously the screen would then show that all words are correct and now it should show that no words are correct.

### Resolved Issues

Fixes: #1446 
Fixes: #992 

---

### DIFF
--- a/src/routes/exercise-finished/ExerciseFinishedScreen.tsx
+++ b/src/routes/exercise-finished/ExerciseFinishedScreen.tsx
@@ -39,7 +39,7 @@ type ExerciseFinishedScreenProps = {
 const ExerciseFinishedScreen = ({ navigation, route }: ExerciseFinishedScreenProps): ReactElement => {
   const { exercise, results, unitTitle } = route.params
   const [isModalVisible, setIsModalVisible] = useState<boolean>(false)
-  const correctResults = results.filter(doc => doc.numberOfTries === 1)
+  const correctOnFirstAttempt = results.filter(doc => doc.numberOfTries === 1)
   const score = calculateScore(results)
 
   const { exercise: notNeededForNavigation1, results: notNeededForNavigation2, ...navigationParams } = route.params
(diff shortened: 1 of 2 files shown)


src/routes/exercise-finished/ExerciseFinishedScreen.tsx
```diff
@@ -39,7 +39,7 @@ type ExerciseFinishedScreenProps = {
 const ExerciseFinishedScreen = ({ navigation, route }: ExerciseFinishedScreenProps): ReactElement => {
   const { exercise, results, unitTitle } = route.params
   const [isModalVisible, setIsModalVisible] = useState<boolean>(false)
-  const correctResults = results.filter(doc => doc.result === 'correct')
+  const correctResults = results.filter(doc => doc.numberOfTries === 1)
```
LeandraH: 💭 I wonder if we should rename this to something like `correctOnFirstAttempt` or something like that

LeandraH (approved):
Nicely done, tested on emulated iOS

PR #1472, opened 2026-07-16
794: Update SVGs to pass in color by JS

### LINKED ISSUE #794: LUN-356: Refactoring SVG Icons in different colors
# f1sh1918 - 24.5.2022, 13:57:19

Currently we are using multiple svg files for icons that just differ in color f.e:
```java
import CheckCircleIconGreen from './check-circle-icon-green.svg'
import CheckCircleIconWhite from './check-circle-icon-white.svg' 
```
CSS enables you to use the "currentColor" Attribute to set the fill-color of a svg. So you can use the css color attribute to adjust the color you want to use instead of setting it fixed in the svg file.

**Example:**
  * ```java
<g fill="currentColor">
```
```java
const BookIconGrey = styled(BookIcon) `
color: grey;
` 
```
  * **Acceptance Criteria**
* remove duplicated svgs
* adjust the fill color to "currentColor" in the svg files
* set the color style you need with styled components for each usage of the icon

**Environment**: -
**Linked issues:**

LeandraH:
### Short Description

Mostly, this PR updates most of our SVGs to use colors passed in from the JS so that there is less duplication.

### Proposed Changes

- Updated most SVGs to use `currentColor`
- Made the file naming of the icons consistent (no `-icon` at the end)
- Refactored the SVGs themselves to be smaller (removing unneeded `<g>` and other things like bizarre translations)

### Side Effects

- The SVGs should be easier to maintain

### Testing

Go through the app and check that all icons still look good

### Resolved Issues

Fixes: #794 

---


assets/images/crystal-ball.svg
david-venhoff, line 1: I think this file is now unused and can be deleted, because its only usage in the alternative words section got replaced by the light bulb icon

david-venhoff (approved):
Wow, really cool!
I did not do a in-depth review of every file due to the big diff, but overall the changes look good to me. I only noticed the back and cancel icon (defined in NavigationHeadLeft.tsx) being grey now instead of black:

Now:

Image

Before:

Image

assets/images/crystal-ball.svg
LeandraH: Thanks! I specifically replaced it because I noticed that we had the same icon twice, and then promptly forgot to remove it 🤣

LeandraH:
> I only noticed the back and cancel icon (defined in NavigationHeadLeft.tsx) being grey now instead of black:

Thanks, good catch, it turns out that having several colors passed into an SVG doesn't quite work, so I reverted that part to using two separate SVG files, and added some documentation about it :)

PR #1475, opened 2026-07-21
1460: Upgrade node

### LINKED ISSUE #1460: Upgrade Node
### Describe the Problem

We're currently still running on Node 20 which is not supported anymore. We should update to 22 or 24.

I would prefer going straight to 24.

We could use 22, it would just mean we have to update two years earlier, but it might also mean less secondary work because nothing changes on 22 anymore. Or we could wait until something break on 20, but let's not do that.

f1sh1918:
### Short Description

Upgrade Node to LTS (24)

### Proposed Changes

- update node in `package.json`
- update circle ci jobs
- make `commit_main` executable via `Trigger Pipeline` to be able to test CI changes before getting merged

### Side Effects

- None

### Testing

N/A

### Resolved Issues

Fixes: #1460 

---


.circleci/src/workflows/commit_main.yml
LeandraH, line 1: Great idea!

LeandraH (approved):
Nice!

PR #1488, opened 2026-07-29
1485: use the pronunciation field from the CMS for the Sprechstudio

### LINKED ISSUE #1485: Use pronunciation field from CMS for Sprechstudio check
### Describe the Problem

We have some loan words like "Baiser" where the Sprechstudio will only accept a literal (i.e. false) pronunciation. Once https://github.com/digitalfabrik/lunes-cms/issues/914 is done, we should use the pronunciation field from there to test the pronunciation.

If the pronunciation field is filled in, the pronunciation should be checked against that.

If the pronunciation field is empty, the pronunciation should be checked against the actual work, like it is already.

I'm running out of ideas

### Related Issues

CMS issue: https://github.com/digitalfabrik/lunes-cms/issues/914
Rejected app idea that already cost me quite some hours of work: https://github.com/digitalfabrik/lunes-app/issues/1447

LeandraH:
### Short Description

This PR uses the pronunciation field from the CMS for the Sprechstudio

### Proposed Changes

- Use the pronunciation field from the backend
- When something is filled in for the pronunciation field, the Sprechstudio compares against that instead of the written word

### Side Effects

- If someone fills nonsense into the pronunciation field, that would be a problem
- The "Baiser" example still doesn't fully work, and it never might. With this, pronouncing it incorrectly, it fails reliably (it passed reliably before). Pronouncing it correctly only gets accepted about one in three. So, that's an improvement, but it's far from perfect.

### Testing

Before the next CMS release, test against the local main branch
Take a job that only has one word, or create a new one, fill in a different pronunciation from the actual word, check in the Sprechstudio that it tests against the pronunciation and not the word.

### Resolved Issues

Fixes: #1485 

---


f1sh1918 (approved):
Tested and works "good"
I think we will never have a perfect solution
I recognized that "die Besee" (spoken) also works even it is "das Besee" but i think some fuzzy match is needed so i think its good as it is

PR #1497, opened 2026-08-10
1493: respect reduced-motion setting

### LINKED ISSUE #1493: Handle wish for reduced motion
### Describe the Problem

We currently ignore the reduced-motion setting completely in our app, we really need to stop doing that.

When there is an animation, try to have it fade instead of moving things if reduced motion is set

Disable all animations?

### Related Issues

Came up as part of a review from Claude in #1424

LeandraH:
### Short Description

We really should have been respecting the prefers-reduced-motion setting for a long time. This PR fixes that situation.

### Proposed Changes

- Add a hook to check if reduced motion is enabled
- Use that hook in the navigators (I disabled all animation there because fade looked weird)
- Use that hook in the bottom sheet, where I did use fade
- Removed a hard-coded 300ms animation in the ListItem. The default is now 250ms, I don't expect anybody to notice the difference.

### Side Effects

- Well, I guess someone might notice a 50ms difference.

### Testing

Turn on the accessibility setting for reduced motion in the device settings, navigate around the app a bit, notice that things aren't sliding around.

### Resolved Issues

Fixes: #1493 

---

### DIFF
--- a/release-notes/unreleased/1493-reduced-motion.yml
+++ b/release-notes/unreleased/1493-reduced-motion.yml
@@ -1,0 +1,6 @@
+issue_key: 1493
+show_in_stores: true
+platforms:
+  - android
+  - ios
+de: Wer im Betriebssystem Bewegungen reduziert hat, sieht Übergänge jetzt als sanftes Ein- und Ausblenden statt als Bewegung.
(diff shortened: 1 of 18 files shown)


f1sh1918:
It looks good so far but i realized that we are importing and calling this a11y info at a lot of places.
This might be a good scenario for a context as we f.e. use for the VolumeService. So the info will only be subscribed once and all places listen to the same context.
Actually this change is not mandatory but it would follow our repo convention

LeandraH:
Great idea, thanks!

david-venhoff (approved):
Thanks, LGTM

PR #1498, opened 2026-08-11
1456: make the speech training exercise harder

### LINKED ISSUE #1456: Exercise mode 'Sprechstudio' is too forgiving
### Steps to Reproduce

1. Go to 'Vokabeln trainieren - Sprechstudio'
2. Record any word as first try to get a wrong answer and the correct word spelled out. (Or use dev-mode).
3. Once you know the correct word, record it again with the word stress on the wrong syllable or an unstressed syllable missing. 

### Expected Behavior
Word stress should be taken into consideration and marked when it is off. Either through an additional feedback that there are improvements to be made or by categorising it  as wrong. The same goes for unstressed syllables that are not pronounced at all. 

### Related Issues
Rating a word correct, even if the article was not pronounced. Seems not to have a ticket yet.

### Examples

Still correct: 
die Muffe pronounced as diiiiiii Muféeee
die Fußbodenheizung as die Bodenheizung
das Waschmaschinenventil as das Waschmachienventil

LeandraH:
### Short Description

This PR makes the Sprechstudio exercises harder, particularly for longer words where a syllable might go missing and for articles.

### Proposed Changes

- Count the syllables in a word and compare between user input and correct word
- Specifically check that the article has been said, and that it's the correct one
- Added a small hint telling the user if a syllable is missing, an article is missing, or the article is wrong

### Side Effects

- The correctness of what the app recognized as right and wrong was way better than previously, but I assume that Apple changed something about their model.

### Testing

This has been tested on my real iPhone, my Android emulator was on strike again, so testing on a real Android device would be very helpful.

Go into the Sprechstudio, and say words correctly, leave out syllables, leave out the article, say the wrong article. See how the feedback compares to what you actually said.

### Resolved Issues

Fixes: #1456 

---


src/routes/training/services/SpeechMatchingService.ts
```diff
@@ -154,36 +110,62 @@ const containsAsTokens = (transcript: string, expected: string, maxExtra: number
...
+
+  if (!hasNoArticle(article) && spokenArticle !== normalizeText(article.value)) {
+    const reason =
+      spokenArticle === null ? SPEECH_FEEDBACK_REASONS.missingArticle : SPEECH_FEEDBACK_REASONS.wrongArticle
+    return incorrect(looksLikeExpectedWord(spokenWord, normalizeText(word)) ? reason : undefined)
+  }
+
+  // Items without an article are graded on the word alone, so an article the recognizer added by itself
+  // is ignored rather than counted against the answer.
+  const comparableTranscript = hasNoArticle(article) ? spokenWord : normalizedTranscript
```
david-venhoff: 🤔 What is the reason we use `normalizedTranscript` here instead of `spokenWord` if there is an article?
Because at this point the article should already be verified, so I don't think we have to keep verifying it later.

david-venhoff (approved):
Thanks!

Nice idea with counting the syllables, hopefully this will reduce false positives.
I tested this on my android (disclaimer: it is still probably not very representative because it does not have google services, which are usually used for speech recognition) and it worked pretty well.

The main problem I found was that the tts service still likes to drop characters sometimes, which can cause issues especially when the word ends on a vowel, because that reduces the syllable count. For example, the recognizer understood "die Krum" while the expected word was "die Krume" and flagged it as false.
But it is kind of to be expected that we get more false negatives, so it is probably not blocking :)

src/routes/training/services/SpeechMatchingService.ts
LeandraH: Good question, it's because we're comparing to the whole thing including the article in the next line. I'll rename things to make that a bit clearer.

Issue #1288, opened 2026-02-16
Update eslint to version 10


david-venhoff:
### Describe the Problem

We are 2 major versions behind on eslint, so it might make sense to update at some point.

We should probably update to react native 0.84 before starting work on this issue: https://reactnative.dev/blog/2026/02/11/react-native-0.84#eslint-v9-flat-config

### Describe Your Preferred Solution

Update to version 10. If it turns out to be a lot of effort, we can alternatively first update to version 9 and then later to version 10.

### Describe Possible Alternatives

Do nothing

david-venhoff:
Integreat PR: https://github.com/digitalfabrik/integreat-app/pull/3841

david-venhoff:
Blocked on https://github.com/jsx-eslint/eslint-plugin-react/issues/3977

PR #1500, opened 2026-08-13
1288 Upgrade to eslint 9

### LINKED ISSUE #1499: Update eslint to version 9

david-venhoff:
### Short Description

Upgrades eslint from 8 to 9.
Eslint had several breaking changes, including a different config format. 
I used https://github.com/digitalfabrik/integreat-app/pull/3841 as a template for the migration.
I first attempted to upgrade to eslint version 10, but that is not supported by some of our dependencies yet.
The issue for that is #1288

### Proposed Changes

- Modernize the eslint config format
- Fix some new lints
- Switch from `eslint-config-airbnb` to `eslint-config-airbnb-extended`, as the previous one is no longer maintained

### Side Effects

I tried my best to make sure the new config is a similar to the previous config as possible, but not all rules match up exactly. For example, the lint requiring camelCase names is a bit more lenient now.

### Testing

All CI checks are green, so maybe just verify that the config looks good and that the most important lints still fire.

### Resolved Issues

Fixes: #1499

---


david-venhoff:
Note to self: Create a new issue to enable the `react-hooks/set-state-in-effect` lint when this gets merged

JoeyStk (approved):
Thank you very much for tackling this issue! Looks good! 🚀

PR #1502, opened 2026-08-18
Fix Sentry

LeandraH:
### Short Description

Our latest test release failed because of a minor Sentry release that handles auth tokens differently. It only accepts a Sentry URL (which we need because ours is self-hosted) if it's handed in from the same source as the auth token. We hand in the token as an environment variable in CircleCI, so the newer release went looking for a Sentry URL, didn't find one, and went for the default one, where our token doesn't exist.

### Proposed Changes

- Instead of just taking the newest Sentry version, we set the version in the package.json
- Make the Sentry URL also an environment variable
- Let the release continue even if the Sentry upload fails, I think we can handle Sentry being marginally less useful

### Side Effects

N/A

### Testing

Hope for the best?

### Resolved Issues

Fixes: https://app.circleci.com/pipelines/gh/digitalfabrik/lunes-app/4763/details?useNewPipelines=true&workflowId=fa828317-fcee-432d-8130-f94836acffc7#step-114-

---

### DIFF
--- a/tools/package.json
+++ b/tools/package.json
@@ -6,6 +6,7 @@
     "app-toolbelt": "$(yarn bin app-toolbelt)"
   },
   "devDependencies": {
+    "@sentry/cli": "^3.4.0",
     "app-toolbelt": "github:digitalfabrik/app-toolbelt#semver:1.2.1"
   }
 }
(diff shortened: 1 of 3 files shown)


JoeyStk (approved):
Thank you very much!
Good job! I think this looks promising :)

PR #1505, opened 2026-08-21
Fixing a crash. Oooops.

LeandraH:
### Short Description

https://chat.tuerantuer.org/digitalfabrik/pl/corwu5qisjgi7ropmx111rjo4o

Fixing a bug that we really should have caught.

### Proposed Changes

- Add a question mark because the alternative words can be (and usually are) undefined.

### Side Effects

- Me feeling sheepish for not testing exercises when testing the PR.

### Testing

Open any training exercise, see that the app doesn't throw anymore.

### Resolved Issues

---

### DIFF
--- a/src/services/CmsApi.ts
+++ b/src/services/CmsApi.ts
@@ -159,7 +159,7 @@ const transformWordResponse = (response: WordResponse): StandardVocabularyItem =
     article: CMSArticleToArticle[article],
     images,
     audio,
-    alternatives: alternativeWords.map(({ alt_word: altWord, article: altArticle }) => ({
+    alternatives: alternativeWords?.map(({ alt_word: altWord, article: altArticle }) => ({
       word: altWord,
       article: CMSArticleToArticle[altArticle],
     })),


david-venhoff (approved):
Thanks for the quick fix!

> the alternative words can be (and usually are) undefined

Technically, I don't think they can be undefined. To me it looks like the problem is that the production server does not yet have the latest version which sends out the new fields.

So we can either be more defensive in the app or upgrade the production CMS.
If you decide for this pr, could you maybe update the `WordResponse` type to indicate that we expect `alternative_words` to be possible null?

Edit: Oh yeah, and the ai is correct that this needs a fallback value in case the alternativeWords are undefined

LeandraH:
> Technically, I don't think they can be undefined. To me it looks like the problem is that the production server does not yet have the latest version which sends out the new fields.

Hmmm, it's weird, the CMS changes for the alternative words went live a while ago, but they don't show up in https://lunes.tuerantuer.org/api/v2/words . I'll take a look at why that is. ~~But I would expect the empty alternative word to be undefined, shouldn't it? And most words don't have alternative words?~~ I see on the test environment that it's then an empty array.

I'll leave this PR open for now just in case that we somehow can't get the CMS release fixed. State of the CMS release: version 2026.8.0 was published to PyPi on Wednesday, and the Salt PR to use the new version was merged on Friday. I'm currently checking with Sven if the Highstate ran.

LeandraH:
Alright, the Highstate did run but before the PR was merged. Sven just ran another one, and the alternative words are now in the live CMS. Closing this PR, shouldn't be necessary anymore.

PR #1506, opened 2026-08-22
1433: Improve discoverability of the collection feature

### LINKED ISSUE #1433: Improve Discoverability and Clarity of the “Sammlung” Feature
### Describe the Problem

The app currently includes a feature called “Sammlung”, represented by a heart icon. Users can add their own vocabulary items there and view prioritized vocabulary. At the moment, the purpose and functionality of the feature do not seem sufficiently clear or easy to discover. In particular the name of the feature is not immediately understandable to users.

Improve the discoverability and clarity of the feature so users can easily understand:
- what the “Sammlung” feature is used for
- how to add their own vocabulary items
- where prioritized vocabulary can be found

- Rework the icon (e.g. replace the heart icon with a more suitable symbol, maybe a plus sign? )
- Rename “Collection” to a more descriptive term
- Add tooltip, onboarding hint, or explanatory text
- Improve visual distinction of saved/prioritized vocabulary

### Open Questions

- Should the heart icon be replaced completely?
- Is “Collection” the best naming for this feature?
- Do we have analytics or user feedback regarding current usage?

qaiser42:
### Short Description

Makes the "Sammlung" tab easier to understand: a consistent heart icon, an explanation of what own vocabulary is for, and empty states that tell users how to add words and favorites.

### Proposed Changes

- Favorite button uses a heart (outline/filled) instead of a star, matching the Sammlung tab icon; adds accessibility labels ("Zu Favoriten hinzufügen" / "Aus Favoriten entfernen")
- Collection overview ("Meine Vokabeln") gets a title and a short explanation; the two list items get descriptions
- Favorites empty state shows the heart icon and explains where to tap (Wortliste, Suche, Übungen) and that favorites also show up under "Vertiefen"
- Own-words list empty state hints at "Eigenes Wort erstellen"
- Release note added

Left open on purpose (product questions from the issue): renaming "Sammlung", onboarding/tooltips.

### Side Effects

- `StarCircleIconGrey` / `StarCircleIconGreyFilled` removed (only `FavoriteButton` used them)
- Every place that renders `FavoriteButton` (word lists, search, detail, exercises) now shows the heart

### Testing

- Sammlung tab → title + description, list item descriptions
- Sammlung → Vokabelliste (empty) → hint text below the empty state
- Sammlung → Favoriten (empty) → heart icon + explanation
- Suche → tap heart on a word → filled heart; word appears under Sammlung → Favoriten; tap again removes it
- TalkBack: heart button announces "Zu Favoriten hinzufügen" / "Aus Favoriten entfernen"

PR comes from a fork, so CI does not run here. Locally: `yarn lint`, `yarn ts:check`, `yarn prettier:check` and the affected jest suites pass.

### Screenshots

| Sammlung overview | Own words (empty) | Favorites (empty) |
|---|---|---|
|  |  |  |

| Search – not favorited | Search – favorited | Detail | Favorites with word |
|---|---|---|---|
|  |  |  |  |

### Resolved Issues

Fixes: #1433


src/components/FavoriteButton.tsx
```diff
@@ -40,12 +41,14 @@ const FavoriteButton = ({ vocabularyItem }: FavoriteButtonProps): ReactElement |
     }
   }
 
+  const { add, remove } = getLabels().favorites
+
   return (
-    <Button testID={isFavorite ? 'remove' : 'add'} onPress={onPress}>
+    <Button testID={isFavorite ? 'remove' : 'add'} onPress={onPress} accessibilityLabel={isFavorite ? remove : add}>
```
LeandraH, line 47: 😍

src/constants/labels.json
```diff
@@ -335,9 +337,13 @@
     "collection": "Sammlung",
     "create": "Eigenes Wort erstellen",
     "overview": {
+      "title": "Deine eigenen Wörter",
```
LeandraH: Hmm, the tab at the top says "Meine Vokabeln", I think we should stick with one wording

Image

LeandraH (commented):
Great idea but I think the issue was a bit more about making that entire tab clearer before opening it (as in, why should the user open it at all?).

I'm also not sure about having the heart icon in the navigation to lead to a screen where only half is the words marked with a heart. That means that the heart means first "Own words plus favorites", but then only the favorites.

I think it's a slightly more complicated issue and we should probably wait for Toni's input when she's back from vacation.

But I really like the descriptions for the empty cases, and the improved a11y. Maybe we strip the PR down to those two very clear improvements and wait for a design for the rest? What do you think?

qaiser42:
Good point, thanks! Stripped the PR down to the empty-state explanations and the a11y labels. Icon change and overview redesign reverted — happy to pick those up once Toni is back with a design. Also removed the „Deine eigenen Wörter“ title, so the wording inconsistency is gone.

LeandraH (approved):
Very cool, thanks!

PR #1507, opened 2026-08-22
1099: Persist only file names of user vocabulary images and audio

### LINKED ISSUE #1099: Uploaded photos in collections no longer visible in Lunes app
### Description:
Lunes is used in a job preparation class to document student internships. Each student has created their own collections and uploaded photos to them. Unfortunately, recently most of the uploaded photos have disappeared. The terms for each entry are still visible, but the corresponding images no longer appear. As an example, the teacher attached a photo of her own collection. Interestingly, one image was still visible, while the others are missing — just like in the students’ collections. (Message from Claudia Schmidt Berufliche Schule Stahl- und Maschinenbau (BS04))

### Expected behavior:
All previously uploaded photos should be visible in the collections as they were originally saved.

### Actual behavior:
Only the text/terms remain visible. Most of the images are no longer displayed.

### Attachments:
Photo of one affected collection is attached for reference.

### Additional Information

Image

qaiser42:
### Short Description

Images and audio recordings of user vocabulary items were stored with the absolute path of the document directory, which changes between app updates on iOS. Only file names are persisted now and resolved to absolute uris when loading the storage.

### Proposed Changes

- Add a `userVocabulary` storage serializer in `Storage.tsx` that strips paths to file names on persist and resolves them via `DocumentDirectoryPath` on load
- Add storage migration 6 → 7 converting existing absolute paths to file names
- Build new image/audio paths in `UserVocabularyProcessScreen` via the shared `getUserVocabularyFileUri` helper

### Side Effects

- In-memory `userVocabulary` items still hold absolute `file://` uris, so consumers (exercises, list, edit/delete) are unchanged

### Testing

- Create a user vocabulary item with images and audio, reinstall/update the app (iOS), images and audio should still be shown
- Edit and delete an existing item, files should be replaced/removed correctly

### Resolved Issues

Fixes: #1099

### DIFF
--- a/release-notes/unreleased/1099-user-vocabulary-files.yml
+++ b/release-notes/unreleased/1099-user-vocabulary-files.yml
@@ -1,0 +1,6 @@
+issue_key: 1099
+show_in_stores: true
+platforms:
+  - android
+  - ios
+de: Bilder und Audioaufnahmen eigener Vokabeln gehen nach einem App-Update nicht mehr verloren.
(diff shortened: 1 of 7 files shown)


LeandraH (approved):
Amazing, thank you for fixing that!

Testing was a bit tricky but I managed to reproduce the issue by installing the prod app, creating a custom word in there, and then installing the Testflight app without deleting the prod app first.

PR #1511, opened 2026-08-26
Fix the screenshot workflow

### LINKED ISSUE #1503: Screenshot workflow on iOS is failing
### Describe the Bug

The screenshot workflow on iOS is failing, e.g. [here](https://app.circleci.com/pipelines/gh/digitalfabrik/lunes-app/4767/details?utm_campaign=workflowcompleted_failed&utm_medium=notification&utm_content=workflowurl_workflow-email&utm_source=email&workflowId=1ac99e65-9257-45d6-82d1-a32217a5530b&job=5f099e1d-394a-4484-ae38-02a77710b012&buildNumber=14303&jobType=build#step-114-).

This is the error message:

```
Element not found: Text matching regex: Bilderrätsel.*

Element with Text matching regex: Bilderrätsel.* not found. Check the UI hierarchy in debug artifacts to verify if the element exists.

Possible causes:
- Element selector may be incorrect - check if there are similar elements with slightly different names/properties.
- Element may be temporarily unavailable due to loading state.
- This could be a real regression that needs to be addressed.
```

### Steps to Reproduce

I haven't yet tried to reproduce the behavior locally, but you can always try to run another `update_screenshots` workflow :D

### Expected Behavior

The workflow should finish successfully and open a PR with new screenshots.

david-venhoff:
### Short Description

Fixes the problem on ios and also a local issue I encountered on android

### Proposed Changes

- Add a scroll instruction to make sure that the training button is clickable (Before it was on screen, but hidden behind the navigation buttons, which caused the app to go to the wrong screen)
- Give the learning and exercise buttons test ids to make the workflow a bit more robust
- On android, the `hideKeyboard` also dismissed the popup, so I replaced it with a double click on the expected button (first click to dismiss the keyboard)

### Side Effects

- Should be none

### Testing

test the android and ios screenshots (I'll also let a CI workflow run)
The workflow was now successful (#1512)

### Resolved Issues

Fixes: #1503 

---

### DIFF
--- a/.maestro/screenshots.yaml
+++ b/.maestro/screenshots.yaml
@@ -33,9 +33,9 @@ name: Lunes store screenshots
     repeat: 6
 - tapOn: 'Development Code'
 - inputText: 'wirschaffendas'
-- hideKeyboard
 - tapOn:
     id: 'debug-seed-screenshot-data'
+    repeat: 2
 - tapOn: 'Schließen'
 
 # 01 — Home tab with seeded selected jobs.
@@ -51,8 +51,7 @@ name: Lunes store screenshots
 
 # Open the unit list (UnitSelection) for the first job.
 - tapOn:
-    text: 'Übung starten'
-    index: 0
+    id: 'start-learning-button'
 - extendedWaitUntil:
     notVisible:
       id: 'loading'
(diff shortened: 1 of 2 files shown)


LeandraH (approved):
Looks great, not tested 😅

Issue #1525, opened 2026-09-03
Open the telc area from a link or QR code


LeandraH:
As a learner I want tapping telc's link or scanning their QR code to drop me straight into the telc area, such that I never pick a Branche or Beruf first.

## Acceptance Criteria:

- [ ] Opening telc's link with the app installed lands in the telc area.
- [ ] The Branche/Beruf preselection is skipped on that path.
- [ ] The area lists only telc's Lehrwerke, sorted alphabetically — public Berufe are not mixed in.
- [ ] The area's Lektionen work in all three Lernmodi.
- [ ] Scanning the QR code does the same as tapping the link.

## Technical Tasks:

- [ ] Add deep-link handling — there is none today: no `linking` config on the navigator, no `Linking.getInitialURL`, no iOS URL scheme or associated domains, and no `VIEW` intent-filter on the launcher activity (`android/app/src/main/AndroidManifest.xml:28-31` is MAIN/LAUNCHER only; the `queries` block at :34 is for outgoing intents).
- [ ] Redeem the key from the link and store the area the way custom disciplines already are (`storageUtils.ts:208`) — no new storage shape.
- [ ] Decide universal/app links vs a custom scheme, and register the domain. This decides what telc prints, so it cannot be revisited later — see https://github.com/digitalfabrik/lunes-cms/issues/984.
- [ ] Enable manual adding with a button on the job list that opens a little text input for the activation code.

The printed QR now carries a URL, not a token, so the phone's own camera opens it. That is a different path from the in-app scanner (`QRCodeReaderOverlay.tsx`), which reads a bare token straight into `getJob({ type: 'load-protected', apiKey: code })`. Keep the in-app scanner working for existing custom disciplines; it is not the path telc's materials use.

f1sh1918:
Just some ideas from my side:
I would recommend using https and custom scheme deeplinks for the following reason. 

A user scans the qr code and opens the link
(https scheme) 

Scenario A: User has already installed the app. Link will be opened. Everything fine. Custom area will be added.

Scenario B: App is not installed. User should be redirected to a landing page. With installation instructions. After installation user can go back to the landing page and click activate button in the browser (or scan the qr code) 
Here a custom scheme link has a wider browser compatibility.

Since deeplinking could not work properly on some devices / os versions it could be helpful to also add a code that can be used for manual area adding

LeandraH:
[...]

PR #1540, opened 2026-09-14
1525: add key for content area to storage

LeandraH:
### Short Description

This PR adds the key for content areas to the storage and sends it to the CMS as part of the authorization header.

### Proposed Changes

- Add content areas to the storage
- Send the content area's key with the authorization header
- Merge all the word lists into a big one for `loadAllWords`
- A little bit of refactoring

### Side Effects

- The authorization header

### Testing

Not much to test yet, I guess you could check that the app still works.

### Resolved Issues

Part of #1525 

---


f1sh1918 (commented):
Already looks good but i think some small things are missing.

Shouldn't 
`useLoadJob` and `getJob` also receive `accessKey` ?
Currently they only get `jobId` instead of `Job` which doesn't include it

Some functions in `CmsApi` still receive `accessKey` as an optional prop. I would tend to pass the whole `Job` or `Unit` or `Word`
There are other occurrences that have an `accessKey` passed.
`getWordById` `getWordsByUnit` `getWordsByJob`, `getUnitsOfJob` `getJobs`

LeandraH:
> Shouldn't `useLoadJob` and `getJob` also receive `accessKey` ? Currently they only get `jobId` instead of `Job` which doesn't include it

True, I had the `accessKey` for `useLoadJob` in the next PR because it functionally belonged there, but may as well put it in here. I have to rewrite the next PR quite a bit anyway 😅 

`getJob` does already get the `accessKey` but no `Job` since it's the function that gets the `Job` from the `jobId`.

> Some functions in `CmsApi` still receive `accessKey` as an optional prop. I would tend to pass the whole `Job` or `Unit` or `Word` There are other occurrences that have an `accessKey` passed. `getWordById` `getWordsByUnit` `getWordsByJob`, `getUnitsOfJob` `getJobs`

Hmm, it doesn't work for all of those. `getJobs` doesn't take an `accessKey` and not all of the other cases have a full job to start with. But I guess we could add something like a `type JobRef = { id: JobId, contentAreaToken?: string }`, which would be satisfied by passing in a complete job. What do you think about that idea?

f1sh1918:
> > Shouldn't `useLoadJob` and `getJob` also receive `accessKey` ? Currently they only get `jobId` instead of `Job` which doesn't include it
> 
> True, I had the `accessKey` for `useLoadJob` in the next PR because it functionally belonged there, but may as well put it in here. I have to rewrite the next PR quite a bit anyway 😅
> 
> `getJob` does already get the `accessKey` but no `Job` since it's the function that gets the `Job` from the `jobId`.
> 
> > Some functions in `CmsApi` still receive `accessKey` as an optional prop. I would tend to pass the whole `Job` or `Unit` or `Word` There are other occurrences that have an `accessKey` passed. `getWordById` `getWordsByUnit` `getWordsByJob`, `getUnitsOfJob` `getJobs`
> 
> Hmm, it doesn't work for all of those. `getJobs` doesn't take an `accessKey` and not all of the other cases have a full job to start with. But I guess we could add something like a `type JobRef = { id: JobId, contentAreaToken?: string }`, which would be satisfied by passing in a complete job. What do you think about that idea?

Actually if you check the prs of @sascha11110 , the `contentAreaToken` is only used once for register the area (just for this endpoint)
This endpoint sends you an `accessKey` (should be stored locally) which is used to access jobs/words/units just to have a clear wording.

Okay probably i'm not deep enough in the frontend structure, i remember sth like `jobIds[]`, which will not include an accessKey i guess, so you are right.

Okay so if we don't pass `useLoadAllJobs` an `accessKey`, the job fetching for telc will be implemented in a separate pr? I was just wondering that sometimes the `accessKey` was added and sometimes not. I expected to have it fully implemented or be done in a separate PR. Else for me it looks like its missing something

LeandraH:
I renamed things to match the CMS, so we have the `code` that we get from the deep link and send to the CMS, and then save the returned `token`. I think that is pretty clear, but I'm also happy to change the naming around some more if you have better ones :)

And yes, fetching the actual job is implemented in #1541 , this is really just about the storage in the app. It made sense in my head to split it up that way but I see your point that it's confusing, sorry!

src/hooks/useLoadFavorite.ts
```diff
@@ -17,7 +17,7 @@ export const loadFavorite = async ({ storageCache, favorite }: LoadFavoriteProps
     }
     return userCreatedFavorite
   }
-  return getWordById(favorite)
+  return getWordById({ id: favorite })
```
f1sh1918, line 20: ❓ don't we have to pass the token here. some favorites might be also token scoped

f1sh1918 (commented):
Found one potential issue. rest works looks good

src/hooks/useLoadFavorite.ts
LeandraH: True, thanks!

LeandraH: Oh, though the entire favorites screen is rewritten in #1541 . Would it make reviewing easier if I just merge that PR into this one?

f1sh1918: no its fine if its done there then nvm

JoeyStk (approved):

JoeyStk:
@LeandraH could you please dismiss my review. I approved by accident :grimacing:

src/hooks/useLoadAllWords.ts
```diff
@@ -1,13 +1,40 @@
...
+  })
+}
+
+const wordsOfContentArea = async (token: string): Promise<StandardVocabularyItem[]> => {
+  try {
+    return await getWords(token)
+  } catch (error) {
+    // An unreachable content area must not take the public and the user vocabulary down with it
+    reportError(error)
+    return []
```
f1sh1918: ❓ are you sure, we shouldn't show any error here? So this will be just send to sentry and swallowed.

f1sh1918 (approved):
One open questions. Rest looks good 👍

src/hooks/useLoadAllWords.ts
LeandraH: Good point, I guess a toast or something with an error would be a good idea here. I'll create an issue for it, since we currently only have the kind of error handling where an entire page doesn't load properly :)

LeandraH: https://github.com/digitalfabrik/lunes-app/issues/1543

Issue #1534, opened 2026-09-09
Add custom content via deeplink


f1sh1918:
### Describe the Problem

After #1533, https://github.com/digitalfabrik/lunes-cms/issues/1000 and https://github.com/digitalfabrik/lunes-cms/issues/982 are ready we can verify if the passed code/accessKey is valid and the content can be added

### Describe Your Preferred Solution

- send the `accessKey` to the backend and verify that there is custom content to add
- save the`accessKey` to the localStorage of the device to be able to get area scoped data from the endpoints
- retrieve job data from the the v2 endpoint using the accessKey to get the particular area content

Issue #1532, opened 2026-09-09
Design for adding telc jobs


LeandraH:
### Describe the Problem

As discussed in our call just now, we would like a design for the following situations:

- a new user follows a deep link that opens in Lunes with a list of the telc jobs, and there is also a button there to get to the list of Lunes jobs (return from the Lunes list to the telc list via the back arrow in the header)
- an existing user wants to add another job and taps on the button for that, which opens a list with all the jobs, telc ones being separated a bit from the Lunes ones
- a new user opens the app and wants to add a telc job; entering the code happens by text copied from the page that Andy showed

### Related Issues

#1525 
https://github.com/digitalfabrik/lunes-cms/issues/984

PR #1541, opened 2026-09-16
1534: add custom content via deeplink

LeandraH:
### Short Description

This PR enables adding custom content via deeplinks, and fixes favorites not having a (potential) token.

### Proposed Changes

- The button on the activation screen does a little bit more, it sends the token to the CMS and stores the returned content area
- The `Favorites` are now normal `VocabularyListItems` with the word filtered from `useLoadAllWords`, meaning we don't specifically request each of them, and don't have to handle tokens on those.

### Side Effects

- Someone who disabled tracking adding a custom area is sent to the CMS with an empty `installation_id`.

### Testing

Run the app, then open a deep link via `npx uri
-scheme open lunes://lunes.app/activation/dashk21/ --ios OR --android`, but since the area doesn't really exist yet, you can only see the new error

### Resolved Issues

Fixes: #1534 

---


src/routes/favorites/FavoritesScreen.tsx
```diff
@@ -43,13 +43,18 @@ const EmptyStateSubtitle = styled.Text`
 
 const FavoritesScreen = ({ navigation }: FavoritesScreenProps): ReactElement => {
   const [favorites] = useStorage('favorites')
+  const { data: allWords } = useLoadAllWords()
```
f1sh1918: 🔧 error and loading state is not handled here

src/hooks/useLoadFavorite.ts
```diff
@@ -1,28 +0,0 @@
-import { Favorite } from '../constants/data'
-import VocabularyItem, { areVocabularyItemIdsEqual, VocabularyItemTypes } from '../models/VocabularyItem'
-import { getWordById } from '../services/CmsApi'
```
f1sh1918, line 3: 🔧 this is now dead code

src/routes/favorites/FavoritesScreen.tsx
```diff
@@ -43,13 +43,18 @@ const EmptyStateSubtitle = styled.Text`
...
+
+  const favoriteItems = favorites
```
f1sh1918: 🔧  Previously, useLoadFavorite called removeFavorite(storageCache, favorite) when a favorite pointed at a since-deleted user-created vocabulary item, removing it from storage. The  
  new code just filters the item out of the rendered list, but never removes it from the favorites array in storage.

f1sh1918 (commented):
Works good in general but found some issues that should be fixed

src/hooks/useLoadFavorite.ts
LeandraH: Ooooh, nice! Thanks!

f1sh1918 (approved):
Error code works fine and adding an existing area is not throwing an error :)

@LeandraH will you create further subtasks for this issue?
Since you actually can not really add content even the name of the pr assets that :)

LeandraH:
True, I guess the main remaining issue is #1532 which currently sounds like it's only the design. I guess I'll open another issue for the implementation that is blocked by the design, good idea!

